PR #41, opened 2022-12-22
Fix generate connecting to Kafka REST proxy

### LINKED ISSUE #37: Generate should not require connection to Kafka REST Proxy
recent regression introduced by #3. `generate` step should run without connecting to Kafka REST or other services

disrupted:
Fixes #37

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 .DS_Store
 .envrc
 .vscode
+pipelines/
+defaults/
--- a/config.yaml
+++ b/config.yaml
@@ -1,4 +1,5 @@
 environment: development
 broker: "http://k8kafka-cp-kafka-headless.kpops.svc.cluster.local:9092"
 kafka_connect_host: "http://localhost:8083"
+kafka_rest_host: "http://localhost:8082"
 schema_registry_url: "http://localhost:8081"
--- a/config.yaml
+++ b/config.yaml
@@ -1,5 +1,4 @@
 environment: development
 broker: "http://k8kafka-cp-kafka-headless.kpops.svc.cluster.local:9092"
 kafka_connect_host: "http://localhost:8083"
-kafka_rest_host: "http://localhost:8082"
 schema_registry_url: "http://localhost:8081"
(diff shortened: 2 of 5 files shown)


philipp94831:
It works as expected again :-)

philipp94831 (changes requested):
Please add a test that the proxy wrapper is not initialized during generate

disrupted:
> Please add a test that the proxy wrapper is not initialized during generate

the proxy wrapper is always initialized as part of the TopicHandler. it just won't have to connect to Kafka REST during `generate`

disrupted:
I'll see what I can do about a test :)

disrupted:
this should now be covered by `test_pipeline.py`. Also `kafka_rest_host` is no longer set

philipp94831:
> this should now be covered by `test_pipeline.py`. Also `kafka_rest_host` is no longer set

I don't quite see how that is covered. There has nothing been running on the port anyways so there was no connection attempt beforehand, right?

disrupted:
There was something else missing. Previously the handlers were mocked. I've removed this mock now, so that the tests in `test_pipeline.py` (which all run the `generate` command) would fail if a connection to Kafka REST proxy is attempted.

philipp94831 (approved):

raminqaf:
But this PR does not prevent calling `helm repo add` and `helm repo update` when calling `generate` right?

raminqaf (changes requested):
Generally, we should avoid initializing the Handlers when calling the `generate` command. They are not used during the generation process. This is a quick fix but still not a good solution.

philipp94831:
> Generally, we should avoid initializing the Handlers when calling the `generate` command. They are not used during the generation process. This is a quick fix but still not a good solution.

I agree. However, I think it makes sense to do this in a follow-up PR in order to get rid of the first issue (Kafka REST connectivity in generate) already. Fine for you?

.gitignore
```diff
@@ -3,3 +3,5 @@
 .DS_Store
 .envrc
 .vscode
+pipelines/
+defaults/
```
raminqaf, line 7: why did you add these?

raminqaf (approved):

philipp94831:
I created #42

.gitignore
disrupted: I usually keep these around for testing. felt a bit anxious to accidentally commit some of it, so I added it to ignore to be safe

disrupted:
> But this PR does not prevent calling `helm repo add` and `helm repo update` when calling `generate` right?

no, this is only targeting #37

PR #47, opened 2022-12-23
Refactor Helm Wrapper

### LINKED ISSUE #15: Helm wrapper should be refactored and work as a Python client for Helm CLI
Initially, the helm wrapper was designed to only support one Helm repo (bakdata-common/streams-bootstrap). But now we use it also for Kafka connect resetters and it can also be used for other deployments per project. We should separate the helm wrapper/Helm execution from the configuration the helm wrapper uses that is specific to a Helm repo/chart.

raminqaf:
Closes #15


kpops/component_handlers/helm_wrapper/helm.py
```diff
@@ -0,0 +1,242 @@
...
+
+RELEASE_NAME_MAX_LEN = 53
+
+
+class Helm:
+    def __init__(self, helm_config: HelmConfig) -> None:
+        self._context = helm_config.context
+        self._debug = helm_config.debug
+
+    def helm_repo_add(
```
philipp94831, line 31: Do we really need helm as part of the name? shouldn't we just call it add_repo?

kpops/component_handlers/kafka_connect/connector_handler.py
```diff
@@ -210,27 +230,29 @@ def clean_connector(
...
             magentaify(
                 f"Connector Cleanup: deploy Connect {connector_type.value} resetter for {connector_name}"
             )
         )
-        self._helm_wrapper.helm_upgrade_install(
-            release_name=release_name,
-            suffix=suffix,
+
+        stdout = self._helm_wrapper.helm_upgrade_install(
+            release_name=trim_release_name(clean_up_release_name, suffix),
```
philipp94831, line 249: Trimming also needs to be done when deleting the old job

kpops/component_handlers/kafka_connect/connector_handler.py
```diff
@@ -245,32 +267,36 @@ def clean_connector(
...
-            release_name=release_name,
-            suffix=suffix,
+            release_name=trim_release_name(release_name, suffix),
             dry_run=dry_run,
         )
 
     @classmethod
-    def from_pipeline_config(
-        cls, pipeline_config: PipelineConfig
-    ) -> ConnectorHandler:  # TODO: annotate as typing.Self once mypy supports it
```
philipp94831, line 265: What happened to the ToDO?

kpops/component_handlers/streams_bootstrap/app_handler.py
```diff
@@ -21,35 +23,54 @@
...
+            RepoAuthFlags(
+                username=helm_repo_config.username, password=helm_repo_config.password
+            ),
+        )
+        self.helm_diff = helm_diff
 
     def install_app(
         self,
         release_name: str,
+        application_type: ApplicationType,
```
philipp94831, line 47: Why do we even use an enum here? In one project we actually introduce a new type and typing is broken because we cannot change the enum.

philipp94831 (changes requested):
Looking great!

kpops/component_handlers/helm_wrapper/helm.py
raminqaf: yeah I don't think we need the prefix

kpops/component_handlers/kafka_connect/connector_handler.py
raminqaf: Sry a harsh refactoring! 😄

kpops/component_handlers/streams_bootstrap/app_handler.py
raminqaf: The enum is then used to specify the chart. Because the `install_app` method is called from the producer as well as the streams app. So They need to specify the application that needs to be installed.
```python
        stdout = self._helm_wrapper.upgrade_install(
            release_name=release_name,
            chart=f"{self.repository_name}/{application_type.value}", <------- It is used here!
            dry_run=dry_run,
            namespace=namespace,
            values=values,
            flags=HelmUpgradeInstallFlags(version=self.chart_version),
        )
```
Another approach would be to pass this information true the `HelmRepoConfig` class. WDYT?

kpops/component_handlers/kafka_connect/connector_handler.py
raminqaf: I am trimming it in the method:
```python
    def __delete_clean_up_job_release(
        self, release_name: str, suffix: str, dry_run: bool
    ) -> None:
        self._helm_wrapper.uninstall(
            namespace=self.namespace,
            release_name=trim_release_name(release_name, suffix),
            dry_run=dry_run,
        )
```

raminqaf: I still need some tests for these trimmings. That is still WIP for this PR!

kpops/component_handlers/streams_bootstrap/app_handler.py
philipp94831: I think a str would be more appropriate. And we can use the enum values as str constants instead

kpops/component_handlers/kafka_connect/connector_handler.py
philipp94831: Hm, I think it's safer to not trim in the method but extract the release name to a variable once and use that everywhere

kpops/component_handlers/helm_wrapper/helm.py
```diff
@@ -0,0 +1,240 @@
...
+
+RELEASE_NAME_MAX_LEN = 53
+
+
+class Helm:
+    def __init__(self, helm_config: HelmConfig) -> None:
+        self._context = helm_config.context
+        self._debug = helm_config.debug
+
+    def repo_add(
```
philipp94831, line 31: I think it reads nicer if the verb comes first, i.e., add_repo

kpops/component_handlers/streams_bootstrap/app_handler.py
```diff
@@ -21,35 +20,52 @@
...
 
     def install_app(
         self,
         release_name: str,
+        application_type: str,
         namespace: str,
         values: dict,
-        app_type: ApplicationType,
         dry_run: bool,
-        local_chart_path: Path | None = None,
```
philipp94831, line 35: Hm, we used that before. I think we should still make it possible to use that. I would even say we should just pass through the chart parameter

kpops/component_handlers/streams_bootstrap/app_handler.py
```diff
@@ -21,35 +20,52 @@
...
+            chart=f"{self.repository_name}/{application_type}",
+            dry_run=dry_run,
             namespace=namespace,
             values=values,
-            app=app_type.value,
-            dry_run=dry_run,
-            local_chart_path=local_chart_path,
+            flags=HelmUpgradeInstallFlags(version=self.chart_version),
         )
+        if dry_run and self.helm_diff.config.enable:
```
philipp94831, line 64: There should also be a way to make this diff reusable for custom Kubernetes apps

kpops/components/base_components/kubernetes_app.py
```diff
@@ -21,19 +27,86 @@ class Config(CamelCaseConfig):
...
+        return self.app.namespace
+
+    def deploy(self, dry_run: bool) -> None:
+        stdout = self.helm_wrapper.upgrade_install(
+            release_name=self.helm_release_name,
+            chart=self.get_helm_chart(),
+            dry_run=dry_run,
+            namespace=self.namespace,
+            values=self.to_helm_values(),
+            flags=HelmUpgradeInstallFlags(version=self.get_helm_repo_config().version),
```
philipp94831, line 71: Our custom component does not have a repo config (we install a local chart). So we should instead extract a method get_version which returns None by default and can be overridden

philipp94831 (changes requested):
Helm repo add does not seem to be executed when running `deploy --dry-run`.
`Helm - Could not install chart. More details: Error: repo bakdata-streams-bootstrap not found`

kpops/components/base_components/kubernetes_app.py
raminqaf: You can override the `get_helm_repo_config` and set the version to `None` there.

philipp94831: Yeah but no repo is needed in our case. And it has some required fields (name/url) which I need to fill

philipp94831: Probably version should part ways with the repo config. They are unrelated in how one uses helm

kpops/components/base_components/kubernetes_app.py
```diff
@@ -100,10 +115,15 @@ def print_helm_diff(self, stdout: str):
...
         self.helm_diff.log_helm_diff(helm_diff, log)
 
-    def get_helm_chart(self) -> str:
+    def get_helm_repo_config(self) -> HelmRepoConfig | NoReturn:
         raise NotImplementedError()
 
-    def get_helm_repo_config(self) -> HelmRepoConfig:
+    def get_helm_chart_version(self) -> str | None:
+        if self.get_helm_repo_config():
+            return self.get_helm_repo_config().version
```
philipp94831, line 123: Hm, I think you should remove it from the config and create a separate field

kpops/components/base_components/kubernetes_app.py
```diff
@@ -100,10 +115,15 @@ def print_helm_diff(self, stdout: str):
         helm_diff = HelmDiff.get_diff(current_release, new_release)
         self.helm_diff.log_helm_diff(helm_diff, log)
 
-    def get_helm_chart(self) -> str:
+    def get_helm_repo_config(self) -> HelmRepoConfig | NoReturn:
         raise NotImplementedError()
```
philipp94831, line 119: default should be None

philipp94831 (dismissed):

kpops/components/base_components/kubernetes_app.py
```diff
@@ -21,19 +29,104 @@ class Config(CamelCaseConfig):
...
+
+    def print_helm_diff(self, stdout: str):
+        current_release = self.helm_wrapper.get_manifest(
+            self.helm_release_name, self.namespace
+        )
+        new_release = Helm.load_helm_manifest(stdout)
+        helm_diff = HelmDiff.get_diff(current_release, new_release)
+        self.helm_diff.log_helm_diff(helm_diff, log)
+
+    def get_helm_repo_config(self) -> HelmRepoConfig | NoReturn:
```
philipp94831, line 118: Why NoReturn and not None?

kpops/components/base_components/kubernetes_app.py
```diff
@@ -72,31 +72,26 @@ def namespace(self) -> str:
...
-        try:
-            stdout = self.helm_wrapper.uninstall(
-                namespace=self.namespace,
-                release_name=self.helm_release_name,
-                dry_run=dry_run,
-            )
-            if dry_run and self.helm_diff.config.enable:
-                self.print_helm_diff(stdout)
-        except Exception as exception:
-            if len(exception.args) == 1 and "release: not found" in exception.args[0]:
```
philipp94831: What happened to this?

raminqaf: The helm wrapper already catches the exception and exits if something doesn't work properly. I thought this might be a bit redundant and too generic for exception handling because we are catching `Excepition`.

kpops/components/streams_bootstrap/producer/producer_app.py
```diff
@@ -0,0 +1,63 @@
...
+    @override
+    def add_extra_output_topic(self, topic_name: str, role: str) -> None:
+        self.app.streams.extra_output_topics[role] = topic_name
+
+    @override
+    def get_helm_chart(self) -> str:
+        return f"{self.config.streams_bootstrap_helm_config.repository_name}/{AppType.PRODUCER_APP.value}"
+
+    @override
+    def get_helm_repo_config(self) -> HelmRepoConfig | NoReturn:
```
disrupted, line 53: why did you annotate this with `NoReturn`? it can only be used on functions which never return aka _always_ raise an exception

kpops/components/streams_bootstrap/producer/producer_app.py
```diff
@@ -0,0 +1,63 @@
...
+
+    class Config(BaseConfig):
+        extra = Extra.allow
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    @override
+    def apply_to_outputs(self, name: str, topic: TopicConfig) -> None:
+        match (topic.type):
```
disrupted, line 34: is it possible to remove the parentheses around the subject?

disrupted (changes requested):

kpops/components/streams_bootstrap/producer/producer_app.py
raminqaf: Sry about that! 😄 It was a mistake, and I forgot to fix it!

kpops/components/streams_bootstrap/producer/producer_app.py
raminqaf: It is!

kpops/components/base_components/kubernetes_app.py
```diff
@@ -21,19 +27,95 @@ class Config(CamelCaseConfig):
...
+        if self._helm_wrapper is None:
+            self._helm_wrapper = Helm(self.config.helm_config)
+            helm_repo_config = self.get_helm_repo_config()
+            if helm_repo_config is not None:
+                self._helm_wrapper.add_repo(
+                    helm_repo_config.repository_name,
+                    helm_repo_config.url,
+                    helm_repo_config.repo_auth_flags,
+                )
+        return self._helm_wrapper
```
disrupted, line 56: perhaps you could consider using `cached_property`. then you don't need to do the caching manually using class vars

kpops/components/streams_bootstrap/producer/producer_app.py
disrupted: cool, I would prefer this slightly leaner syntax. what do you think?

raminqaf: Yeah feels more elegant 🌟

kpops/components/base_components/kubernetes_app.py
raminqaf: Okay I changed it to this:
```python
    @cached_property
    def helm_wrapper(self) -> Helm:
        helm_wrapper = Helm(self.config.helm_config)
        helm_repo_config = self.get_helm_repo_config()
        if helm_repo_config is not None:
            helm_wrapper.add_repo(
                helm_repo_config.repository_name,
                helm_repo_config.url,
                helm_repo_config.repo_auth_flags,
            )
        return helm_wrapper
```

but I am getting the following error:
```console
tests/components/test_kubernetes_app.py:None (tests/components/test_kubernetes_app.py)
tests/components/test_kubernetes_app.py:15: in <module>
    from kpops.components.base_components.kubernetes_app import (
kpops/components/__init__.py:1: in <module>
    from kpops.components.base_components import (
kpops/components/base_components/__init__.py:1: in <module>
    from kpops.components.base_components.kafka_app import KafkaApp
kpops/components/base_components/kafka_app.py:8: in <module>
    from kpops.components.base_components.kubernetes_app import (
kpops/components/base_components/kubernetes_app.py:31: in <module>
    class KubernetesApp(PipelineComponent):
pydantic/main.py:222: in pydantic.main.ModelMetaclass.__new__
    ???
pydantic/fields.py:506: in pydantic.fields.ModelField.infer
    ???
pydantic/fields.py:436: in pydantic.fields.ModelField.__init__
    ???
pydantic/fields.py:546: in pydantic.fields.ModelField.prepare
    ???
pydantic/fields.py:570: in pydantic.fields.ModelField._set_default_and_type
    ???
pydantic/fields.py:439: in pydantic.fields.ModelField.get_default
    ???
pydantic/utils.py:695: in pydantic.utils.smart_deepcopy
    ???
/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:172: in deepcopy
    y = _reconstruct(x, memo, *rv)
/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:271: in _reconstruct
    state = deepcopy(state, memo)
/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:146: in deepcopy
    y = copier(x, memo)
/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:231: in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/copy.py:161: in deepcopy
    rv = reductor(4)
E   TypeError: cannot pickle '_thread.RLock' object
```

raminqaf: It seems that they don't support it...
https://github.com/pydantic/pydantic/issues/1241

disrupted: ah too bad, typical pydantic headache 🙄
at least the workaround looks alright

tests/components/test_producer_app.py
```diff
@@ -0,0 +1,260 @@
...
+            producer_app.helm_wrapper, "uninstall"
+        )
+
+        producer_app.destroy(dry_run=True, clean=False, delete_outputs=False)
+
+        mock_helm_uninstall.assert_called_once_with(
+            "test-namespace", "example-name", True
+        )
+
+    # TODO: Assert schema deletion
```
raminqaf, line 191: @disrupted let's do this tomorrow together. For some reason I couldn't access the schema handler mock 🤔

disrupted: @raminqaf yep, feel free to ping me tomorrow

kpops/component_handlers/schema_handler/schema_handler.py
```diff
@@ -19,28 +20,24 @@
...
-                    f"Either implement it by inheriting from {SchemaProvider.__module__}.{SchemaProvider.__name__} and implementing its abstract method."
+                    "The Schema Registry URL is set but you haven't specified the component module path. Please provide a valid component module path where your SchemaProvider implementation exists."
                 )
-        return self._schema_provider
+            schema_provider_class = find_class(self.components_module, SchemaProvider)  # type: ignore[type-abstract]
+            return schema_provider_class()
+        except ClassNotFoundError:
+            raise ValueError(
+                f"No schema provider found in components module {self.components_module}. "
+                f"Either implement it by inheriting from {SchemaProvider.__module__}.{SchemaProvider.__name__} and implementing its abstract method."
```
disrupted, line 39: The "Either" probably lost its meaning here from a previous refactoring. There only seems to be one solution

kpops/component_handlers/schema_handler/schema_handler.py
```diff
@@ -19,28 +20,24 @@
...
-                    )
-                schema_provider_class = find_class(self.components_module, SchemaProvider)  # type: ignore[type-abstract]
-                self._schema_provider = schema_provider_class()
-            except ClassNotFoundError:
+        try:
+            if not self.components_module:
                 raise ValueError(
-                    f"No schema provider found in components module {self.components_module}. "
-                    f"Either implement it by inheriting from {SchemaProvider.__module__}.{SchemaProvider.__name__} and implementing its abstract method."
+                    "The Schema Registry URL is set but you haven't specified the component module path. Please provide a valid component module path where your SchemaProvider implementation exists."
```
disrupted, line 32:
```suggestion
                    f"The Schema Registry URL is set but you haven't specified the component module path. Please provide a valid component module path where your {SchemaProvider.__name__} implementation exists."
```

tests/component_handlers/kafka_connect/test_connect_handler.py
```diff
@@ -276,36 +291,34 @@ def test_should_call_helm_upgrade_install_and_uninstall_when_clean_connector_wit
...
+                    release_name="test-connector-clean",
+                    namespace="test-namespace",
+                    chart=f"{helm_repo_config.repository_name}/kafka-connect-resetter",
                     dry_run=True,
-                    helm_command_config=HelmCommandConfig(
-                        debug=False,
-                        force=False,
-                        timeout="5m0s",
+                    flags=HelmUpgradeInstallFlags(
+                        version="1.0.4",
```
disrupted, line 307: this is already set as the default value?

disrupted (changes requested):

kpops/components/base_components/kubernetes_app.py
```diff
@@ -21,19 +28,92 @@ class Config(CamelCaseConfig):
...
+    def get_helm_repo_config(self) -> HelmRepoConfig | None:
+        return None
+
+    def get_helm_chart(self) -> str:
+        raise NotImplementedError(
+            f"Please implement the get_helm_chart() method of the {self.__module__} module."
+        )
+
+    def get_helm_chart_version(self) -> str | None:
+        return self.version
```
disrupted, line 114: what is the purpose of this function? public attributes can be accessed directly, there's no need for a getter-function in Python. If it should be "read-only" make it a `@property`

disrupted (changes requested):

philipp94831:
Running a clean job does not work if the clean release does not exist (which is totally fine). Worked before

tests/component_handlers/kafka_connect/test_connect_handler.py
raminqaf: Yeah it is the default. I moved it here:
https://github.com/bakdata/kpops/blob/7f6cf1d62efe2f511e5d0cdd44eeab24044413cb/kpops/cli/pipeline_config.py#L35

kpops/cli/pipeline_config.py
```diff
@@ -20,19 +24,20 @@ class TopicNameConfig(BaseSettings):
...
+class KafkaConnectResetterHelmConfig(BaseSettings):
+    helm_config: HelmRepoConfig = Field(
+        default=HelmRepoConfig(
             repository_name="bakdata-kafka-connect-resetter",
             url="https://bakdata.github.io/kafka-connect-resetter/",
-            version="1.0.4",
         ),
         description="Configuration of Kafka connect resetter Helm Chart",
     )
+    version = "1.0.4"
```
disrupted, line 35: this requires a type annotation because dataclasses / Pydantic models can't just infer them

disrupted (changes requested):

kpops/component_handlers/helm_wrapper/helm.py
```diff
@@ -0,0 +1,230 @@
...
+            stderr=subprocess.PIPE,
+            text=True,
+        )
+        process.wait()
+        stdout, stderr = process.communicate()
+
+        log.debug(stdout)
+
+        Helm.parse_helm_command_stderr_output(stderr)
+        return stdout
```
disrupted, line 210: I would prefer the higher level `subprocess.run` API

```suggestion
        process = subprocess.run(
            command,
            stdout=subprocess.PIPE,
            stderr=subprocess.PIPE,
            text=True,
        )
        log.debug(stdout)
        Helm.parse_helm_command_stderr_output(process.stderr)
        return process.stdout
```

kpops/component_handlers/helm_wrapper/helm.py
```diff
@@ -0,0 +1,230 @@
...
+            bash_command.append("--force")
+        if helm_command_config.wait:
+            bash_command.append("--wait")
+        if helm_command_config.wait_for_jobs:
+            bash_command.append("--wait-for-jobs")
+        if helm_command_config.version:
+            bash_command.extend(["--version", helm_command_config.version])
+        return bash_command
+
+    def __execute(self, command: list[str]) -> str:
```
disrupted, line 195: there's a mix of `command` and `bash_command` in this class which is a bit inconsistent. Let's decide on one

kpops/component_handlers/kafka_connect/connector_handler.py
```diff
@@ -134,6 +148,57 @@ def destroy_connector(self, connector_name: str, dry_run: bool) -> None:
...
+                f"Connector Cleanup: deploy Connect {connector_type.value} resetter for {connector_name}"
+            )
+        )
+
+        stdout = self.__install_kafka_resetter(
+            trimmed_name, connector_name, connector_type, dry_run, kwargs
+        )
+
+        if dry_run and self.helm_diff.config.enable:
+            current_release = self._helm_wrapper.get_manifest(
```
disrupted, line 191: how about `self._helm` as the class is no longer called `HelmWrapper`?

kpops/components/streams_bootstrap/__init__.py
```diff
@@ -1,2 +1,6 @@
-from kpops.components.streams_bootstrap.producer import ProducerApp  # noqa F401
-from kpops.components.streams_bootstrap.streams_app import StreamsApp  # noqa F401
+from kpops.components.streams_bootstrap.producer.producer_app import (  # noqa F401
```
disrupted, line 1: please remove `noqa F401` and set
```python
__all__ = (ProducerApp, StreamsApp,)
```

kpops/components/streams_bootstrap/streams/streams_app.py
```diff
@@ -0,0 +1,79 @@
...
+    @override
+    def get_helm_chart(self) -> str:
+        return f"{self.config.streams_bootstrap_helm_config.repository_name}/{AppType.STREAMS_APP.value}"
+
+    @override
+    def get_helm_repo_config(self) -> HelmRepoConfig | None:
+        return self.config.streams_bootstrap_helm_config
+
+    @override
+    def get_clean_up_helm_chart(self):
```
disrupted, line 62: missing return type

disrupted (changes requested):
overall looks great! 🌟 thanks for completing the component tests

kpops/component_handlers/kafka_connect/model.py
```diff
@@ -41,3 +44,25 @@ class KafkaConnectConfigErrorResponse(BaseModel):
...
+
+class KafkaConnectResetterValues(BaseModel):
+    connector_type: Literal["source", "sink"]
+    config: KafkaConnectResetterConfig
+    name_override: str
+
+    class Config(CamelCaseConfig):
+        pass
+
+    def dict(self, **_):
```
disrupted, line 67: consider adding a return type annotation

poetry.lock
```diff
@@ -25,17 +25,18 @@ trio = ["trio (>=0.16,<0.22)"]
...
 [package.extras]
-dev = ["cloudpickle", "coverage[toml] (>=5.0.2)", "furo", "hypothesis", "mypy (>=0.900,!=0.940)", "pre-commit", "pympler", "pytest (>=4.3.0)", "pytest-mypy-plugins", "sphinx", "sphinx-notfound-page", "zope.interface"]
-docs = ["furo", "sphinx", "sphinx-notfound-page", "zope.interface"]
-tests = ["cloudpickle", "coverage[toml] (>=5.0.2)", "hypothesis", "mypy (>=0.900,!=0.940)", "pympler", "pytest (>=4.3.0)", "pytest-mypy-plugins", "zope.interface"]
-tests-no-zope = ["cloudpickle", "coverage[toml] (>=5.0.2)", "hypothesis", "mypy (>=0.900,!=0.940)", "pympler", "pytest (>=4.3.0)", "pytest-mypy-plugins"]
+cov = ["attrs[tests]", "coverage-enable-subprocess", "coverage[toml] (>=5.3)"]
+dev = ["attrs[docs,tests]"]
+docs = ["furo", "myst-parser", "sphinx", "sphinx-notfound-page", "sphinxcontrib-towncrier", "towncrier", "zope.interface"]
+tests = ["attrs[tests-no-zope]", "zope.interface"]
+tests-no-zope = ["cloudpickle", "cloudpickle", "hypothesis", "hypothesis", "mypy (>=0.971,<0.990)", "mypy (>=0.971,<0.990)", "pympler", "pympler", "pytest (>=4.3.0)", "pytest (>=4.3.0)", "pytest-mypy-plugins", "pytest-mypy-plugins", "pytest-xdist[psutil]", "pytest-xdist[psutil]"]
```
disrupted, line 39: is it necessary to update all the dependencies in the lockfile? I would prefer moving this to a separate PR if it's not needed for the Helm refactoring

disrupted (commented):
otherwise LGTM :)

poetry.lock
raminqaf: @disrupted Good point. I reverted it to the version on the main branch

disrupted (approved):

PR #52, opened 2023-01-02
Refactor usage of Pydantic aliases

disrupted:


tests/component_handlers/schema_handler/test_schema_handler.py
```diff
@@ -149,8 +149,8 @@ def test_should_log_info_when_submit_schemas_that_not_exists_and_dry_run_true(
     )
     topic_config = TopicConfig(
         type=OutputTopicTypes.OUTPUT,
-        keySchema=None,
-        valueSchema="com.bakdata.kpops.test.SchemaHandlerTest",
+        key_schema=None,
```
philipp94831, line 152: Is this breaking our existing configs?

disrupted: no, they use aliases for the camel case version

kpops/components/base_components/models/to_section.py
```diff
@@ -28,6 +28,7 @@ class TopicConfig(BaseModel):
 
     class Config(BaseConfig):
         extra = Extra.forbid
+        allow_population_by_field_name = True
```
raminqaf, line 31: Does this mean that we use the field name (like `shema_registry_url`) in the `pipeline.yaml`?

disrupted: yes, both snake case and camel case version are supported, although snake case should be preferred

raminqaf (approved):

PR #53, opened 2023-01-02
Use component name instead of type to set default output topic name

MichaelKora:


tests/pipeline/test_pipeline.py
```diff
@@ -240,3 +242,38 @@ def test_with_custom_config(self, tmpdir, snapshot):
...
+                str(RESOURCE_PATH / "no-topics-defaults"),
+                "--config",
+                str(RESOURCE_PATH / "no-topics/config.yaml"),
+            ],
+        )
+        assert result.exit_code == 0, result.exception
+
+        enriched_pipeline = load_yaml_file(Path(output_file_path))
+        assert isinstance(enriched_pipeline, dict)
+        for app_details in enriched_pipeline["components"]:
```
philipp94831, line 269: Do not iterate but rather make assertions on each element of the list

tests/pipeline/test_pipeline.py
```diff
@@ -240,3 +242,38 @@ def test_with_custom_config(self, tmpdir, snapshot):
...
+            nameOverride = app_details["app"]["nameOverride"]
+            output_topic = app_details["app"]["streams"]["outputTopic"]
+            assert output_topic == nameOverride
```
philipp94831, line 272: Assert against the exact value and not something that is coincidentally equal

tests/pipeline/test_pipeline.py
```diff
@@ -240,3 +242,38 @@ def test_with_custom_config(self, tmpdir, snapshot):
...
+                "--config",
+                str(RESOURCE_PATH / "no-topics/config.yaml"),
+            ],
+        )
+        assert result.exit_code == 0, result.exception
+
+        enriched_pipeline = load_yaml_file(Path(output_file_path))
+        assert isinstance(enriched_pipeline, dict)
+        for app_details in enriched_pipeline["components"]:
+            nameOverride = app_details["app"]["nameOverride"]
```
disrupted, line 270: please use _snake_case_ for variable names

disrupted (changes requested):

tests/pipeline/test_pipeline.py
```diff
@@ -240,12 +240,51 @@ def test_with_custom_config(self, tmpdir, snapshot):
...
-                error_topic = app_details["app"]["streams"]["errorTopic"]
-                assert error_topic == "random-error"
+
+        producer_details = enriched_pipeline["components"][0]
+        output_topic = producer_details["app"]["streams"]["outputTopic"]
+        assert output_topic == "random-topic"
+
+        streams_app_details = enriched_pipeline["components"][1]
+        output_topic = streams_app_details["app"]["streams"]["outputTopic"]
+        assert output_topic == "random-topic"
```
philipp94831, line 250: So the output topic is the same for both steps?

tests/pipeline/test_pipeline.py
```diff
@@ -240,12 +240,51 @@ def test_with_custom_config(self, tmpdir, snapshot):
...
+                str(RESOURCE_PATH / "no-topics/pipeline.yaml"),
+                "--save",
+                "--out-path",
+                output_file_path,
+                "--defaults",
+                str(RESOURCE_PATH / "no-topics-defaults"),
+                "--config",
+                str(RESOURCE_PATH / "no-topics/config.yaml"),
+            ],
+        )
```
disrupted, line 274: please add `catch_exceptions=False` to the runner method

tests/pipeline/test_pipeline.py
```diff
@@ -240,12 +240,51 @@ def test_with_custom_config(self, tmpdir, snapshot):
...
+        assert result.exit_code == 0, result.exception
```
disrupted, line 275:
```suggestion
        assert result.exit_code == 0
```

this is no longer needed if the runner doesn't catch exceptions

tests/pipeline/resources/no-topics/config.yaml
```diff
@@ -0,0 +1,5 @@
+environment: development
```
disrupted, line 1: aren't those the same contents as the default `config.yaml`?

tests/pipeline/resources/no-topics/pipeline.yaml
```diff
@@ -0,0 +1,13 @@
+- type: producer
```
disrupted, line 1: isn't that the same file as `tests/pipeline/resources/custom-config/pipeline.yaml`?

tests/pipeline/test_pipeline.py
MichaelKora: yes i declared it [here](https://github.com/bakdata/kpops/blob/feat/default-topics/tests/pipeline/resources/custom-config/config.yaml#L4)

But if this might cause some confusions, i could change it to `"${component_name}-test-topic"`

tests/pipeline/resources/no-topics/config.yaml
MichaelKora: indeed they are.. i can get rid of one of them

tests/pipeline/resources/no-topics/pipeline.yaml
MichaelKora: indeed they are.. i can get rid of one of them

disrupted (approved):
LGTM, just merge/rebase on main

PR #60, opened 2023-01-10
Refactor Kubernetes app properties

disrupted:


kpops/components/base_components/kafka_app.py
```diff
@@ -28,7 +28,7 @@ class Config(CamelCaseConfig):
 
 class KafkaAppConfig(KubernetesAppConfig):
     streams: KafkaStreamsConfig
-    nameOverride: str | None
+    name_override: str | None
```
philipp94831, line 31: does this affect the way values.yaml are generated? because we require it to be `nameOverride`

disrupted: I don't think so, it should use the camel case version for serialization just like `StreamsConfig` and `KafkaStreamsConfig`. Should be covered by the integration tests in `test_pipeline.py` as well

raminqaf (approved):

PR #70, opened 2023-01-26
Generate schema for pipeline.yaml and config.yaml

### LINKED ISSUE #14: Provide JSON schema for pipeline yaml
Writing complex YAML files like pipelines for KPOps, it is pretty helpful to have IDE support. This can be easily achieved by providing an appropriate JSON schema. Pydantic supports the export: https://pydantic-docs.helpmanual.io/usage/schema/

disrupted:
Closes #14 
Supersedes #67


gen_schema.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+schema.field_schema = field_schema
+
+ComponentType = (
+    StreamsApp
+    | ProducerApp
+    | KafkaSourceConnector
+    | KafkaSinkConnector
+    # | PipelineComponent
```
raminqaf, line 32: Do we need this?

raminqaf:
what is the reason behind all these conflicts!? 😱

gen_schema.py
disrupted: no, it's unnecessary

sujuka99:
If you generate a schema from PipelineConfig, we could have autocompletion for config.yaml as well. It was much simpler and worked out of the box. I did it by accident while exploring.

Simply `{PipelineConfig.schema_json(indent=2)` generated `pipeline_config.json` in [this PR](https://github.com/bakdata/kpops/pull/67/files#diff-55d4aac95c43ee8c92de827972ed5fd3c002f04e66c0817c1f726d1aec1ce346). 

I see no downsides of having auto completion for that file as well, but I don't know if it should be added in this PR or a subsequent one.

sujuka99 (dismissed):
Looks great!


Just a small comment, if you generate a schema from `PipelineConfig`, we could have autocompletion for `config.yaml` as well. It was much simpler and worked out of the box. I did it by accident while exploring.

Simply `{PipelineConfig.schema_json(indent=2)` generated `pipeline_config.json` in [this PR](https://github.com/bakdata/kpops/pull/67/files#diff-55d4aac95c43ee8c92de827972ed5fd3c002f04e66c0817c1f726d1aec1ce346). 

I see no downsides of having auto completion for that file as well, but I don't know if it should be added in this PR or a subsequent one.

gen_schema.py
```diff
@@ -0,0 +1,41 @@
+from typing import Annotated, Any, Sequence
```
raminqaf, line 1: do we want to keep this file at the root of the project? Or should we move it to the kpops folder?

gen_schema.py
```diff
@@ -0,0 +1,41 @@
...
+from pydantic.schema import SkipField
+
+from kpops.components.base_components.kafka_connect import (
+    KafkaSinkConnector,
+    KafkaSourceConnector,
+)
+from kpops.components.streams_bootstrap.producer.producer_app import ProducerApp
+from kpops.components.streams_bootstrap.streams.streams_app import StreamsApp
+
+original_field_schema = schema.field_schema
```
raminqaf, line 14: Move to the `field_schema` method?

gen_schema.py
```diff
@@ -0,0 +1,41 @@
...
+def field_schema(field: ModelField, **kwargs: Any) -> Any:
+    if field.field_info.extra.get("hidden_from_schema"):
+        raise SkipField(f"{field.name} field is being hidden")
+    else:
+        return original_field_schema(field, **kwargs)
+
+
+schema.field_schema = field_schema
+
+PipelineComponent = StreamsApp | ProducerApp | KafkaSourceConnector | KafkaSinkConnector
```
raminqaf, line 27: Should we also add `KubernetesApp`?

kpops/components/base_components/base_defaults_component.py
```diff
@@ -17,21 +17,23 @@
...
+    type: ClassVar[str] = Field(default=..., const=True)
 
-    enrich: bool = Field(default=False, exclude=True)
-    config: PipelineConfig = Field(default=..., exclude=True)
-    handlers: ComponentHandlers = Field(default=..., exclude=True)
+    enrich: bool = Field(default=False, exclude=True, hidden_from_schema=True)
+    config: PipelineConfig = Field(default=..., exclude=True, hidden_from_schema=True)
+    handlers: ComponentHandlers = Field(
+        default=..., exclude=True, hidden_from_schema=True
+    )
```
raminqaf, line 26: 🤩

kpops/components/streams_bootstrap/producer/producer_app.py
```diff
@@ -19,7 +21,10 @@ class ProducerApp(KafkaApp):
     This producer holds configuration to use as values for the streams bootstrap produce helm chart.
     """
 
-    _type = "producer"
+    type: ClassVar[str] = "producer"
+    schema_type: Literal["producer"] = Field(
```
raminqaf, line 25: Just for curiosity, What is the difference between `Literal` and `Final` in python? Could you define schema_type `Final`?

kpops/components/streams_bootstrap/streams/streams_app.py
```diff
@@ -13,7 +15,10 @@ class StreamsApp(KafkaApp):
     StreamsApp component that configures a streams bootstrap app
     """
 
-    _type = "streams-app"
+    type: ClassVar[str] = "streams-app"
+    schema_type: Literal["streams-app"] = Field(
+        default="streams-app", alias="schema_type", exclude=True
```
raminqaf, line 20: Do we need the `alias` property here? Since the name of the field is exactly the same as the `alias`

raminqaf (commented):
LGTM! I had some notices and questions!

gen_schema.py
disrupted: ideally this should be a kpops command imo, but I would like to do it in a followup when we also add the documentation for it. for now I would simply make it a standalone script if that's ok

gen_schema.py
disrupted: that doesn't work. we're monkeypatching one of pydantic's builtin functions and before that we have to store the original function somewhere. That's what this line does

gen_schema.py
disrupted: good point, should we also add `KafkaApp` then?

kpops/components/streams_bootstrap/producer/producer_app.py
disrupted: `Literal` is used for the discriminator (differentiation of models) here as it is a literal string value. It's needed for the schema generation, so that it can select the right model schema based on the `type` of a pipeline step. 
`Final` I haven't used so far, but from my understanding it is used as a special indicator for type checkers to tell that the attribute can never be overridden.

kpops/components/streams_bootstrap/streams/streams_app.py
disrupted: yeah, unfortunately, otherwise it automatically gets a camelcase alias. Perhaps we could exclude this field from the camelcase generator somehow?

gen_schema.py
raminqaf: Yeah, this is totally fine for now. But I don't think this should be part of `kpops` command. IMO the user should have the generated schema and import it. In other words, we should create a CI job that runs after each PR and updates the schemas. In that way, we have the most updated schemas, and the users can use them.

gen_schema.py
raminqaf: Yes go for it 🚀

kpops/components/streams_bootstrap/streams/streams_app.py
raminqaf: Yeah, if that is not complicated 😄 because this is, at first glimpse, a bit confusing. Otherwise, it is fine.

gen_schema.py
disrupted: Yes, sure, I agree we should provide a default schema. Exposing this as a command would have the advantage that the user can generate their own schema including all the custom components inside their `COMPONENTS_MODULE`. I think that would be the next logical step and improve the user experience. Otherwise they will likely get validation errors when they use custom components in their `pipeline.yaml`.

sujuka99 (approved):

gen_schema.py
raminqaf: Agree! Please provide a follow-up issue on this!

raminqaf (dismissed):

gen_schema.py
disrupted: I've added both of them. Mypy is unfortunately complaining about the use of non-overlapping `Literal` for the `schema_type`, so I had to manually silence these errors

```
Incompatible types in assignment (expression has type "Literal['streams-app']",
base class "KafkaApp" defined the type as "Literal['kafka-app']")  [assignment]
```

raminqaf (approved):

disrupted:
> Just a small comment, if you generate a schema from `PipelineConfig`, we could have autocompletion for `config.yaml` as well. It was much simpler and worked out of the box. I did it by accident while exploring.
> 
> Simply `{PipelineConfig.schema_json(indent=2)` generated `pipeline_config.json` in [this PR](https://github.com/bakdata/kpops/pull/67/files#diff-55d4aac95c43ee8c92de827972ed5fd3c002f04e66c0817c1f726d1aec1ce346).
> 
> I see no downsides of having auto completion for that file as well, but I don't know if it should be added in this PR or a subsequent one.

Good point! I've added it.

gen_schema.py
disrupted: I've opened #71 and #72 for this

sujuka99 (approved):
LGTM

PR #88, opened 2023-02-02
Refactor change calculation

### LINKED ISSUE #68: No diff shown if release does not exist

disrupted:


kpops/utils/dict_differ.py
```diff
@@ -16,9 +16,9 @@ class DiffType(str, Enum):
     REMOVE = "remove"
 
 
-class Change(BaseModel):
-    old_value: Any | None
-    new_value: Any | None
+class Change(NamedTuple):
+    old_value: Any
+    new_value: Any
```
disrupted, line 21: I am a bit puzzled why this was introduced. A type annotation of `Any` is worthless to us, as it hides any meaningful information we have on the value and makes it more error-prone and complex to narrow. Even `Unknown` is usually better because the type checker can infer the type.

kpops/utils/dict_differ.py
```diff
@@ -16,9 +16,9 @@ class DiffType(str, Enum):
     REMOVE = "remove"
 
 
-class Change(BaseModel):
-    old_value: Any | None
-    new_value: Any | None
```
disrupted, line 21: I am a bit puzzled why this was introduced. A type annotation of `Any` is worthless to us, as it hides any meaningful information we have on the value and makes it more error-prone and complex to narrow. Even `Unknown` is usually better because the type checker can infer the type.

philipp94831 (changes requested):
Please add a test

kpops/component_handlers/helm_wrapper/helm_diff.py
```diff
@@ -25,26 +26,23 @@ def get_diff(
...
         for new_resource in new_release_index.values():
-            changes.append(Change(old_value={}, new_value=new_resource.template))
+            changes.append(Change({}, new_resource.template))
 
         return changes
 
-    def log_helm_diff(self, changes: list[Change], logger: logging.Logger):
+    def log_helm_diff(self, changes: Sequence[Change], logger: logging.Logger):
         for change in changes:
-            if change.old_value and change.new_value:
```
disrupted, line 44: this line was the cause for the issue, rest of PR is refactoring and fixing of example pipeline

kpops/utils/dict_differ.py
philipp94831: @raminqaf can explain. I think the return value of the diff package can have different types

raminqaf: We should then change it to an `Uknown` I did in that way because the diff was returning various types back then

kpops/component_handlers/helm_wrapper/helm_diff.py
```diff
@@ -25,26 +26,25 @@ def get_diff(
...
-                if diff := render_diff(
-                    change.old_value,
-                    change.new_value,
-                    ignore=self.config.ignore,
-                ):
-                    logger.info("\n" + diff)
+            if diff := render_diff(
+                *change,
```
philipp94831, line 47: Not a fan of using `*` here. If we change the Change class, this will break

kpops/component_handlers/helm_wrapper/helm_diff.py
```diff
@@ -8,43 +9,41 @@
...
 
 class HelmDiff:
-    def __init__(self, config: HelmDiffConfig):
-        self.config = config
+    def __init__(self, config: HelmDiffConfig) -> None:
+        self.config: HelmDiffConfig = config
 
     @staticmethod
-    def get_diff(
+    def calc_changes(
```
philipp94831, line 16:
```suggestion
    def calculate_changes(
```

kpops/utils/dict_differ.py
```diff
@@ -15,16 +17,63 @@ class DiffType(str, Enum):
...
+
+
+T = TypeVar("T")
 
-class Change(BaseModel):
-    old_value: Any | None
-    new_value: Any | None
 
+@dataclass
+class Change(Generic[T]):  # Generic NamedTuple requires Python 3.11+
```
philipp94831, line 29: why do we need to make it generic? We only use it with a fixed type T

philipp94831 (changes requested):
Which test represents the original issue (missing diff log when release does not exist?

kpops/utils/dict_differ.py
disrupted: `T` is used as a `TypeVar` here to avoid `Any` for the value type. I've changed it from https://github.com/bakdata/kpops/pull/88#discussion_r1094407585 so that Mypy can do its type checking.
Does that answer your question? I wasn't sure what you mean exactly

disrupted:
> Which test represents the original issue (missing diff log when release does not exist?

[`24a4f51` (#88)](https://github.com/bakdata/kpops/pull/88/commits/24a4f51e341f4c62ad81bf2eed0b46efda7b91c9)
the test existed already but didn't really do much previously

kpops/utils/dict_differ.py
philipp94831: Why don't we use unknown or dict? as far as i have seen, dict is always the tye of the generic

philipp94831:
> > Which test represents the original issue (missing diff log when release does not exist?
> 
> [`24a4f51` (#88)](https://github.com/bakdata/kpops/pull/88/commits/24a4f51e341f4c62ad81bf2eed0b46efda7b91c9) the test existed already but didn't really do much previously

So the relevant part of the test is https://github.com/bakdata/kpops/pull/88/commits/24a4f51e341f4c62ad81bf2eed0b46efda7b91c9#diff-d5d41eca49984737405a5547f2acbe18857038a53a83777b0aa87b0ebb862d56R171?

disrupted:
yep, that's right

kpops/utils/dict_differ.py
disrupted: `test_diff.py` uses `dict`, `int`, and `None` https://github.com/bakdata/kpops/blob/e1c7de6667baf8ef09d911dcf050c3df43695111/tests/utils/test_diff.py#L120

disrupted: based on the existing tests and your answers in https://github.com/bakdata/kpops/pull/88#discussion_r1094407585 I am expecting the values to have various possible types.

philipp94831: Yeah but in the prod code it's always dict, right?

disrupted: I don't have an answer for that. The Helm diff only uses `dict`. But this class was added for the topic config diff and there I don't know the possible types. @raminqaf answered that he has seen various different types

philipp94831: Pretty sure that the Helm diff also uses different types. So I am not even sure if dict is correct. There was a reason that it was `any` before

raminqaf:
@disrupted This all the return types of the `diff` function:
```python
def diff(first: Any,
         second: Any,
         node: Any = None,
         ignore: Any = None,
         path_limit: Any = None,
         expand: bool = False,
         tolerance: float = EPSILON,
         absolute_tolerance: Any = None,
         dot_notation: bool = True) -> Generator[tuple[str, str | list, list[tuple[int, AbstractSet]]] | tuple[str, list, tuple[MutableSequence, MutableSequence]] | tuple[str, str | list, list[tuple[Any, Any]]] | tuple[str, str | list, list[tuple[Any, MutableSequence] | tuple[Any, Any]]] | tuple[str, str | list, list[tuple[Any, MutableSequence]]] | tuple[str, str | list, tuple[MutableMapping | MutableSequence, MutableMapping | MutableSequence]], Any, None]
```

tests/components/test_kafka_sink_connector.py
```diff
@@ -331,64 +327,58 @@ def test_clean(
...
-                mocker.call.helm.add_repo(
-                    "bakdata-kafka-connect-resetter",
-                    "https://bakdata.github.io/kafka-connect-resetter/",
-                    RepoAuthFlags(),
-                ),
-                mocker.call.helm.uninstall(
-                    namespace="test-namespace",
-                    release_name=CONNECTOR_CLEAN_NAME,
-                    dry_run=True,
+        assert log_info_mock.mock_calls == [
```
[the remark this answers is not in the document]
disrupted: one of the niceties of pytest :)

kpops/utils/dict_differ.py
raminqaf: If this requires python 3.11+ so we cannot support 3.10 anymore?

disrupted: generic dataclasses are supported on Python 3.10. only generic namedtuples (which would make sense to use) require 3.11+

raminqaf (approved):

raminqaf (approved):

philipp94831 (approved):

PR #95, opened 2023-02-06
Fix ATM fraud example

### LINKED ISSUE #89: Update the ATM example to use an up-to-date configuration.
The example ATM pipeline's `config.yaml` is broken due to the use of at least 1 key that is deprecated. 
```
kafka_connect_resetter_config:
    namespace: "kpops"
```

Currently the example doesn't work as is, update it so that it does.

disrupted:
Fixes #89 
due to recent changes, this was no longer working

### DIFF
--- a/tests/pipeline/test_pipeline.py
+++ b/tests/pipeline/test_pipeline.py
@@ -287,3 +287,29 @@ class TestPipeline:
         assert error_topic == "resources-custom-config-app2-error"
 
         snapshot.assert_match(enriched_pipeline, "test-pipeline")
+
+    def test_atm_fraud_example(self, tmpdir, snapshot):
+        os.environ["KPOPS_ENVIRONMENT"] = "development"
+        output_file_path = tmpdir.join("pipeline.yaml")
+        result = runner.invoke(
+            app,
+            [
+                "generate",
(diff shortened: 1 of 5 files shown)


philipp94831 (commented):
Should we add a workflow that at least runs generate on this pipeline in order to test it?

raminqaf:
fixes https://github.com/bakdata/kpops/issues/89?

philipp94831 (approved):

tests/pipeline/test_pipeline.py
```diff
@@ -287,3 +287,29 @@ def test_default_config(self, tmpdir, snapshot):
         assert error_topic == "resources-custom-config-app2-error"
 
         snapshot.assert_match(enriched_pipeline, "test-pipeline")
+
+    def test_atm_fraud_example(self, tmpdir, snapshot):
```
raminqaf, line 291: Should we move this test to its own file? like `test_atm_fraud_pipelie.py` So we have a new snapshot from it?
I have this feeling that the massive snapshot is hard to debug and understand sometimes.

disrupted: good idea, can I do it in the followup #96? I have already refactored the test fixtures there

PR #96, opened 2023-02-06
Refactor integration test

disrupted:
- extract example to separate snapshot
- prefer Pytest fixtures over unittest
- remove `--save` flag from generate command, print YAML by default


kpops/cli/main.py
```diff
@@ -213,13 +209,6 @@ def generate(
     if print_yaml:
```
raminqaf, line 209: I think we can remove this one too 😉

tests/pipeline/test_pipeline.py
```diff
@@ -35,12 +41,11 @@ def test_load_pipeline(self, tmpdir, snapshot):
 
         assert result.exit_code == 0
 
-        enriched_pipeline = load_yaml_file(Path(output_file_path))
+        enriched_pipeline = load_yaml(result.stdout)
```
raminqaf, line 44: Wouldn't a `yaml.safe_load(resutl.stdout)` be sufficient here?

kpops/pipeline_generator/pipeline.py
```diff
@@ -248,11 +248,9 @@ def print_yaml(self, substitution: dict | None = None) -> None:
         syntax = Syntax(
             substitute(str(self), substitution), "yaml", background_color="default"
```
raminqaf, line 249: Don't we need the `word_wap=True` anymore? 🤔

disrupted: while word wrap is neat for narrow terminals, the output is then no longer valid YAML. (for deserialization)
That's why I had to change it.

raminqaf: Oh, I thought we must set that.  I thought the `Console(width=1000)` was ignored by the `CliRunner()`

tests/pipeline/test_pipeline.py
disrupted: yep, good point!

kpops/pipeline_generator/pipeline.py
disrupted: It is not ignored, we just unfortunately can't set the terminal size for the `CliRunner` in the tests, which would make more sense imo. At least I didn't find a way to do so, I tried using click click `context_settings` on the `Typer` app and `$TERMINAL_SIZE` environment variable.

tests/pipeline/test_pipeline.py
```diff
@@ -13,9 +16,15 @@
...
 class TestPipeline:
-    def test_load_pipeline(self, tmpdir, snapshot):
+    @pytest.fixture(autouse=True)
+    def setup_teardown(self) -> Iterator[None]:
+        # setup
         os.environ["KPOPS_ENVIRONMENT"] = "development"
-        output_file_path = tmpdir.join("pipeline.yaml")
+        yield  # testing
+        # teardown
+        del os.environ["KPOPS_ENVIRONMENT"]
```
raminqaf, line 25: This is a bit strange. After the discussion [here](https://github.com/bakdata/kpops/pull/100#discussion_r1099787094). I have this feeling that all the values are read from the [config.yaml](https://github.com/bakdata/kpops/blob/main/config.yaml#L2) in the root directory.
IMO we don't need this at all since the mandatory `broker` address set is also not set by an env, but the tests run green. WDYT?

disrupted: true! `--config` already has a default set, so this is unnecessary

raminqaf: Kill it with kindness hehehe

raminqaf (approved):
LGTM

raminqaf:
@sujuka99 Please have look too!

sujuka99 (approved):
LGTM!

PR #100, opened 2023-02-07
Clean up dry-run logging

### LINKED ISSUE #87: Recolour the log messages
Currently, some log message colors do not indicate a meaningful sign. @philipp94831 has a good overview of that and would add his feedback.

philipp94831:
Fixes #87

### DIFF
--- a/kpops/component_handlers/topic/handler.py
+++ b/kpops/component_handlers/topic/handler.py
@@ -153,7 +153,7 @@ class TopicHandler:
             log.error(
                 f"Topic Creation: partition count of topic {topic_name} changed! Partitions count of topic {topic_name} is {partition_count}. The given partitions count {topic_spec.partitions_count}."
             )
-            exit(1)  # TODO shouldn't we raise instead of exiting?
+            exit(1)  # FIXME raise instead https://github.com/bakdata/kpops/issues/101
 
     @staticmethod
     def __check_replication_factor(
@@ -174,7 +174,7 @@ class TopicHandler:
             log.error(
                 f"Topic Creation: replication factor of topic {topic_name} changed! Replication factor of topic {topic_name} is {replication_factor}. The given replication count {topic_spec.replication_factor}."
             )
-            exit(1)  # TODO shouldn't we raise instead of exiting?
+            exit(1)  # FIXME raise instead https://github.com/bakdata/kpops/issues/101
 
     def __dry_run_topic_deletion(self, topic_name: str) -> None:
         try:
--- a/kpops/component_handlers/topic/handler.py
+++ b/kpops/component_handlers/topic/handler.py
@@ -146,16 +146,14 @@ class TopicHandler:
         if partition_count == (
             topic_spec.partitions_count or int(broker_config["num.partitions"])
         ):
-            log.info(
-                yellowify(
+            log.debug(
                     f"Topic Creation: partition count of topic {topic_name} did not change. Current partitions count {partition_count}. Updating configs."
-                )
             )
         else:
             log.error(
                 f"Topic Creation: partition count of topic {topic_name} changed! Partitions count of topic {topic_name} is {partition_count}. The given partitions count {topic_spec.partitions_count}."
             )
-            exit(1)
+            exit(1)  # TODO shouldn't we raise instead of exiting?
 
     @staticmethod
     def __check_replication_factor(
@@ -169,16 +167,14 @@ class TopicHandler:
             topic_spec.replication_factor
             or int(broker_config["default.replication.factor"])
         ):
-            log.info(
-                yellowify(
+            log.debug(
                     f"Topic Creation: replication factor of topic {topic_name} did not change. Current replication factor {replication_factor}. Updating configs."
-                )
             )
         else:
             log.error(
                 f"Topic Creation: replication factor of topic {topic_name} changed! Replication factor of topic {topic_name} is {replication_factor}. The given replication count {topic_spec.replication_factor}."
             )
-            exit(1)
+            exit(1)  # TODO shouldn't we raise instead of exiting?
 
     def __dry_run_topic_deletion(self, topic_name: str) -> None:
         try:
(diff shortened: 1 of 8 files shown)


kpops/component_handlers/topic/handler.py
```diff
@@ -146,16 +146,14 @@ def __check_partition_count(
...
-                )
+            log.debug(
+                f"Topic Creation: partition count of topic {topic_name} did not change. Current partitions count {partition_count}. Updating configs."
             )
         else:
             log.error(
                 f"Topic Creation: partition count of topic {topic_name} changed! Partitions count of topic {topic_name} is {partition_count}. The given partitions count {topic_spec.partitions_count}."
             )
-            exit(1)
+            exit(1)  # TODO shouldn't we raise instead of exiting?
```
philipp94831, line 156: @raminqaf @disrupted Any reason why we exit and don't raise? To me this looks strange especially as we don't have any way to handle it

tests/components/test_kubernetes_app.py
```diff
@@ -33,6 +33,7 @@ def config(self) -> PipelineConfig:
             defaults_path=DEFAULTS_PATH,
             environment="development",
             helm_diff_config=HelmDiffConfig(enable=True),
+            broker="",  # This is missing
```
philipp94831, line 36: @raminqaf @disrupted PyCharm was complaining that broker is not set. Field default is `...`. Do you know why it works in CI? Can also be because I am on Windows and there are problems with environment variable config

disrupted: a lot of code analysis tools have problems with Pydantic, especially with off-spec stuff that works differently to the `dataclass_transforms` stuff.

Usually the workaround is to change the definition from `Field(...)` to `Field(default=...)` but I think that's already the case here.

kpops/component_handlers/topic/handler.py
disrupted: I don't know the reason, but I agree that raising an exception would make sense

tests/components/test_kubernetes_app.py
philipp94831: It's not code analysis but when I run Pytest via PyCharm it just aborts

kpops/component_handlers/topic/handler.py
philipp94831: We can also do that outside this PR, I just stumbled upon it

tests/components/test_kubernetes_app.py
disrupted: ah, I don't know then.

philipp94831: But as far as I understand, broker is a required Field. Obviously it was not set in this constructor. How is it then set in tests? Is there an environment variable defined somewhere?

kpops/component_handlers/topic/handler.py
raminqaf: Agree we should raise an Exception. Will you create an issue for it please?

philipp94831: #101

philipp94831: Should I retain the TODOs in the code (and rephrase them a little)?

tests/components/test_kubernetes_app.py
raminqaf: `default =...` means it is mandatory. But I'd like to know how this was not broken before. 😅 
I think the value is read from the `config.yaml` file in the root

raminqaf: From here:
https://github.com/bakdata/kpops/blob/main/config.yaml#L2

kpops/components/base_components/kubernetes_app.py
```diff
@@ -100,7 +100,13 @@ def to_helm_values(self) -> dict:
...
 
     def print_helm_diff(self, stdout: str) -> None:
-        current_release = self.helm.get_manifest(self.helm_release_name, self.namespace)
+        current_release = list(
+            self.helm.get_manifest(self.helm_release_name, self.namespace)
+        )
+        if current_release:
+            log.info(f"Helm release {self.helm_release_name} already exists")
+        else:
+            log.info(f"Helm release {self.helm_release_name} does not exist")
```
raminqaf, line 109: I don't know if you are a fan of this syntax but:
```python
log.info(f"Helm release {self.helm_release_name} already exists") if current_release else log.info(f"Helm release {self.helm_release_name} does not exist")
```

tests/components/test_kubernetes_app.py
philipp94831: Ah then it is just a problem with the base dir in my PyCharm

kpops/components/base_components/kubernetes_app.py
philipp94831: Show me the Python way

philipp94831: But this gets formatted to
```
log.info(
            f"Helm release {self.helm_release_name} already exists"
        ) if current_release else log.info(
            f"Helm release {self.helm_release_name} does not exist"
        )
```
so has even more lines

raminqaf: @disrupted I would like to know your opinion on this 🐍

raminqaf: Okay then lets revert it

raminqaf (approved):
LGTM

kpops/components/base_components/kubernetes_app.py
disrupted: I am fine with the original. The only shorter option I can think of would be

```python
log.info(f"Helm release {self.helm_release_name} {'already exists' if current_release else 'does not exist'}")
```

raminqaf: Let's stick with the good old if/else statement

disrupted (approved):

PR #105, opened 2023-02-09
Fix incomplete inflate component

### LINKED ISSUE #106: Inflate creates invalid components
Components added during inflate drop their component `type` and are therefore invalid.

disrupted:
Fixes #106


philipp94831 (changes requested):
Please add a test

disrupted:
> Please add a test

it's covered by the snapshot

philipp94831:
> > Please add a test
> 
> it's covered by the snapshot

This one? https://github.com/bakdata/kpops/pull/105/files#diff-188124f32b96fef3236bd30f0f27057ca6f48ea69b80e09bfcb85f09c6d689dbR531

philipp94831:
> > This one? [#105 (files)](https://github.com/bakdata/kpops/pull/105/files#diff-188124f32b96fef3236bd30f0f27057ca6f48ea69b80e09bfcb85f09c6d689dbR531)
> 
> yes, specifically this line https://github.com/bakdata/kpops/pull/105/files#diff-188124f32b96fef3236bd30f0f27057ca6f48ea69b80e09bfcb85f09c6d689dbR537
> 
> should I still add another assertion to check that every inflated component in the output of `generate` has a type?

More tests always sounds good. But if it is covered by the test, I am also fine

philipp94831 (dismissed):

kpops/components/base_components/kafka_connect.py
```diff
@@ -33,7 +33,7 @@
 
 
 class KafkaConnector(PipelineComponent, ABC):
-    type: ClassVar[str] = "kafka-connect"
+    type = "kafka-connect"
```
raminqaf, line 36: The type `str` is not needed to be specified here.

disrupted: it's not, though? I don't understand if you are asking it as a question or making a statement 😄

raminqaf (approved):

kpops/pipeline_generator/pipeline.py
```diff
@@ -230,7 +230,7 @@ def substitute_component_specific_variables(
     def __str__(self) -> str:
         return yaml.dump(
             PipelineComponents(components=self.components).dict(
-                exclude_unset=True, exclude_none=True, by_alias=True
```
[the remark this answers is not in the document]
disrupted: this was causing an the faulty behavior that when we instantiate a component without explicitly passing `type=<component-type>` to the constructor, during serialization (output of `kpops generate`) the component wouldn't include `type` anymore, even though it is a constant attribute with default value.

as for its intended use, I am not sure why it was needed in the first place. The integration test snapshot confirm that it is unchanged. Only the inflated component (which was invalid before) is now complete.

PR #118, opened 2023-02-16
WIP fix substitution

disrupted:


kpops/components/base_components/kubernetes_app.py
```diff
@@ -131,3 +131,11 @@ def __check_compatible_name(self) -> None:
             raise ValueError(
                 f"The component name {self.name} is invalid for Kubernetes."
             )
+
+    def dict(self, *, exclude=None, **kwargs) -> dict[str, Any]:
+        # HACK: workaround for Pydantic to exclude cached properties during model export
```
philipp94831, line 136: Not sure if this is really a Pydantic issue but rather we are mixing (cached state) with actual properties. So we should (not in this PR) rather refactor that a bit I guess

philipp94831 (approved):

kpops/components/base_components/kubernetes_app.py
disrupted: yeah, I agree this could use some improving.
Pydantic does however introduce weird behavior here, as it always disregards `property` during serialization, but not `cached_property` 🤷🏻‍♂️

philipp94831 (approved):

Issue #74, opened 2023-01-30
Add titles and descriptions to Pydantic model fields


disrupted:
Having meaningful descriptions for all `PipelineComponent` models and their fields would improve both schema generation and documentation.
Additionally, we should fix some of the type constraints, e.g. the `TopicConfig.configs` dict should also allow other types than `str`, such as `int`

PR #122, opened 2023-02-21
Update streams-bootstrap autoscaling config

disrupted:
Related to #74

### DIFF
--- a/kpops/components/streams_bootstrap/streams/model.py
+++ b/kpops/components/streams_bootstrap/streams/model.py
@@ -53,7 +53,7 @@ class StreamsConfig(KafkaStreamsConfig):
 
 
 class StreamsAppAutoScaling(BaseModel):
-    consumergroup: str
+    consumer_group: str
     lag_threshold: int
     topics: list[str] = []
 
--- a/kpops/components/streams_bootstrap/streams/model.py
+++ b/kpops/components/streams_bootstrap/streams/model.py
@@ -5,6 +5,7 @@ from kpops.components.base_components.kafka_app import (
     KafkaAppConfig,
     KafkaStreamsConfig,
 )
+from kpops.utils.pydantic import CamelCaseConfig
 
 
 class StreamsConfig(KafkaStreamsConfig):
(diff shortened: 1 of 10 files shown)


kpops/components/streams_bootstrap/streams/model.py
```diff
@@ -52,10 +53,23 @@ def dict(
...
+        description="Name of the consumer group used for checking the offset on the topic and processing the related lag.",
+    )
+    lag_threshold: int = Field(
+        title="Lag threshold",
+        description="Average target value to trigger scaling actions.",
+    )
+    topics: list[str] = Field(
+        default=[],
+        description="List of auto-generated Kafka Streams topics used by the streams app.",
+    )
```
disrupted, line 70: should we add the other attributes as well?
https://github.com/bakdata/streams-bootstrap/tree/master/charts/streams-app

philipp94831: Sounds good. I don't think anything will break. Only down-side is that KPOps could become harder to use with older/newer chart versions if we are too strict

kpops/components/streams_bootstrap/streams/model.py
```diff
@@ -52,10 +53,68 @@ def dict(
...
+    offset_reset_policy: str = Field(
+        default="earliest",
+        title="Offset reset policy",
+        description="The offset reset policy for the consumer if the the consumer group is not yet subscribed to a partition.",
+    )
+    min_replicas: int = Field(
+        default=0,
+        title="Min replica count",
+        description=(
+            "Minimum number of replicas KEDA will scale the resource down to. "
```
yannick-roeder, line 92:
```suggestion
            "Minimum number of replicas KEDA will scale the resource down to."
```

kpops/components/streams_bootstrap/streams/model.py
```diff
@@ -52,10 +53,68 @@ def dict(
...
+        description=(
+            "This is the interval to check each trigger on. "
+            "https://keda.sh/docs/2.7/concepts/scaling-deployments/#pollinginterval"
+        ),
+    )
+    cooldown_period: int = Field(
+        default=300,
+        title="Cooldown period",
+        description=(
+            "The period to wait after the last trigger reported active before scaling the resource back to 0. "
```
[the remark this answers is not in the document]
disrupted: that's on purpose for the concatenation in this multi-line string. From those descriptions the pipeline.yaml schema is generated

sujuka99 (approved):
LGTM, but pls wait for one more

raminqaf (approved):
LGTM

yannick-roeder (approved):

PR #125, opened 2023-02-22
Skip FromSection for producers

disrupted:
Producer apps don't support FromSection

### DIFF
--- a/kpops/components/streams_bootstrap/producer/producer_app.py
+++ b/kpops/components/streams_bootstrap/producer/producer_app.py
@@ -26,6 +26,7 @@ class ProducerApp(KafkaApp):
         default="producer", exclude=True
     )
     app: ProducerValues
+    from_: None = None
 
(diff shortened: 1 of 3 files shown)


philipp94831 (approved):

kpops/components/streams_bootstrap/producer/producer_app.py
```diff
@@ -26,6 +26,7 @@ class ProducerApp(KafkaApp):
         default="producer", exclude=True
     )
     app: ProducerValues
+    from_: None = None
```
raminqaf, line 29: Shouldn't that be of type `FromSection`?

disrupted: we want to disallow setting a `FromSection` for producers, because it doesn't support it

raminqaf (approved):

raminqaf (approved):

PR #130, opened 2023-02-24
Helm diff should not render NOTES.txt

### LINKED ISSUE #54: Helm diff should not render NOTES.txt
Helm Charts can contain a [NOTES.txt file](https://helm.sh/docs/chart_template_guide/notes_files/). When creating the diff, these notes are present because they appear in the dry-run upgrade but not in the manifest. This leads to incorrect diffs. We should exclude the notes from the diff

sujuka99:
Closes #54

### DIFF
--- a/kpops/component_handlers/helm_wrapper/model.py
+++ b/kpops/component_handlers/helm_wrapper/model.py
@@ -87,6 +87,7 @@ class YamlReader:
     def __iter__(self) -> Iterator[str]:
         # discard all output before template documents
         index = self.content.index("---")
-        self.content = self.content[index:-1]
+        rindex = self.content.rindex("NOTES:")
+        self.content = self.content[index:rindex]
         yield from self.content.splitlines()
         yield "---"  # add final divider to make parsing easier
--- a/kpops/component_handlers/helm_wrapper/model.py
+++ b/kpops/component_handlers/helm_wrapper/model.py
@@ -87,9 +87,9 @@ class YamlReader:
     def __iter__(self) -> Iterator[str]:
         # discard all output before template documents
         index = self.content.index("---")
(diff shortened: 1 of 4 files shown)


kpops/component_handlers/helm_wrapper/model.py
```diff
@@ -87,6 +87,11 @@ class YamlReader:
     def __iter__(self) -> Iterator[str]:
         # discard all output before template documents
         index = self.content.index("---")
-        self.content = self.content[index:-1]
+        notes = "\n\nNOTES:\n"
```
disrupted, line 90: how about making this a constant variable (uppercase) and defining it outside the method. Since it is not changing it doesn't have to live inside the method

disrupted (commented):
looks good, just a minor note

kpops/component_handlers/helm_wrapper/model.py
```diff
@@ -87,6 +87,11 @@ class YamlReader:
...
+        if notes in self.content:
+            rindex = self.content.index(notes)
```
disrupted, line 92: perhaps we should rename the variables `index` and `rindex` as the name is a bit misleading now. since `rindex` refers to the last occurrence
how about `start` & `end`?

kpops/component_handlers/helm_wrapper/model.py
sujuka99: Makes sense, done

kpops/component_handlers/helm_wrapper/model.py
sujuka99: agreed

disrupted (approved):
let's wait with merging until we have tested it to confirm it's working

PR #133, opened 2023-02-27
Read from specific pipeline components

disrupted:


tests/pipeline/resources/read-from-component/pipeline.yaml
```diff
@@ -0,0 +1,16 @@
...
+      ${output_topic_name}:
+        type: output
+
+- type: streams-app
+  name: consumer1
+
+- type: streams-app
+  name: consumer2
+  from:
+    topics: {}
```
philipp94831, line 14: What if I omit `topics`?

disrupted: `topics` can't be omitted, this would give a validation error

philipp94831: But we could make it an empty map by default?

disrupted: yes

disrupted: I've changed it

kpops/components/base_components/models/from_section.py
```diff
@@ -36,7 +36,8 @@ def extra_topic_role(cls, values):
 
 
 class FromSection(BaseModel):
-    topics: dict[str, FromTopic]
+    topics: dict[str, FromTopic] = {}
+    components: list[str] = []
```
sujuka99, line 40: Is it possible to add descriptions here?
Something like
```
topics: dict[str, FromTopic] = Field(
	default={},
	description="Topics to read from.",
)
components: list[str]  = Field(
	default=[],
	description="Components to read from.",
)
```

docs/docs/resources/pipeline-components/kafka-sink-connector.yaml
```diff
@@ -13,6 +13,8 @@
       ${pipeline_name}-input-topic:
         type: input # required
         # role: topic-role # only used if type is `extra`
+    components: # read from specific component
+      - ${pipeline_name}-producer
   # Topic(s) into which the component will write output
   to:
     topics: # required
```
sujuka99, line 20:
```suggestion
    topics: 
```

docs/docs/resources/pipeline-components/kubernetes-app.yaml
```diff
@@ -14,6 +14,8 @@
       ${pipeline_name}-input-topic:
         type: input # required
         # role: topic-role # only used if type is `extra` or `extra-pattern`
+    components: # read from specific component
+      - ${pipeline_name}-producer
   # Topic(s) into which the component will write output
   to:
     topics: # required
```
sujuka99, line 21:
```suggestion
    topics:
```

docs/docs/resources/pipeline-components/streams-app.yaml
```diff
@@ -62,6 +62,8 @@
       ${pipeline_name}-input-topic:
         type: input # required
         # role: topic-role # only used if type is `extra` or `extra-pattern`
+    components: # read from specific component
+      - ${pipeline_name}-producer
   # Topic(s) into which the component will write output
   to:
     topics: # required
```
sujuka99, line 69:
```suggestion
    topics:
```

sujuka99 (changes requested):

kpops/components/base_components/models/from_section.py
disrupted: good idea! thanks

docs/docs/resources/pipeline-components/kafka-sink-connector.yaml
```diff
@@ -13,9 +13,11 @@
       ${pipeline_name}-input-topic:
         type: input # required
         # role: topic-role # only used if type is `extra`
+    components: # read from specific component
+      - ${pipeline_name}-producer
```
philipp94831, line 17: Why do you need to specify `${pipeline_name}`? Also producer is maybe a bit irritating. Could just be my-first-component or sth like that

tests/pipeline/resources/read-from-component/pipeline.yaml
```diff
@@ -0,0 +1,26 @@
...
+        type: output
+
+- type: streams-app
+  name: consumer1
+
+- type: streams-app
+  name: consumer2
+  from:
+    components:
+      - ${pipeline_name}-producer
```
philipp94831, line 15: This should work without pipeline_name as we can only reference components within our pipeline

philipp94831 (changes requested):

docs/docs/resources/pipeline-components/kafka-sink-connector.yaml
disrupted: `${pipeline_name}` is only necessary if this is the prefix of the component (the default). It simply has to match the full name of the component

tests/pipeline/resources/read-from-component/pipeline.yaml
disrupted: it does if the other component has `prefix: ""`

docs/docs/resources/pipeline-components/kafka-sink-connector.yaml
philipp94831: I know. But it should be possible to reference without prefix

tests/pipeline/resources/read-from-component/pipeline.yaml
disrupted: prefix definition for reference https://github.com/bakdata/kpops/blob/3a4fac55d2e5e0ddffd02a216aad9fda82c90b9f/kpops/components/base_components/pipeline_component.py#L30

docs/docs/resources/pipeline-components/kafka-sink-connector.yaml
philipp94831: Prefix is only used to name components uniquely in Kubernetes and Connect. However, when referencing them, I don't want to deal with prefixes. Therefore, it should not be required to use prefix here.

disrupted: that's not possible. kpops uses both prefix + name to distinguish components https://github.com/bakdata/kpops/blob/3a4fac55d2e5e0ddffd02a216aad9fda82c90b9f/kpops/pipeline_generator/pipeline.py#LL162C13-L162C29

disrupted: please create a new issue for it. At the moment `prefix` is a core attribute of every pipeline component, not something specific to Kubernetes and Kafka Connect.

philipp94831: There is no need for an issue. It is required for read from component

philipp94831: Wherever you index the components by name, you just need to make sure to use the name without the prefix

tests/pipeline/snapshots/snap_test_example.py
```diff
@@ -210,6 +210,7 @@
                 },
             },
             "from": {
+                "components": [],
```
raminqaf, line 213: Should it be excluded from the generated YAML, or is it fine to have it as an empty list?

raminqaf: Can they also both exist? concretely, reading from topic(s) and component(s) at the same time

kpops/pipeline_generator/pipeline.py
```diff
@@ -155,44 +145,29 @@ def parse_components(self, component_list: list[dict]) -> None:
...
             config=self.config,
             handlers=self.handlers,
             **component_data,
         )
         component = self.enrich_component(component)
-        # weave from topics
-        if previous_component and previous_component.to:
-            component.weave_from_topics(previous_component.to)
 
         # inflate & enrich components
```
raminqaf, line 158: Maybe extract in a method and remove the comment :)

sujuka99 (approved):

kpops/pipeline_generator/pipeline.py
disrupted: I sympathize with the motive, in this case I suppose it's actually easier to understand if we leave the logic in `apply_component`, because it goes through several steps (enrich base component, inflate, enrich inflated components, weave from topics, finally add components to component index). otherwise it gets pretty fragmented and hard to name each of the methods :D

tests/pipeline/snapshots/snap_test_example.py
disrupted: I can't think of any downsides other than it being a bit verbose maybe. that being said I don't have a preference, fine with either option 🤷🏻

disrupted: updated the test & snapshot

raminqaf (approved):

PR #147, opened 2023-03-06
Refactor loading of component defaults to independent function

disrupted:
this way it can be imported from the package and used for development purposes


kpops/cli/main.py
```diff
@@ -224,6 +224,8 @@ def generate(
                 '--steps'"
         )
 
+    return pipeline
```
disrupted, line 227: technically not needed anymore, was only used for testing. but perhaps it makes sense to return the generated pipeline anyways instead of only printing to stdout

kpops/components/base_components/base_defaults_component.py
```diff
@@ -58,31 +58,44 @@ def extend_with_defaults(self, kwargs: dict[str, Any]) -> dict:
...
+            config
+        )
+        defaults = load_defaults(
+            self.__class__, main_default_file_path, environment_default_file_path
+        )
+        kwargs = update_nested(kwargs, defaults)
         return kwargs
 
 
+def load_defaults(
```
philipp94831, line 71: Please add a test that makes assertions on the returned dict

philipp94831 (changes requested):

PR #151, opened 2023-03-10
Fix config.yaml `defaults_path` being overridden by CLI

### LINKED ISSUE #144: `config.yaml`'s `defaults_path` setting is always overriden by the CLI.
The config.yaml `defaults_path` setting probably doesn't work correctly right now, because the value is [always overridden](https://github.com/bakdata/kpops/blob/500dceb2eb87fa96529e0ce8ad5ad9deb357062f/kpops/cli/main.py#L182) by the CLI.

Check if both setting `defaults_path` inside of `config.yaml` is working and overriding it using the CLI/environment variable is working correctly. This also should be covered by a test which it probably isn't atm.

Update the documentation accordingly after implementing the fix.

sujuka99:
closes #144


sujuka99 (commented):
Left: fix docs

examples/bakdata/atm-fraud-detection/config.yaml
```diff
@@ -12,4 +12,4 @@ kafka_rest_host: "http://localhost:8082"
 
 kafka_connect_host: "http://localhost:8083"
 
-defaults_path: ./defaults
+defaults_path: examples/bakdata/atm-fraud-detection/defaults
```
disrupted, line 15: it would be better if we could use relative paths here. otherwise, the path depends on where kpops is executed from

kpops/cli/main.py
```diff
@@ -38,7 +38,7 @@
 )
 
 DEFAULT_PATH_OPTION: Path = typer.Option(
-    default=Path("defaults"),
+    default=None,
```
disrupted, line 41: `None` can't be set as a default for `Path` as they are not compatible types

sujuka99: Oh, I didn't know. Strangely it worked as intended when I tried `kpops generate` and all the tests passed. Does that mean that the tests' coverage is not good?

examples/bakdata/atm-fraud-detection/config.yaml
sujuka99: What would the path be relative to?

kpops/cli/main.py
disrupted: Python's type hints are only meant for static analysis and ignored at runtime. Please change `Path` to `Optional[Path]` and update all the references below. I suppose mypy doesn't catch this mistake because it doesn't have knowledge on how typer works

sujuka99: Any idea how I can make a Path() object falsy?

sujuka99: Ah I hadn't refreshed the page and didnt see[ your response](https://github.com/bakdata/kpops/pull/151#discussion_r1135667375)

kpops/cli/main.py
```diff
@@ -175,11 +175,14 @@ def log_action(action: str, pipeline_component: PipelineComponent):
...
-    config: Path, defaults: Path, verbose: bool
+    config: Path, defaults: Optional[Path], verbose: bool
 ) -> PipelineConfig:
     setup_logging_level(verbose)
     PipelineConfig.Config.config_path = config
-    pipeline_config = PipelineConfig(defaults_path=defaults)
+    if defaults:
+        pipeline_config = PipelineConfig(defaults_path=defaults)
+    else:
+        pipeline_config = PipelineConfig()
```
disrupted, line 185: wouldn't it be easier to do "calculate defaults path relative to config.yaml" here? saves the hassle of working with dict keys and such

sujuka99: I first tried that, but for some reason I kept getting errors, I couldn't extract `defaults_path` from the `PipelineConfig` object and decided not to waste more time and to at least get a working solution first. I could try it again after I write the unit test.

kpops/cli/pipeline_config.py
```diff
@@ -30,7 +30,8 @@ class PipelineConfig(BaseSettings):
     """
 
     defaults_path: Path = Field(
-        default=...,
+        default=Path("defaults"),
```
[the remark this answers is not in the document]
sujuka99: It does make things a bit simpler, I would do it, good point!

kpops/cli/pipeline_config.py
```diff
@@ -117,5 +118,6 @@ def customise_sources(
 def yaml_config_settings_source(settings: PipelineConfig) -> dict | list:
     path_to_config = settings.Config.config_path
     if path_to_config.exists():
-        return load_yaml_file(path_to_config)
+        config = load_yaml_file(path_to_config)
+        return config
```
disrupted, line 122: you can revert this now

disrupted (dismissed):

kpops/cli/pipeline_config.py
sujuka99: done

tests/pipeline/test_pipeline.py
```diff
@@ -235,6 +236,53 @@ def test_with_custom_config(self, snapshot: SnapshotTest):
...
+
+        streams_app_details = enriched_pipeline["components"][1]
+        output_topic = streams_app_details["app"]["streams"]["outputTopic"]
+        assert output_topic == "app2-test-topic"
+        error_topic = streams_app_details["app"]["streams"]["errorTopic"]
+        assert error_topic == "app2-dead-letter-topic"
+
+        snapshot.assert_match(enriched_pipeline, "test-pipeline")
+        abs_config_yaml.close()
+        os.remove(str(RESOURCE_PATH / "custom-config/temp_config.yaml"))
```
disrupted, line 284: good effort! while this works, I believe we could clean it up a little. For example, we'd always want to use a context manager for `open()` and avoid closing resources manually.

Also we wanna make sure to always delete the temp file, even in case the test fails. please use try-finally for the file deletion.

disrupted (changes requested):

tests/pipeline/test_pipeline.py
```diff
@@ -235,6 +236,53 @@ def test_with_custom_config(self, snapshot: SnapshotTest):
...
+        yaml.dump(config_dict, abs_config_yaml)
+        result = runner.invoke(
+            app,
+            [
+                "generate",
+                "--pipeline-base-dir",
+                PIPELINE_BASE_DIR,
+                str(RESOURCE_PATH / "custom-config/pipeline.yaml"),
+                "--config",
+                str(RESOURCE_PATH / "custom-config/temp_config.yaml"),
```
disrupted, line 262: you're declaring this path a few times. Please create a variable for it, e.g. `abs_config_path`. Instead of `os.remove` you can then directly do `abs_config_path.unlink()`

tests/pipeline/test_pipeline.py
```diff
@@ -235,6 +236,53 @@ def test_with_custom_config(self, snapshot: SnapshotTest):
 
         snapshot.assert_match(enriched_pipeline, "test-pipeline")
 
+    def test_with_custom_config_with_absolute_defaults_path(
+        self, snapshot: SnapshotTest
+    ):
+        with open(
+            str(RESOURCE_PATH / "custom-config/config.yaml"), "r"
```
disrupted, line 243: casting to `str` is unnecessary, `open` supports `Path` instances :)

kpops/cli/pipeline_config.py
```diff
@@ -30,7 +30,8 @@ class PipelineConfig(BaseSettings):
     """
 
     defaults_path: Path = Field(
-        default=...,
+        default=Path("."),
+        example="defaults",
```
raminqaf, line 34: We should add in the description what the default value is. In other words, what is the behaviour if the defaults_path is not set

tests/pipeline/test_pipeline.py
sujuka99: Thanks for pointing those out!

I was a bit hesitant to put the whole test in a `with` statement, although I didn't have a particular reason for that. Is it correct that I included everything after creating the temp file in the `try`?

tests/pipeline/test_pipeline.py
sujuka99: Good point, done

DIdn't know about `unlink()`, very nice! No need to import `os` :)

tests/pipeline/test_pipeline.py
sujuka99: Cool, removed the unnecessary step

tests/pipeline/test_pipeline.py
```diff
@@ -247,41 +246,41 @@ def test_with_custom_config_with_absolute_defaults_path(
...
+
+                streams_app_details = enriched_pipeline["components"][1]
+                output_topic = streams_app_details["app"]["streams"]["outputTopic"]
+                assert output_topic == "app2-test-topic"
+                error_topic = streams_app_details["app"]["streams"]["errorTopic"]
+                assert error_topic == "app2-dead-letter-topic"
+
+                snapshot.assert_match(enriched_pipeline, "test-pipeline")
+            finally:
+                temp_config_path.unlink()
```
raminqaf, line 283: Do you realy need a `try/finally` block here? can't you just call the unlink after the assert_match method?

sujuka99: https://github.com/bakdata/kpops/pull/151#discussion_r1139163873

sujuka99: I thought the same, but I didn't consider that the file would not be deleted under some circumstances. I would probably prefer to err on the safe side, who knows what might happen. It's a test anyway, it shouldn't impact kpops' performance, right?

raminqaf: Okay that explains and answers my question. All good and fine! Thanks! 😄

raminqaf (approved):

docs/docs/schema/config.json
```diff
@@ -76,6 +76,8 @@
                 "defaults_path": {
                     "title": "Defaults Path",
                     "description": "The path to the folder containing the defaults.yaml file and the environment defaults files.",
```
disrupted, line 78: please update the schema with the new description

tests/pipeline/test_pipeline.py
```diff
@@ -235,6 +235,51 @@ def test_with_custom_config(self, snapshot: SnapshotTest):
...
         snapshot.assert_match(enriched_pipeline, "test-pipeline")
 
+    def test_with_custom_config_with_absolute_defaults_path(
+        self, snapshot: SnapshotTest
+    ):
+        with open(RESOURCE_PATH / "custom-config/config.yaml", "r") as rel_config_yaml:
+            config_dict = yaml.safe_load(rel_config_yaml)
+        defaults_path = (
+            RESOURCE_PATH / "custom-config" / Path(config_dict["defaults_path"])
+        )
```
disrupted, line 245: let's just use the folder name directly
```suggestion
        defaults_path = (RESOURCE_PATH / "no-topics-defaults").absolute()
```

disrupted (dismissed):

raminqaf (approved):

tests/pipeline/test_pipeline.py
disrupted: I would swap `try` and the `with` block
then you can remove one indentation level of everything below the `with open` context manager

tests/pipeline/test_pipeline.py
```diff
@@ -235,6 +235,50 @@ def test_with_custom_config(self, snapshot: SnapshotTest):
...
+        with open(RESOURCE_PATH / "custom-config/config.yaml", "r") as rel_config_yaml:
+            config_dict = yaml.safe_load(rel_config_yaml)
+        config_dict["defaults_path"] = str(
+            (RESOURCE_PATH / "no-topics-defaults").absolute()
+        )
+        temp_config_path = RESOURCE_PATH / "custom-config/temp_config.yaml"
+        try:
+            with open(temp_config_path, "w") as abs_config_yaml:
+                yaml.dump(config_dict, abs_config_yaml)
+                result = runner.invoke(
```
disrupted, line 250: dedent this line and below. only `yaml.dump` needs to be inside the context manager, i.e. writing to the file. afterwards the `abs_config_yaml` file can be closed

disrupted (changes requested):

disrupted (approved):

PR #163, opened 2023-03-20
fix(README): documentation leads to user-guide

sujuka99:
[Home](https://bakdata.github.io/kpops/1.0/) page is currently empty and ugly. No need to have an extra step before reaching the [user guide](https://bakdata.github.io/kpops/1.0/user/what-is-kpops/).

To be reverted when [Home](https://bakdata.github.io/kpops/latest/) is improved enough.

Opens #164

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
 ## Documentation
 
 For detailed usage and installation instructions, check out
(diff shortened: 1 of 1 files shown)


sujuka99:
Opens #164

disrupted (approved):

PR #165, opened 2023-03-20
Release and publish workflows should be separated from each other

### LINKED ISSUE #85: Release and publish workflows should be separated from each other
Currently, the [release.yaml](https://github.com/bakdata/kpops/blob/main/.github/workflows/release.yaml) file publishes the Python packages to PyPi **and then** creates a git tag, GitHub release, and updates/pushes the docs.
We should change the order and separate the release.yaml into to workflows: release, publish.

## Release Wrokflow
This workflow should be triggered manually, and it is similar to the current [release.yaml](https://github.com/bakdata/kpops/blob/main/.github/workflows/release.yaml) workflow. Except it only creates a git tag and creates a GitHub release

## Publish Workflow
This workflow will be **triggered only if a new tag is pushed**. This workflow will publish the Python packages to PyPi and update/pubish the docs to GitHub pages.

sujuka99:
Closes #85

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,36 @@
+name: Release to PyPi, Update docs
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  release-to-pypi:
+    uses: bakdata/ci-templates/.github/workflows/python-poetry-release.yaml@v1.13.2
+    with:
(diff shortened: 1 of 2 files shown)


sujuka99:
[Reason for closing](https://github.com/bakdata/kpops/issues/85#issuecomment-1476371674)

PR #166, opened 2023-03-22
Make schema generation a builtin CLI command

### LINKED ISSUE #72: Make schema generation a builtin CLI command
Currently schema generation is an external script `gen_schema.py`. We could add this as a builtin kpops command. Additionally, this should take all custom components in the user's `<COMPONENTS_MODULE>` into account and generate a complete pipeline schema for both internal and custom components.

sujuka99:
closes #72

### DIFF
--- a/docs/docs/user/references/editor-integration.md
+++ b/docs/docs/user/references/editor-integration.md
@@ -13,7 +13,7 @@ KPOps provides JSON schemas that enable autocompletion and validation for some o
 [yaml-language-server](https://github.com/redhat-developer/yaml-language-server#clients){target=_blank} in your editor of choice. (requires LSP support)
 2. Configure the extension with the settings below.
 
-???+ note "`settings.json`"
+??? note "`settings.json`"
 
     ```json
       --8<--
--- a/docs/docs/user/references/editor-integration.md
+++ b/docs/docs/user/references/editor-integration.md
@@ -20,3 +20,6 @@ KPOps provides JSON schemas that enable autocompletion and validation for some o
       ./docs/resources/editor_integration/settings.json
       --8<--
     ```
+
+!!! tip "Advanced usage"
+    It is possible to generate schemas with the [`kpops schema`](../cli-commands/#kpops-schema) command. Useful when using custom components or when using a pre-release version of KPOps.
--- a/kpops/cli/main.py
+++ b/kpops/cli/main.py
@@ -188,17 +188,23 @@ def create_pipeline_config(
     return pipeline_config
 
 
-@app.command(help="Generate json schema.")
+@app.command(
+    help="Generates json schemas. " 
+    "The schemas can be used to enable support for kpops files in a text editor."
+)
 def gen_schema(
-    components_module: Optional[str] = COMPONENTS_MODULES,
-    path: Path = typer.Option(
-        default=Path("."), help="Target path for the generation of schemas."
+    path: Path = typer.Argument(
+        default=...,
+        help="Target path for the generation of schemas."
     ),
+    components_module: Optional[str] = COMPONENTS_MODULES,
     pipeline: bool = typer.Option(
-        True, help="Generate json schema of pipeline components."
+        default=True,
+        help="Generate json schema of pipeline components. Always includes the built-in kpops components. "
+        "To include custom components, provide [COMPONENTS_MODULES]"
     ),
     config: bool = typer.Option(
-        False, help="Generate json schema pipeline configuration."
+        default=False, help="Generate json schema pipeline configuration."
     ),
 ) -> None:
     if pipeline:
--- a/kpops/cli/main.py
+++ b/kpops/cli/main.py
@@ -17,7 +17,7 @@ from kpops.component_handlers.schema_handler.schema_handler import SchemaHandler
 from kpops.component_handlers.topic.handler import TopicHandler
 from kpops.component_handlers.topic.proxy_wrapper import ProxyWrapper
 from kpops.pipeline_generator.pipeline import Pipeline
-from kpops.utils.gen_schema import gen_schemas
+from kpops.utils.gen_schema import gen_pipeline_schema
 
 if TYPE_CHECKING:
     from kpops.components.base_components import PipelineComponent
@@ -85,7 +85,7 @@ COMPONENTS_MODULES: str | None = typer.Argument(
 )
 
 SCHEMA_PATH: Path = typer.Option(
-    default = Path("~"),
+    default = Path("."),
     help="Target path for the generation of schemas."
 )
 
@@ -194,8 +194,8 @@ def create_pipeline_config(
 
 
 @app.command(help="Generate actual schema")
-def gen_schema(path: Path = SCHEMA_PATH) -> None:
-    gen_schemas(path)
+def gen_schema(components_module: Optional[str] = COMPONENTS_MODULES, path: Path = SCHEMA_PATH) -> None:
+    gen_pipeline_schema(components_module, path)
 
 
 @app.command(
--- a/kpops/cli/main.py
+++ b/kpops/cli/main.py
@@ -85,8 +85,7 @@ COMPONENTS_MODULES: str | None = typer.Argument(
 )
 
 SCHEMA_PATH: Path = typer.Option(
-    default = Path("."),
-    help="Target path for the generation of schemas."
+    default=Path("."), help="Target path for the generation of schemas."
 )
 
 logger = logging.getLogger()
@@ -194,7 +193,9 @@ def create_pipeline_config(
 
 
 @app.command(help="Generate actual schema")
-def gen_schema(components_module: Optional[str] = COMPONENTS_MODULES, path: Path = SCHEMA_PATH) -> None:
+def gen_schema(
+    components_module: Optional[str] = COMPONENTS_MODULES, path: Path = SCHEMA_PATH
+) -> None:
     gen_pipeline_schema(components_module, path)
 
 
--- a/kpops/cli/main.py
+++ b/kpops/cli/main.py
@@ -189,19 +189,18 @@ def create_pipeline_config(
 
 
 @app.command(
-    help="Generates json schemas. " 
+    help="Generates json schemas. "
     "The schemas can be used to enable support for kpops files in a text editor."
 )
 def gen_schema(
     path: Path = typer.Argument(
-        default=...,
-        help="Target path for the generation of schemas."
+        default=..., help="Target path for the generation of schemas."
     ),
     components_module: Optional[str] = COMPONENTS_MODULES,
     pipeline: bool = typer.Option(
         default=True,
         help="Generate json schema of pipeline components. Always includes the built-in kpops components. "
-        "To include custom components, provide [COMPONENTS_MODULES]"
+        "To include custom components, provide [COMPONENTS_MODULES]",
     ),
     config: bool = typer.Option(
         default=False, help="Generate json schema pipeline configuration."
--- a/kpops/cli/main.py
+++ b/kpops/cli/main.py
@@ -17,6 +17,7 @@ from kpops.component_handlers.schema_handler.schema_handler import SchemaHandler
 from kpops.component_handlers.topic.handler import TopicHandler
 from kpops.component_handlers.topic.proxy_wrapper import ProxyWrapper
 from kpops.pipeline_generator.pipeline import Pipeline
+from kpops.utils.gen_schema import gen_schemas
 
 if TYPE_CHECKING:
     from kpops.components.base_components import PipelineComponent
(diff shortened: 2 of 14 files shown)


kpops/utils/gen_schema.py
```diff
@@ -0,0 +1,82 @@
...
+    if components_module:
+        components = chain(
+            components, _find_classes(components_module, PipelineComponent)
+        )
+
+    # Create a variable that will hold the union of all component types
+    PipelineComponent_ = KafkaConnector
+
+    for component in components:
+        PipelineComponent_ = PipelineComponent_ | component  # type: ignore[assignment]
```
disrupted, line 58: it's pretty cool that we can build the type union dynamically, I didn't know that 👍🏻 
However, I am not a big fan of the for loop and building the type union inside it. I'll make a proposal for changing this to use tuples if you don't mind

kpops/cli/main.py
```diff
@@ -187,6 +188,30 @@ def create_pipeline_config(
     return pipeline_config
 
 
+@app.command(
+    help="Generates json schemas. "
+    "The schemas can be used to enable support for kpops files in a text editor."
+)
+def gen_schema(
```
disrupted, line 195: I would prefer simply `schema` 🙂

kpops/cli/main.py
```diff
@@ -187,6 +188,30 @@ def create_pipeline_config(
...
+    path: Path = typer.Argument(
```
disrupted, line 196: what's your opinion on printing the schema to stdout instead of saving it to a file? of course we could then only generate one schema at a time (for pipeline _or_ config)

disrupted: the ideal "end game" use case I have in mind: 
1. open any deployment project in my code editor
2. it automatically executes `kpops schema ... components` generating the pipeline schema including all the project-specific custom components
3. the schema is loaded into YAML language server
4. start editing with validation + autocompletion for every component

sujuka99: Sounds cool. I can enforce that only 1 flag can be set at a time.

kpops/cli/main.py
sujuka99: I like it, will do

kpops/utils/gen_schema.py
sujuka99: I don't

kpops/utils/gen_schema.py
```diff
@@ -0,0 +1,137 @@
...
+        raise SkipField(f"{field.name} field is being hidden")
+    else:
+        return original_field_schema(field, **kwargs)
+
+
+schema.field_schema = field_schema
+
+log = logging.getLogger("")
+
+ModelField
```
disrupted, line 35: what's the purpose of this?

kpops/utils/gen_schema.py
```diff
@@ -0,0 +1,137 @@
...
+from rich.console import Console
+
+from kpops.cli.pipeline_config import PipelineConfig
+from kpops.cli.registry import _find_classes
+from kpops.components.base_components.pipeline_component import PipelineComponent
+
+
+class SchemaScope(str, Enum):
+    pipeline = "pipeline"
+    config = "config"
```
disrupted, line 17: please uppercase enum members

kpops/cli/main.py
```diff
@@ -187,6 +187,31 @@ def create_pipeline_config(
...
+        help="""
+        Scope of the generated schema
+
+        pipeline: Schema of PipelineComponents. Always includes the built-in kpops components. To include custom components, provide [COMPONENTS_MODULES].
+
+        config: Schema of PipelineConfig.""",
+    ),
+    components_module: Optional[str] = COMPONENTS_MODULES,
+) -> None:
+    if scope == "pipeline":
```
disrupted, line 209: compare with enum member, e.g. `SchemaScope.PIPELINE`

disrupted: also consider using match-case instead (pattern matching)

kpops/utils/gen_schema.py
```diff
@@ -0,0 +1,137 @@
...
+
+    schema = schema_json_of(
+        Sequence[AnnotatedPipelineComponents],
+        title="kpops pipeline schema",
+        by_alias=True,
+        indent=4,
+    ).replace("schema_type", "type")
+    Console(
+        width=1000  # HACK: overwrite console width to avoid truncating output
+    ).print(schema)
```
disrupted, line 123: would regular print work as well? I don't think we need rich console here

kpops/utils/gen_schema.py
```diff
@@ -0,0 +1,137 @@
...
+    :param components_module: Python module. Only the classes that inherit from PipelineComponent will be considered.
+    :type components_module: str or None
+    :rtype: None
+    """
+
+    components = tuple(_find_classes("kpops.components", PipelineComponent))
+    if components_module:
+        # record components types used so far
+        used_fields: list[str] = [
+            component.__fields__.get("schema_type").default  # type: ignore [union-attr]
```
disrupted, line 98: please fix some of these type checking issues
if you're accessing an optional member here using `get(...)` the return value is `str | None`, therefore accessing the `.default` attribute is unsafe

kpops/utils/gen_schema.py
```diff
@@ -0,0 +1,137 @@
...
+log = logging.getLogger("")
+
+ModelField
+
+
+def is_valid_component(
+    used_fields: list[str], component: type[PipelineComponent]
+) -> bool:
+    """
+    Check whether a PipelineComponent subclass has a valid definition for the schema generation.
```
disrupted, line 42: good idea 👍🏻

disrupted (changes requested):

kpops/utils/gen_schema.py
```diff
@@ -0,0 +1,137 @@
...
+    :param used_fields: types defined so far
+    :type used_fields: list[str]
+    :param component: component type to be validated
+    :type component: type[PipelineComponent]
+    :return: Whether component is valid for schema generation
+    :rtype: bool
+    """
+    component_name: str = component.__name__
+    # PipelineComponent always has a field 'type'
+    type: str = component.__fields__.get("type").field_info.default  # type: ignore[union-attr]
```
disrupted, line 53: `type` is a builtin function in Python, let's not override it. as a workaround you can change the variable name to something like `type_` or `typ`

disrupted: also consider using the classmethod which we've already added as a helper
`component.get_component_type()`

kpops/utils/gen_schema.py
```diff
@@ -0,0 +1,137 @@
...
+    if not component.__fields__.get("schema_type"):
+        log.warning(f"SKIPPED {component_name}, schema_type is not defined")
+        return False
+    schema_type: ModelField = component.__fields__.get("schema_type")  # type: ignore[assignment]
```
disrupted, line 57: you're already checking if the `schema_type` attribute exists in the condition above. here you can directly access it then: `component.__fields__["schema_type"]`

disrupted: even better would be to rely on type narrowing

example
```python
schema_type = component.__fields__.get("schema_type") # ModelField | None
if not schema_type:
    return
schema_type # ModelField
```

kpops/utils/gen_schema.py
```diff
@@ -0,0 +1,137 @@
...
+
+    :param components_module: Python module. Only the classes that inherit from PipelineComponent will be considered.
+    :type components_module: str or None
+    :rtype: None
+    """
+
+    components = tuple(_find_classes("kpops.components", PipelineComponent))
+    if components_module:
+        # record components types used so far
+        used_fields: list[str] = [
```
disrupted, line 97: what's your opinion on using `set` instead of `list`? it's more efficient to do lookups against it

kpops/utils/gen_schema.py
sujuka99: It would work, good point.

kpops/utils/gen_schema.py
sujuka99: `set` seems like a good idea :)

sujuka99: done

kpops/utils/gen_schema.py
sujuka99: done

kpops/utils/gen_schema.py
sujuka99: done

kpops/utils/gen_schema.py
sujuka99: done

kpops/utils/gen_schema.py
sujuka99: done

kpops/cli/main.py
sujuka99: done

kpops/utils/gen_schema.py
sujuka99: done

kpops/utils/gen_schema.py
sujuka99: done

kpops/utils/gen_schema.py
```diff
@@ -0,0 +1,139 @@
...
+    :param used_fields: types defined so far
+    :type used_fields: list[str]
+    :param component: component type to be validated
+    :type component: type[PipelineComponent]
+    :return: Whether component is valid for schema generation
+    :rtype: bool
+    """
+    component_name: str = component.__name__
+    component_type: str = component.get_component_type()
+    schema_type: Optional[ModelField] = component.__fields__.get("schema_type")
```
disrupted, line 59: is there a reason why you're using explicit type annotations here? it looks to me that those are unnecessary as they can be easily inferred by the type checker

disrupted (commented):
looks great, just commenting something I noticed

docs/docs/user/references/editor-integration.md
```diff
@@ -13,10 +13,13 @@ KPOps provides JSON schemas that enable autocompletion and validation for some o
...
+??? note "`settings.json`"
 
     ```json
       --8<--
       ./docs/resources/editor_integration/settings.json
       --8<--
     ```
+
+!!! tip "Advanced usage"
+    It is possible to generate schemas with the [`kpops schema`](../cli-commands/#kpops-schema) command. Useful when using custom components or when using a pre-release version of KPOps.
```
disrupted, line 25: in the future we should update this section with a small guide on how the custom schema generation can be integrated into a project, because this is a really cool feature to have 🙂

kpops/utils/gen_schema.py
```diff
@@ -0,0 +1,139 @@
...
+log = logging.getLogger("")
+
+
+def is_valid_component(
+    used_fields: set[str], component: type[PipelineComponent]
+) -> bool:
+    """
+    Check whether a PipelineComponent subclass has a valid definition for the schema generation.
+
+    :param used_fields: types defined so far
```
disrupted, line 50: the parameter name `used_fields` might be a little bit misleading

kpops/utils/gen_schema.py
sujuka99: It helps me a bit to read what I've written and to catch/avoid mistakes.

Should I remove them?

docs/docs/user/references/editor-integration.md
sujuka99: Agreed, created an [issue](https://github.com/bakdata/kpops/issues/173) for it.

kpops/utils/gen_schema.py
disrupted: are you using a type checker in your editor? something like Pylance/Pyright or MyPy? these should both be able to infer the types automatically

disrupted: I would prefer to not have explicit type annotations here unless they really make sense, as it makes it a bit verbose

kpops/utils/gen_schema.py
sujuka99: agreed, changed to `defined_component_types`

kpops/utils/gen_schema.py
sujuka99: I use Pylance.

Understandable, I've removed them.

tests/cli/test_schema_generation.py
```diff
@@ -34,12 +34,12 @@ class SubPipelineComponent(EmptyPipelineComponent):
     )
 
 
-# schema_type is ineritted from SubPipelineComponent
+# schema_type is inerited from SubPipelineComponent
```
disrupted:
```suggestion
# schema_type is inherited from SubPipelineComponent
```

kpops/utils/gen_schema.py
disrupted: make sure to configure [Inlay Type Hints](https://devblogs.microsoft.com/python/python-in-visual-studio-code-july-2022-release/#inlay-type-hints) for Pylance

tests/cli/test_schema_generation.py
sujuka99: saw this, took 2 tries, but it's fixed now haha

disrupted (approved):

Issue #110, opened 2023-02-14
Add filtering to `generate` command for printing the `pipeline.yaml`


raminqaf:
When we call the' generate' command, we cannot filter the steps we want to print. This is because we are printing the entire pipeline at once and not iterating through each pipeline component.
Moreover, not having this filtering mechanism introduced [bug](https://github.com/bakdata/kpops/issues/109).
We should change this behavior and add a filtering mechanism to the print.

PR #169, opened 2023-03-23
Error when running generate with --steps

### LINKED ISSUE #109: Error when running generate with --steps
```
╭───────────────────── Traceback (most recent call last) ──────────────────────╮
│ /usr/local/lib/python3.10/site-packages/kpops/cli/main.py:224 in generate    │
│                                                                              │
│   221 │   │   for component in steps_to_apply:                               │
│   222 │   │   │   component.template(api_version, ca_file, cert_file)        │
│   223 │   elif cert_file or ca_file or api_version or steps:                 │
│ ❱ 224 │   │   raise TypeError(                                               │
│   225 │   │   │   "The following flags can only be used in conjuction with ` │
│   226 │   │   │   │   '--cert-file'\n \                                      │
│   227 │   │   │   │   '--ca-file'\n \                                        │
╰──────────────────────────────────────────────────────────────────────────────╯
TypeError: The following flags can only be used in conjuction with `--template`:
                 '--cert-file'
                 '--ca-file'
                 '--api-version'
                 '--steps'
```

sujuka99:
fixes #109

### DIFF
--- a/kpops/cli/main.py
+++ b/kpops/cli/main.py
@@ -220,8 +220,8 @@ def generate(
         for component in steps_to_apply:
             component.template(api_version, ca_file, cert_file)
     elif cert_file or ca_file or api_version or steps:
-        raise TypeError(
-            "The following flags can only be used in conjuction with `--template`: \n \
+        log.warning(
+            "The following flags are considered only when `--template` is set: \n \
                 '--cert-file'\n \
                 '--ca-file'\n \
                 '--api-version'\n \


sujuka99:
Can the warning also cause issues? If yes, I will just remove the else entirely.

kpops/cli/main.py
```diff
@@ -220,8 +220,8 @@ def generate(
         for component in steps_to_apply:
             component.template(api_version, ca_file, cert_file)
     elif cert_file or ca_file or api_version or steps:
-        raise TypeError(
-            "The following flags can only be used in conjuction with `--template`: \n \
+        log.warning(
```
sujuka99, line 223: Can the warning also cause issues? If yes, I will just remove the else entirely.

philipp94831 (commented):
Wouldn't it be better to also just filter the steps when generating the output?

sujuka99:
I think it's better to release a patch as soon as possible as KPOps is supposed to be stable now.

philipp94831:
I don't think anything is really broken. Just the behavior was different some time ago. But if people are just starting to use it now, all is fine

kpops/cli/main.py
raminqaf: No the warning does not cause an issue :)

raminqaf:
> I don't think a quick fix is needed. Just #110 :-)

@sujuka99 I agree with @philipp94831. The fix for this bug is implementing #110. As far as I know. None of the projects are facing this bug now. You can go ahead and merge this because I still think that throwing an exception is unnecessary.

raminqaf (approved):

PR #174, opened 2023-03-28
Add CLI Usage doc generation to CI

### LINKED ISSUE #162: Add CLI Usage doc generation to CI
Generate [cli-commands.md](https://github.com/bakdata/kpops/blob/main/docs/docs/user/references/cli-commands.md) in the CI

Currently the generation is a bit weird, the command has to be
`poetry run typer ./kpops/cli/main.py utils docs --name kpops --output ./docs/docs/user/references/cli-commands.md`

if `--name` is not specified or is different, `kpops` is missing from the example commands, but with it the page is called `kpops`

Proposed workaround:
Execute the following:

```
run typer ./kpops/cli/main.py utils docs --name kpops --output ./docs/docs/user/references/cli-commands.md
sed -i '1s/.*/# CLI Usage/' ./docs/docs/user/references/cli-commands.md
```

sujuka99:
closes #162


justfile
```diff
@@ -13,4 +13,6 @@ docs-config := base-directory / "docs/mkdocs.yml"
 
 # Serve current docs located in ./docs/docs
 serve-docs port="8000":
+    typer {{base-directory}}/kpops/cli/main.py utils docs --name kpops --output {{base-directory}}/docs/docs/user/references/cli-commands.md
```
raminqaf, line 16: Shouldn't be like this?
```suggestion
    poetry run typer {{base-directory}}/kpops/cli/main.py utils docs --name kpops --output {{base-directory}}/docs/docs/user/references/cli-commands.md
```

.github/workflows/gen-docs.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+      - name: Install optional dependencies
+        shell: bash
+        run: poetry install --with docs
+
+      - name: Generate CLI Usage page
+        shell: bash
+        run: |
+          # Generate page with typer-cli
+          poetry run typer ./kpops/cli/main.py utils docs --name kpops --output ./docs/docs/user/references/cli-commands.md
```
raminqaf, line 27: Instead of repeating yourself here. Use the just file that you created.

justfile
raminqaf, line 16: Besides, this should not be under the `serve-docs` command. Please put it in separate command

.github/workflows/gen-docs.yaml
raminqaf, line 1: Instead of a workflow, maybe change this to action and just add it to the next step of the `test` job

.github/workflows/ci.yaml
```diff
@@ -54,3 +54,8 @@ jobs:
     uses: ./.github/workflows/publish-test-pypi.yaml
     secrets:
       test-pypi-token: ${{ secrets.TEST_PYPI_TOKEN }}
+    
+  gen-docs:
+    name: Generate docs
+    needs: [test]
+    uses: ./.github/workflows/gen-docs.yaml
```
raminqaf, line 61: You generate the docs after you publish to test pypi. It is not critical but I would say it is better to do it the other way around

.github/workflows/gen-docs.yaml
```diff
@@ -0,0 +1,29 @@
...
+          # Fix title
+          sed -i '1s/.*/# CLI Usage/' ./docs/docs/user/references/cli-commands.md
```
raminqaf, line 29: Who commits the changes of the file?

raminqaf (changes requested):

justfile
sujuka99: It should

.github/workflows/gen-docs.yaml
sujuka99: done

.github/workflows/gen-docs.yaml
sujuka99: I created another job. Is there a reason to put it in test?

.github/workflows/gen-docs.yaml
sujuka99: Very good point :DD

I've fixed this I believe.

justfile
sujuka99: done

.github/actions/gen-docs/action.yaml
```diff
@@ -0,0 +1,56 @@
...
+    description: "GitHub Token (must be a PAT for repository dispatch)"
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install python and set up Poetry
+      uses: bakdata/ci-templates/actions/python-setup-poetry@v1.5.3
+      with:
+        poetry-version: "1.3.2"
```
raminqaf, line 22: Maybe it is a good idea to add this as an input parameter.  Incase we change that poetry version in the CI workflow we don't forget it here

docs/docs/dev/ci.md
```diff
@@ -0,0 +1 @@
+# CI
```
raminqaf, line 1: I guess this is still WIP?

.github/workflows/ci.yaml
```diff
@@ -48,6 +48,23 @@ jobs:
...
 
+  gen-docs:
+    runs-on: ubuntu-22.04
+    needs: test
+    if: "! github.event.pull_request.draft"
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
```
raminqaf, line 59: Can you move this to action?

.github/actions/gen-docs/action.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - name: Configure git
+      shell: bash
+      env:
+        USERNAME: ${{ inputs.username }}
+        EMAIL: ${{ inputs.email }}
+        TOKEN: ${{ inputs.token }}
+      run: |
+        git config --local user.name ${USERNAME}
+        git config --local user.email ${EMAIL}
+        git config --local user.password ${TOKEN}
```
raminqaf, line 38: You can replace all of this with this action:
https://github.com/bakdata/ci-templates/blob/v1.17.0/actions/commit-and-push/action.yaml

raminqaf (changes requested):

docs/docs/dev/ci.md
sujuka99: It was, I added some to it now.
I wouldn't document more than the action I added now in this PR. WDYT?

gen_docs.sh
raminqaf, line 1: Why not use the justfile? 😄

pyproject.toml
```diff
@@ -60,6 +60,7 @@ mkdocs-material-extensions = "^1.1.1"
 mkdocs-glightbox = "^0.3.1"
 mike = "^1.1.2"
 typer-cli = "^0.0.13"
+mkdocs-codeinclude-plugin = "^0.2.1"
```
raminqaf, line 63: Not used?

docs/docs/dev/ci.md
raminqaf: The documentation is a bit overkill. A brief explanation of how the workflows are working is completely sufficient.
We don't want to reuse these actions in other repositories. The documentation should only inform the developers how we setup the CI/CD

raminqaf (changes requested):

pyproject.toml
sujuka99: I ended up not using it at all. Should I remove it?

It would be a nice tool further in time. How can we draw attention to it?

gen_docs.sh
sujuka99: I didn't want to chain things up like that.
Also, this way we don't need to install `Just` to run the generation.

What benefits do you see in keeping the commands in the justfile?

docs/docs/dev/ci.md
sujuka99: It is also outdated now, I have deleted the action entirely. Should I document anything at all? Seems a bit out of scope for this PR.

Perhaps just the added hook and step in the CI?

pyproject.toml
raminqaf: If it is not used, It should be removed. You mentioned that you are using this [plugin](https://facelessuser.github.io/pymdown-extensions/extensions/snippets/). What are the differences between them? Is one better than the other one?
If it is so and we need to change it, it makes sense to create an issue. This is how you draw attention to it. 😄

docs/docs/dev/ci.md
raminqaf: Yeah, let's just skip it for this PR.

gen_docs.sh
raminqaf: I see your point of not installing `just`. But in the bash file, your are using a command like this:
```
scriptDir=$(dirname -- "$(readlink -f -- "$BASH_SOURCE")")
```
Which is hard to read, and this script will break whenever we change the `gen_docs.sh` file.
In the `justfile`, you are using a command like that:
```just
base-directory := justfile_directory()
```
which is way easier to read and understand.
Moreover, you can define other variables that can be used for the documentation-related commands in your `justfile`. For example:

```just
doc-path := base-directory / "docs"
```

kpops/cli/main.py
```diff
@@ -187,6 +187,11 @@ def create_pipeline_config(
     return pipeline_config
 
 
+@app.command(help="help")
+def test_(test_=typer.Option(default=True, help="help")):
+    pass
```
raminqaf, line 192: Remove this if you finish testing the CI

.pre-commit-config.yaml
```diff
@@ -29,3 +29,10 @@ repos:
         stages: [manual]
         args: ["--py310-plus"]
         exclude: ^tests/.*snapshots/
+  - repo: local
+    hooks:
+      - id: gen-docs
+        name: gen-docs
+        entry: ./gen_docs.sh
```
raminqaf, line 36: Okay, I get the point with the `justfile` now... I don't know if it is possible to run `just` here...

raminqaf: I think it is not possible
https://pre-commit.com/#supported-languages

gen_docs.sh
raminqaf: I think it is impossible to run the `just` command in the pre-commit hook (please check that). If so I would remove the command from `just`, since it is not giving us any benefit from `just`.

.pre-commit-config.yaml
sujuka99: It would have to be a bash script containing `just generate-docs`

pyproject.toml
sujuka99: I cannot say whether one is better, but I think the one we are using has more features currently. The thing that [snippets](https://facelessuser.github.io/pymdown-extensions/extensions/snippets/) is missing is using [expressions](https://github.com/rnorth/mkdocs-codeinclude-plugin#usage) to find and include code from other files.

[Codeinclude](https://github.com/rnorth/mkdocs-codeinclude-plugin), however, cannot import files locally or from the cloud via URL.

That's why I think we can use the 2 plugins concurrently.

NOTE: I couldn't make [codeinclude](https://github.com/rnorth/mkdocs-codeinclude-plugin) work for some reason, so even if the problem is in me, it's a clear indicator that its documentation isn't great currently.

gen_docs.sh
sujuka99: I cannot find a way to use the just command without specifying a bash script that runs it.

We could keep the command in just to allow for easier generation when working on the CLI and the docs?

raminqaf: Okay let's keep it in just

docs/docs/dev/ci.md
sujuka99: okay, removed the doc

pyproject.toml
raminqaf: Okay interesting findings. I would still remove codeinclude and create an issue about why and where we can use this plugin. And then tackle it in another PR.

sujuka99: https://github.com/bakdata/kpops/issues/179

gen_docs.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+scriptDir=$(dirname -- "$(readlink -f -- "$BASH_SOURCE")")
+
+echo "generate CLI Usage"
+poetry run typer $scriptDir/kpops/cli/main.py utils docs --name kpops --output $scriptDir/docs/docs/user/references/cli-commands.md
+
+echo "Fix name"
```
raminqaf, line 8:
```suggestion
echo "Changing title name to CLI Usage"
```

gen_docs.sh
```diff
@@ -0,0 +1,16 @@
...
+
+echo "Changing title to CLI Usage"
+sed -i '1s/.*/# CLI Usage/' $scriptDir/docs/docs/user/references/cli-commands.md
+
+if [ -n "$(git diff --exit-code $scriptDir/docs)" ]
+then
+    echo "::error::Generatable documentation is not up-to-date. Updating..." && exit 1
+else
+    echo "::notice::Documentation is up-to-date."
+fi
```
raminqaf, line 16: Could you please check if this works?
```suggestion
#!/usr/bin/env bash

scriptDir=$(dirname -- "$(readlink -f -- ["${BASH_SOURCE[0]}"])")
output="$scriptDir"/docs/docs/user/references/cli-commands.md
echo "generate CLI Usage"
poetry run typer "$scriptDir"/kpops/cli/main.py utils docs --name kpops --output "$output"

echo "Changing title to CLI Usage"
sed -i '1s/.*/# CLI Usage/' "$output"

if [ -n "$(git diff --exit-code "$scriptDir"/docs)" ]
then
    echo "::error::Generatable documentation is not up-to-date. Updating..." && exit 1
else
    echo "::notice::Documentation is up-to-date."
fi
```

sujuka99: It works, but I had to remove the square brackets surrounding `BASH_SOURCE` as [they return bool](https://linuxhint.com/bash-test-command/).

raminqaf (approved):
🚀

PR #191, opened 2023-04-06
Add titles and descriptions to Pydantic model fields

### LINKED ISSUE #74: Add titles and descriptions to Pydantic model fields

sujuka99:
Closes #74

### DIFF
--- a/docs/docs/schema/pipeline.json
+++ b/docs/docs/schema/pipeline.json
@@ -10,6 +10,7 @@
                 "kafka-sink-connector": "#/definitions/KafkaSinkConnector",
                 "kafka-source-connector": "#/definitions/KafkaSourceConnector",
                 "kubernetes-app": "#/definitions/KubernetesApp",
+                "pipeline-component": "#/definitions/PipelineComponent",
                 "producer": "#/definitions/ProducerApp",
                 "streams-app": "#/definitions/StreamsApp"
             }
@@ -29,6 +30,9 @@
             },
             {
                 "$ref": "#/definitions/KubernetesApp"
+            },
+            {
+                "$ref": "#/definitions/PipelineComponent"
             },
             {
                 "$ref": "#/definitions/ProducerApp"
@@ -297,6 +301,14 @@
                     "default": "kafka-app",
                     "type": "string"
                 },
+                "type": {
+                    "title": "Schema Type",
+                    "default": "kafka-app",
+                    "enum": [
+                        "kafka-app"
+                    ],
+                    "type": "string"
+                },
                 "name": {
                     "title": "Name",
                     "description": "Component name",
@@ -333,15 +345,6 @@
                     "title": "Prefix",
                     "description": "Pipeline prefix that will prefix every component name.\n        If you wish to not have any prefix you can specify an empty string.,",
                     "default": "${pipeline_name}-",
-                    "type": "string"
-                },
-                "type": {
-                    "title": "Component type",
-                    "description": "Base component for Kafka-based components.\n\n    Producer or streaming apps should inherit from this class.",
-                    "default": "kafka-app",
-                    "enum": [
-                        "kafka-app"
-                    ],
                     "type": "string"
                 },
                 "repoConfig": {
@@ -401,6 +404,14 @@
                     "default": "kafka-connector",
                     "type": "string"
                 },
+                "type": {
+                    "title": "Schema Type",
+                    "default": "kafka-connector",
+                    "enum": [
+                        "kafka-connector"
+                    ],
+                    "type": "string"
+                },
                 "name": {
                     "title": "Name",
                     "description": "Component name",
@@ -437,15 +448,6 @@
                     "title": "Prefix",
                     "description": "Pipeline prefix that will prefix every component name.\n        If you wish to not have any prefix you can specify an empty string.,",
                     "default": "${pipeline_name}-",
-                    "type": "string"
-                },
-                "type": {
-                    "title": "Component type",
-                    "description": "Base class for all Kafka connectors\n\n    Should only be used to set defaults",
-                    "default": "kafka-connector",
-                    "enum": [
-                        "kafka-connector"
-                    ],
                     "type": "string"
                 },
                 "repoConfig": {
@@ -501,6 +503,14 @@
                     "default": "kafka-sink-connector",
                     "type": "string"
                 },
+                "type": {
+                    "title": "Schema Type",
+                    "default": "kafka-sink-connector",
+                    "enum": [
+                        "kafka-sink-connector"
+                    ],
+                    "type": "string"
+                },
                 "name": {
                     "title": "Name",
                     "description": "Component name",
@@ -537,15 +547,6 @@
                     "title": "Prefix",
                     "description": "Pipeline prefix that will prefix every component name.\n        If you wish to not have any prefix you can specify an empty string.,",
                     "default": "${pipeline_name}-",
-                    "type": "string"
-                },
-                "type": {
-                    "title": "Component type",
-                    "description": "Kafka sink connector model",
-                    "default": "kafka-sink-connector",
-                    "enum": [
-                        "kafka-sink-connector"
-                    ],
                     "type": "string"
                 },
                 "repoConfig": {
@@ -601,6 +602,14 @@
                     "default": "kafka-source-connector",
                     "type": "string"
                 },
+                "type": {
+                    "title": "Schema Type",
+                    "default": "kafka-source-connector",
+                    "enum": [
+                        "kafka-source-connector"
+                    ],
+                    "type": "string"
+                },
                 "name": {
                     "title": "Name",
                     "description": "Component name",
@@ -637,15 +646,6 @@
                     "title": "Prefix",
                     "description": "Pipeline prefix that will prefix every component name.\n        If you wish to not have any prefix you can specify an empty string.,",
                     "default": "${pipeline_name}-",
-                    "type": "string"
-                },
-                "type": {
-                    "title": "Component type",
-                    "description": "Kafka source connector model",
-                    "default": "kafka-source-connector",
-                    "enum": [
-                        "kafka-source-connector"
-                    ],
                     "type": "string"
                 },
                 "repoConfig": {
@@ -712,6 +712,14 @@
                     "default": "kubernetes-app",
                     "type": "string"
                 },
+                "type": {
+                    "title": "Schema Type",
+                    "default": "kubernetes-app",
+                    "enum": [
+                        "kubernetes-app"
+                    ],
+                    "type": "string"
+                },
                 "name": {
                     "title": "Name",
                     "description": "Component name",
@@ -750,15 +758,6 @@
                     "default": "${pipeline_name}-",
                     "type": "string"
                 },
-                "type": {
-                    "title": "Component type",
-                    "description": "Base class for all Kubernetes apps.\n\n    All built-in components are Kubernetes apps, except for the Kafka connectors.",
-                    "default": "kubernetes-app",
-                    "enum": [
-                        "kubernetes-app"
-                    ],
-                    "type": "string"
-                },
                 "repoConfig": {
                     "title": "Repoconfig",
                     "description": "Configuration of the Helm chart repo to be used for deploying the component",
@@ -783,6 +782,52 @@
                 "name",
                 "app",
                 "namespace"
+            ]
+        },
+        "PipelineComponent": {
+            "title": "PipelineComponent",
+            "type": "object",
+            "properties": {
+                "type": {
+                    "title": "Type",
+                    "default": "pipeline-component",
+                    "type": "string"
+                },
+                "type": {
+                    "title": "Schema Type",
+                    "default": "pipeline-component",
+                    "enum": [
+                        "pipeline-component"
+                    ],
+                    "type": "string"
+                },
+                "name": {
+                    "title": "Name",
+                    "type": "string"
+                },
+                "from": {
+                    "title": "From",
+                    "allOf": [
+                        {
+                            "$ref": "#/definitions/FromSection"
+                        }
+                    ]
+                },
+                "app": {
+                    "title": "App"
+                },
+                "to": {
+                    "$ref": "#/definitions/ToSection"
+                },
+                "prefix": {
+                    "title": "Prefix",
+                    "description": "Pipeline prefix that will prefix every component name. If you wish to not have any prefix you can specify an empty string.",
+                    "default": "${pipeline_name}-",
+                    "type": "string"
+                }
+            },
+            "required": [
+                "name"
             ]
         },
         "ProducerStreamsConfig": {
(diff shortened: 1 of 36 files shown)


kpops/component_handlers/helm_wrapper/helm.py
```diff
@@ -79,9 +79,7 @@ def upgrade_install(
         values: dict,
         flags: HelmUpgradeInstallFlags = HelmUpgradeInstallFlags(),
     ) -> str:
-        """
-        Prepares and executes the helm upgrade install command
-        """
+        """Prepares and executes the helm upgrade install command"""
```
disrupted, line 82:
```suggestion
        """Prepares and executes the `helm upgrade --install` command"""
```

kpops/component_handlers/kafka_connect/model.py
```diff
@@ -65,5 +66,6 @@ class KafkaConnectResetterValues(BaseModel):
     class Config(CamelCaseConfig):
         pass
 
+    @override
```
disrupted, line 69: great 👍🏻

kpops/components/base_components/base_defaults_component.py
```diff
@@ -17,12 +17,41 @@
...
-    config: PipelineConfig = Field(default=..., exclude=True, hidden_from_schema=True)
+    enrich: bool = Field(
+        default=False,
+        description="Whether to enrich component with defaults",
+        exclude=True,
+        hidden_from_schema=True,
+    )
+    config: PipelineConfig = Field(
+        default=...,
+        description="Pipeline configuration to be accessed by this component",
```
disrupted, line 46: it doesn't hurt to have a description here, but it wouldn't be required since the `config` and `handlers` attributes aren't part of the generated pipeline schema

disrupted (commented):
wow, you did a great job!

kpops/components/base_components/base_defaults_component.py
```diff
@@ -35,19 +64,25 @@ def __init__(self, **kwargs) -> None:
...
-        Merges tmp_defaults with all tmp_defaults for parent classes
+        """Merge parent components' defaults with own
 
-        :param kwargs: the init kwargs for pydantic
-        :return: enriched kwargs with tmp_defaults
-        """
+        Merges tmp_defaults with all tmp_defaults for parent classes.
 
+        :param kwargs: The init kwargs for pydantic
+        :returns: Enriched kwargs with tmp_defaults
```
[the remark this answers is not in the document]
sujuka99: That makes sense, I was confused, but thought that maybe it was intentional, will fix it

kpops/components/base_components/models/to_section.py
```diff
@@ -15,15 +16,40 @@ class OutputTopicTypes(str, Enum):
...
+    value_schema: str | None = Field(
+        default=None, alias="valueSchema", description="Value schema class name"
+    )
+    partitions_count: int | None = Field(
+        default=None, description="Number of partitions into which the topic is divided"
+    )
+    replication_factor: int | None = Field(
+        default=None, description="Replication topic of the topic"
+    )
+    configs: dict[str, str | int] = Field(default={}, description="Topic configs")
```
disrupted, line 48: we should validate that this change is working as expected. I had an issue recently in a similar Pydantic models, where the value type was defined as `str | bool` causing all bools to be cast to str :( I had to change the value type to `Any` as a workaround

kpops/components/base_components/base_defaults_component.py
sujuka99: Yeah, all class attributes in `BaseDefaultsComponent` are in the same position. I added the descriptions because it seemed like a way to document them, but It could be seen as kinda redundant as I have written the same descriptions in the class docstring right above.

A third option is adding a docstring below the assignments as outlined in [PEP 257](https://peps.python.org/pep-0257/#what-is-a-docstring), but I don't think it is needed.

Do you think that I should remove the descriptions from the assignments?

kpops/components/base_components/base_defaults_component.py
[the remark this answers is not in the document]
sujuka99: Would it be the same if I replace the param `kwargs: dict[str, Any]` with `**kwargs`?

kpops/components/base_components/models/to_section.py
sujuka99: :/// How can I go about that?

kpops/components/base_components/kafka_app.py
```diff
@@ -36,21 +36,44 @@ class KafkaAppConfig(KubernetesAppConfig):
...
+    :type version: str, optional
     """
 
-    type: str = "kafka-app"
+    type: str = Field(default="kafka-app", description="Component type")
     schema_type: Literal["kafka-app"] = Field(  # type: ignore[assignment]
-        default="kafka-app", exclude=True
+        default="kafka-app",
+        title="Component type",
+        description=__doc__.partition(":param")[0].strip(),
```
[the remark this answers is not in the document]
sujuka99: I could write a function that takes a class attribute name and returns the description from the docstring to further remove the repetition, but this could be considered going a little overboard. What do you think?

kpops/components/base_components/models/to_section.py
disrupted: ideally we already have a snapshot test for a pipeline that uses topic configs. could you check if that's the case?
otherwise I can test it in a deployment project for you (and we can add a separate test for it)

kpops/components/base_components/kafka_app.py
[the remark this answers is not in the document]
disrupted: interesting, I had no idea this is possible
yeah, this sounds like a good idea if you're using this in multiple places

kpops/components/base_components/base_defaults_component.py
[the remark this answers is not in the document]
disrupted: oh, absolutely! this should be `**kwargs` indeed

sujuka99:
Noticed that since I added a more detailed docstring to `ToSection` the JSON schema now includes the parameters definitions as well, which is undesirable for editor integration.

To combat this, I found [this solution](https://stackoverflow.com/a/74520984/11610149) and I wrote functions which would extract the desired descriptions. (Wrote one for class attributes, too)

I have yet to figure out how to access the `__doc__` attribute of a class from a nested class.

kpops/components/base_components/kafka_app.py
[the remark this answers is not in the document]
sujuka99: Welll, I didn't know if it was either :D
Ended up doing it a bit differently

disrupted:
> P.S. Why is the last commit not being checked by the CI??

no idea 🤷🏻‍♂️

disrupted:
> P.S. No CI on this PR I guess

lol

docs/docs/schema/pipeline.json
```diff
@@ -424,15 +496,18 @@
         },
         "KafkaSinkConnector": {
             "title": "KafkaSinkConnector",
+            "description": "Kafka sink connector model",
             "type": "object",
             "properties": {
                 "type": {
                     "title": "Type",
+                    "description": "Component type,",
```
disrupted, line 504: is it possible to remove the trailing comma in the description?

kpops/utils/docstring.py
```diff
@@ -0,0 +1,85 @@
+import logging
+
+from pydantic import BaseModel
+
+log = logging.getLogger("docstring_utils")
+
+
+def describe_attr(name: str, docstr: str) -> str:
```
disrupted, line 8: Pyright says `__doc__` can be of type `str | None`. Please take this into account for the function signature

docs/docs/schema/pipeline.json
sujuka99: Must be fixed now, good catch

kpops/utils/docstring.py
```diff
@@ -0,0 +1,83 @@
...
+    :rtype: str
+    """
+    docstr = docstr.partition(f":param {name}:")[2]
+
+    docstr = _trim_description_end(docstr)
+
+    return docstr
+
+
+def describe_class(class_: type[BaseModel] | str) -> str:
```
disrupted, line 27: I can only find one usage where the `class_` argument is actually a class (`type[BaseModel]`). I would suggest passing there `model.__doc__` as well so that this function has less responsibilities. Then you could change the function signature also to `docstr: str | None`

sujuka99: I did, you are right

kpops/components/base_components/models/to_section.py
sujuka99: [tests/pipeline/resources/defaults.yaml](https://github.com/bakdata/kpops/blob/main/tests/pipeline/resources/defaults.yaml#L25) uses some

kpops/utils/docstring.py
```diff
@@ -1,51 +1,48 @@
...
     :param name: Attribute name
     :type name: str
     :param docstr: Docstring from which to read. Note that the class' docstring is stored in attr `__doc__`
-    :type docstr: str
+    :type docstr: str, None
     :returns: Description of the class attribute read from the class docstring
     :rtype: str
     """
-    docstr = docstr.partition(f":param {name}:")[2]
+    if not isinstance(docstr, str):
```
disrupted, line 18: simplify
```suggestion
    if docstr is None:
```

kpops/utils/docstring.py
```diff
@@ -1,51 +1,48 @@
...
+        return ""
 
+    docstr = docstr.partition(f":param {name}:")[2]
     docstr = _trim_description_end(docstr)
 
     return docstr
```
disrupted: collapse to one line
```suggestion
    return _trim_description_end(docstr)
```

kpops/utils/docstring.py
```diff
@@ -1,51 +1,48 @@
...
     """Return class description from its docstring
 
     Excludes parameters and return definitions
 
     Works only with reStructuredText docstrings.
 
-    :param class_: Class whose description is to be isolated or the class' docstring
-    :type class_: type[BaseModel], str
+    :param class_docstr: Class whose description is to be isolated or the class' docstring
+    :type class_docstr: type[BaseModel], str, None
```
disrupted, line 35:
```suggestion
    :type class_docstr: str, None
```

kpops/utils/docstring.py
```diff
@@ -1,51 +1,48 @@
...
+def describe_class(class_docstr: str | None) -> str:
     """Return class description from its docstring
 
     Excludes parameters and return definitions
 
     Works only with reStructuredText docstrings.
 
-    :param class_: Class whose description is to be isolated or the class' docstring
-    :type class_: type[BaseModel], str
+    :param class_docstr: Class whose description is to be isolated or the class' docstring
```
disrupted, line 34:
```suggestion
    :param class_docstr: The class' docstring
```

kpops/utils/docstring.py
```diff
@@ -1,51 +1,48 @@
...
     """
 
-    if isinstance(class_, str):
-        class_doc = _trim_description_end(class_)
-    else:
-        class_doc = _trim_description_end(class_.__doc__)
+    class_docstr = _trim_description_end(class_docstr)
 
-    return class_doc
+    return class_docstr
```
disrupted: I would suggest handling the `None` case as an early return here

this way you don't have to deal with it in `_trim_description_end`

kpops/utils/docstring.py
```diff
@@ -54,18 +51,10 @@ def _trim_description_end(desc: str) -> str:
...
     """
-    desc_enders: list[str] = [
-        ":param ",
-        ":type ",
-        ":return:",
-        ":rtype:",
-        "defaults to ",
-    ]
-
     try:
```
disrupted, line 58: if you make the `desc` a `str`, i.e. handle `None` cases outside this function (see comment above for reference), you can get rid of the whole try-except block

kpops/utils/docstring.py
```diff
@@ -74,6 +63,14 @@ def _trim_description_end(desc: str) -> str:
         log.debug("Returned an empty string.", exc_info=True)
         return ""
 
+    desc_enders: list[str] = [
```
disrupted, line 66: just fyi: type annotation is not necessary here as it can be easily inferred by the type checker

disrupted (changes requested):

kpops/utils/docstring.py
sujuka99: True

kpops/utils/docstring.py
```diff
@@ -0,0 +1,75 @@
...
+
+    end_index = len(desc)
+
+    for desc_ender in desc_enders:
+        if (desc_ender in desc) and (desc.index(desc_ender) < end_index):
+            end_index = desc.index(desc_ender)
+
+    desc = desc[:end_index].strip().rstrip(",")
+
+    return desc
```
disrupted, line 75: try to avoid adding too many empty newlines if there is no point for separation. in this case you could also collapse it to return directly

kpops/utils/docstring.py
```diff
@@ -0,0 +1,75 @@
...
+
+    Works only with reStructuredText docstrings.
+
+    :param docstr: The class' docstring
+    :type docstr: str, None
+    :returns: Class description taken from its docstiirng
+    :rtype: str
+    """
+    if docstr is None:
+        log.debug("Returned an empty string.", exc_info=True)
```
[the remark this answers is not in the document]
sujuka99: It's a remnant of my trials to put extra information about where exactly a description is missing.

I never managed to find how to do that, so the log is not useful as it is, you are right.

kpops/utils/docstring.py
sujuka99: Okay, I never know where to have an emtpy newline. Is there a convention for that?

sujuka99: Sure thing

docs/docs/schema/pipeline.json
```diff
@@ -204,42 +244,51 @@
...
                 },
                 "caFile": {
                     "title": "Cafile",
+                    "description": "Certificate file",
                     "type": "string",
                     "format": "path"
                 },
                 "insecureSkipTlsVerify": {
                     "title": "Insecureskiptlsverify",
+                    "description": "If true, Kubernetes API server's certificate will not be checked for validity\n        ",
```
disrupted, line 268: nitpick: would it be possible to strip trailing whitespace (newlines and spaces)?

sujuka99: tbh I have no idea why they don't get stripped

disrupted: perhaps we should trim all linebreaks in general and let the the editor take care of it. wdyt?

disrupted: we can also do it in a followup, perhaps it's better to see a diff on this

disrupted (approved):
LGTM

docs/docs/schema/pipeline.json
sujuka99: I agree. I managed to remove the newlines characters, but I can't seem to remove the whitespaces..

sujuka99: > perhaps we should trim all linebreaks in general and let the the editor take care of it. wdyt?

I definitely agree

disrupted (approved):

PR #196, opened 2023-04-20
Respect object docstring titles

sujuka99:
Also rename function describe_class more accurately as it can be used effectively on any string.

### DIFF
--- a/kpops/utils/docstring.py
+++ b/kpops/utils/docstring.py
@@ -22,7 +22,7 @@ def describe_attr(name: str, docstr: str | None) -> str:
 
 
 def describe_object(docstr: str | None) -> str:
-    """Return free-text description from a docstring
+    """Return description from an object's docstring
 
     Excludes parameters and return definitions
 
@@ -30,7 +30,7 @@ def describe_object(docstr: str | None) -> str:
 
     :param docstr: The docstring
     :type docstr: str, None
-    :returns: Description taken from the docstiirng
+    :returns: Description taken from the docstring
     :rtype: str
     """
     if docstr is None:
@@ -38,7 +38,8 @@ def describe_object(docstr: str | None) -> str:
 
     # reST docstrings have a short description/title as their first line.
     # Optionally, they have a longer description below. Here we separate the
-    # title from the rest as `_trim_description_end()` removes all newlines.
+    # title from the rest with a newline as `_trim_description_end()`
+    # removes all newlines.
     docstr = docstr.strip()
     if "\n" in docstr:
         title_end = docstr.index("\n")
--- a/kpops/utils/docstring.py
+++ b/kpops/utils/docstring.py
@@ -53,8 +53,9 @@ def _trim_description_end(desc: str) -> str:
     Also removes all whitespaces and newlines and replaces them with a single space.
 
     A description is defined here as a string of text written in natural language.
-    A description ends at the occurence of a separator such as `:returns:`.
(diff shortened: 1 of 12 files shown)


kpops/utils/docstring.py
```diff
@@ -21,29 +21,39 @@ def describe_attr(name: str, docstr: str | None) -> str:
...
 
     Excludes parameters and return definitions
 
     Works only with reStructuredText docstrings.
 
-    :param docstr: The class' docstring
+    :param docstr: The docstring
     :type docstr: str, None
-    :returns: Class description taken from its docstiirng
+    :returns: Description taken from the docstiirng
```
disrupted, line 33: typo 🙃

kpops/utils/docstring.py
```diff
@@ -21,29 +21,39 @@ def describe_attr(name: str, docstr: str | None) -> str:
     return _trim_description_end(docstr)
 
 
-def describe_class(docstr: str | None) -> str:
-    """Return class description from its docstring
+def describe_object(docstr: str | None) -> str:
+    """Return free-text description from a docstring
```
disrupted, line 25: Return description from an object's docstring

kpops/utils/docstring.py
```diff
@@ -21,29 +21,39 @@ def describe_attr(name: str, docstr: str | None) -> str:
...
         return ""
+
+    # reST docstrings have a short description/title as their first line.
+    # Optionally, they have a longer description below. Here we separate the
+    # title from the rest as `_trim_description_end()` removes all newlines.
+    docstr = docstr.strip()
+    if "\n" in docstr:
+        title_end = docstr.index("\n")
+        desc = docstr[:title_end] + "\n" + _trim_description_end(docstr[title_end:])
+        return desc.rstrip()
```
disrupted, line 46: so this removes the first line of a docstring and then only selects the part below until the attributes section?

kpops/utils/docstring.py
disrupted: not sure what free-text means tbh

kpops/utils/docstring.py
```diff
@@ -21,30 +21,41 @@ def describe_attr(name: str, docstr: str | None) -> str:
...
 
 
 def _trim_description_end(desc: str) -> str:
     """Remove the unwanted text that comes after a description in a docstring
 
+    Also removes all whitespaces and newlines and replaces them with a single space.
+
     A description is defined here as a string of text written in natural language.
-    A description ends at the occurence of a separator such as `:returns:`
+    A description ends at the occurence of a separator such as '\:returns\:'.
```
disrupted, line 56: why is `:` escaped?

sujuka99: words surrounded by `:` are special and not rendered correctly by the lsp

kpops/utils/docstring.py
sujuka99: No, it separates the title from the rest with a newline. Otherwise you end up with descriptions like:

image

kpops/utils/docstring.py
sujuka99: Now I'm uncertain, too. Might not be correct in English

Anyway, I replaced it with yours, it's simpler

kpops/utils/docstring.py
disrupted: makes sense, thanks for explaining

disrupted (approved):

PR #197, opened 2023-04-21
Generate schema in CI

### LINKED ISSUE #71: Generate schema in CI
Create a new CI workflow that runs from main to re-generate the kpops schemas (for pipeline.yaml and config.yaml), then commit and push those changes.

sujuka99:
closes #71


sujuka99:
TODO: See if it's possible to order the schema upon generation. see [json.dumps()](https://docs.python.org/3/library/json.html#json.dumps)

.github/workflows/ci.yaml
```diff
@@ -40,6 +40,12 @@ jobs:
       - name: Typing (mypy)
         run: poetry run pre-commit run mypy --all-files
 
+      - name: Generate schema (kpops schema)
+        run: poetry run pre-commit run gen-schema --all-files --show-diff-on-failure
+
+      - name: Generate docs (typer-cli)
+        run: poetry run pre-commit run gen-docs --all-files --show-diff-on-failure
```
disrupted, line 47: great! 🎉

disrupted: doesn't the docs part also need the docs dependencies installed?

.github/scripts/run_mypy.sh
```diff
@@ -3,7 +3,10 @@
 
 set -o errexit
 
+scriptDir=$(realpath ${BASH_SOURCE[0]} --canonicalize-existing)
+rootDir=${scriptDir%/*/*/*}
+
 # change directory to the project root directory.
-cd "$(dirname "$0")"
+cd $rootDir
```
disrupted, line 10: why is this change necessary? and what does it do exactly?

docs/docs/user/references/cli-commands.md
```diff
@@ -179,4 +179,5 @@ $ kpops schema [OPTIONS] SCOPE:{pipeline|config} [COMPONENTS_MODULE]
 
 **Options**:
 
+* `--include-stock-components / --no-include-stock-components`: Include the built-in KPOps components.  [default: include-stock-components]
```
disrupted, line 182: why have this as an option, like what's the use case for excluding the builtin components?

.github/scripts/run_mypy.sh
sujuka99: It reads the script's location and goes up 3 times to reach the project dir.

`BASH_SOURCE[0]` should have the same value as `$0`, but I read that `BASH_SOURCE[0]` is more likely to always produce the wanted result (which is the script's parent directory)

`realpath` turns it from a relative to an absolute path

finally `${scriptDir%/*/*/*}` is [parameter expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html#Shell-Parameter-Expansion) that cuts off the string such that we end up with the path 3 times up. I could have probably written `cd ../../..` instead of that/

docs/docs/user/references/cli-commands.md
sujuka99: tbh testing is my reason to add it, but perhaps someone might want to only have a schema for that if they are using only custom components

Do you think I should remove it? I couldn't figure out a better way to abstract from the implementation of the stock components when testing the function.

I hate that I constantly have to generate new snapshots and I would argue that it makes it easier for errors to slip through.

.github/workflows/ci.yaml
sujuka99: It did, but in order to run the pre-commit locally and more importantly for it to fail, you needed to have them installed and I moved the `typer-cli` dependency to the dev dependencies.

docs/docs/user/references/cli-commands.md
disrupted: ah ok, I see what you mean. tbf I wouldn't say it has value to regular users, but I agree with the testability argument

.github/scripts/run_mypy.sh
sujuka99: I forgot to mention, I moved the script, that's the reason I had to change the contents.

disrupted: hm, I am not sure if moving the script inside the github directory makes sense, since it is meant to be used locally as well as for the CI.

sujuka99: I just changed the [path](https://github.com/bakdata/kpops/pull/197/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9L36)

sujuka99: 🤔 I think I get it, it shouldn't be in a github-specific folder, right? Should I move them to a dir outside or do you prefer for me to just leave them in the root directory

hooks/run_mypy.sh
```diff
@@ -4,7 +4,7 @@
 set -o errexit
 
 scriptDir=$(realpath ${BASH_SOURCE[0]} --canonicalize-existing)
```
disrupted, line 6: what's your opinion on naming conventions? I think in bash it's common to use snake_case variable names or CONSTANT_CASE for environment stuff

sujuka99: I am a big proponent. You are right, I just checked, I will fix the names. Should have done that in the beginning 🤔

docs/docs/schema/config.json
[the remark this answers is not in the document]
disrupted: that's great!

kpops/utils/gen_schema.py
```diff
@@ -76,32 +76,59 @@ def is_valid_component(
         return True
 
 
-def gen_pipeline_schema(
-    components_module: str | None = None,
-) -> None:
+def _add_components(components_module: str, components: tuple = ()) -> tuple:
```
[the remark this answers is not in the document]
sujuka99: I am now overwhelmed by mypy errors of all kinds :X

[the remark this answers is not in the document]
disrupted: I think the type checker doesn't really like the empty tuple generation as `()`, if I change it to `tuple()` it passes for Pylance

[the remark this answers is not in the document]
disrupted: btw I am getting
```
realpath: --canonicalize-existing: No such file or directory
```

when I run `pre-commit run -a mypy`

kpops/utils/gen_schema.py
```diff
@@ -76,7 +76,9 @@ def _is_valid_component(
         return True
 
 
-def _add_components(components_module: str, components: tuple = ()) -> tuple:
+def _add_components(
+    components_module: str, components: tuple[type[PipelineComponent]] | None = None
+) -> tuple:
```
disrupted, line 81: return type as well 🙃

kpops/utils/gen_schema.py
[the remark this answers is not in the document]
sujuka99: Hmm, you are using `zsh`, right? Could that be the reason?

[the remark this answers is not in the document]
sujuka99: `mypy` doesn't like the tuple being empty at all it seems.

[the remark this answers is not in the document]
sujuka99: It is also concerned about the `components` being none despite the logical check.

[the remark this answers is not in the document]
disrupted: that's weird, it sounds like it doesn't perform type narrowing correctly or it has a very strict idea of how tuples should be typed/used. could you check if there are any related tickets on the mypy issue tracker?

hooks/gen_schema.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+ROOT_DIR=$(realpath ${0%/*/*} --canonicalize-existing)
```
disrupted, line 3: the version of `realpath` that ships with macOS by default also doesn't have the `--canonicalize-existing` flag. is it necessary?

sujuka99: We can do away with it.

kpops/utils/gen_schema.py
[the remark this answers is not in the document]
sujuka99: I would use the `assert` statement, but I am still left with the following:

```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

kpops/utils/gen_schema.py:96: error: Incompatible types in assignment
(expression has type "Tuple[<nothing>, ...]", variable has type
"Optional[Tuple[Type[PipelineComponent]]]")  [assignment]
            components = tuple()
                         ^~~~~~~
kpops/utils/gen_schema.py:110: error: Incompatible types in assignment
(expression has type "Tuple[Type[PipelineComponent], ...]", variable has type
"Optional[Tuple[Type[PipelineComponent]]]")  [assignment]
        components += tuple(custom_components)
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
kpops/utils/gen_schema.py:137: error: Argument 2 to "_add_components" has
incompatible type "Tuple[Type[PipelineComponent], ...]"; expected
"Optional[Tuple[Type[PipelineComponent]]]"  [arg-type]
            components = _add_components(components_module, components)
                                                            ^~~~~~~~~~
Found 3 errors in 1 file (checked 109 source files)

```

Why would narrowing the type be bad? Besides, I am asserting that the tuple is not None and then mypy still thinks it is of type Optional??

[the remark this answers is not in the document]
sujuka99: I would use the `assert` statement, but I am still left with the following:

```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

kpops/utils/gen_schema.py:96: error: Incompatible types in assignment
(expression has type "Tuple[<nothing>, ...]", variable has type
"Optional[Tuple[Type[PipelineComponent]]]")  [assignment]
            components = tuple()
                         ^~~~~~~
kpops/utils/gen_schema.py:110: error: Incompatible types in assignment
(expression has type "Tuple[Type[PipelineComponent], ...]", variable has type
"Optional[Tuple[Type[PipelineComponent]]]")  [assignment]
        components += tuple(custom_components)
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
kpops/utils/gen_schema.py:137: error: Argument 2 to "_add_components" has
incompatible type "Tuple[Type[PipelineComponent], ...]"; expected
"Optional[Tuple[Type[PipelineComponent]]]"  [arg-type]
            components = _add_components(components_module, components)
                                                            ^~~~~~~~~~
Found 3 errors in 1 file (checked 109 source files)

```

I am asserting that the tuple is not None and then mypy still thinks it is of type Optional??
Also, why is passing a tuple as an arg, instead of an optional tuple flagged?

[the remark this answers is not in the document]
sujuka99: I could silence all errors, but I don;t understand why `mypy` handles tuples like that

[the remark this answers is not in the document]
disrupted: yeah, that's very unfortunate

kpops/utils/gen_schema.py
```diff
@@ -76,32 +76,65 @@ def is_valid_component(
...
+    :param components: Tuple of components to which to add, defaults to ()
+    :type components: tuple, optional
+    :return: Extended tuple
+    :rtype: tuple
     """
-    Generate a json schema from the models of pipeline components.
+    if components is None:
+        components = tuple()  # type: ignore[assignment]
+    # HACK: mypy doesn't narrow the tuple's type without this assertion
+    assert components is not None
```
disrupted, line 98: I guess this is not needed anymore then?

disrupted (changes requested):

tests/cli/test_schema_generation.py
```diff
@@ -82,12 +83,64 @@ class SubPipelineComponentCorrect(SubPipelineComponent):
...
+            app,
+            [
+                "schema",
+                "pipeline",
+                "--no-include-stock-components",
+            ],
+            catch_exceptions=False,
+        )
+        assert result.exit_code == 0
+        snapshot.assert_match(result.stdout, "test-schema-generation")
```
disrupted, line 141: why do we need a snapshot? can't you just `assert result.stdout == "No components are provided, no schema is generated."`

disrupted:
the **gen-docs** hook is failing on my local machine

```
gen-docs.................................................................Failed
- hook id: gen-docs
- exit code: 1
- files were modified by this hook

Docs saved to: /Users/disrupted/bakdata/kpops/docs/docs/user/references/cli-commands.md
sed: 1: "/Users/disrupted/bakdat ...": extra characters at the end of d command
```

might be due to different version of `sed`?!

tests/cli/test_schema_generation.py
```diff
@@ -97,11 +150,34 @@ def test_gen_pipeline_schema_without_custom_module(self, snapshot: SnapshotTest)
             catch_exceptions=False,
         )
 
+        assert result.exit_code == 0
+        assert len(result.stdout) > 0
```
disrupted, line 154: do you think a snapshot for the internal components would make sense?

tests/cli/test_schema_generation.py
```diff
@@ -113,8 +189,7 @@ def test_gen_pipeline_schema_with_custom_module(self, snapshot: SnapshotTest):
         )
 
         assert result.exit_code == 0
-
-        snapshot.assert_match(result.stdout, "test-schema-generation")
+        assert len(result.stdout) > 0
```
disrupted, line 192: snapshot?

kpops/utils/gen_schema.py
sujuka99: This removes 3 more errors haha

Here is what happens without it and with the 3 silencing comments:

```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

kpops/utils/gen_schema.py:102: error: Item "None" of
"Optional[Tuple[Type[PipelineComponent]]]" has no attribute "__iter__" (not
iterable)  [union-attr]
            for component in components
                             ^~~~~~~~~~
kpops/utils/gen_schema.py:110: error: Unsupported left operand type for +
("None")  [operator]
        components += tuple(custom_components)  # type: ignore[assignment]
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
kpops/utils/gen_schema.py:110: note: Error code "operator" not covered by "type: ignore" comment
kpops/utils/gen_schema.py:110: note: Left operand is of type "Optional[Tuple[Type[PipelineComponent]]]"
kpops/utils/gen_schema.py:111: error: Incompatible return value type (got
"Optional[Tuple[Type[PipelineComponent]]]", expected
"Tuple[Type[PipelineComponent]]")  [return-value]
        return components
               ^~~~~~~~~~
Found 3 errors in 1 file (checked 109 source files)
```

sujuka99:
> the **gen-docs** hook is failing on my local machine
> 
> ```
> gen-docs.................................................................Failed
> - hook id: gen-docs
> - exit code: 1
> - files were modified by this hook
> 
> Docs saved to: /Users/disrupted/bakdata/kpops/docs/docs/user/references/cli-commands.md
> sed: 1: "/Users/disrupted/bakdat ...": extra characters at the end of d command
> ```
> 
> might be due to different version of `sed`?!

Yea, it [behaves differently](https://github.com/lmquang/til/issues/18) on mac, I will fix it

tests/cli/test_schema_generation.py
sujuka99: Not really, that's the main reason for adding the external components flag. I don't want the snapshot to update each time a component changes slightly.

Besides, the internal components schema is generated in the pre-commit hooks, so in a way, it replaces the snapshot test imo.

tests/cli/test_schema_generation.py
sujuka99: https://github.com/bakdata/kpops/pull/197/files#r1177753316

tests/cli/test_schema_generation.py
sujuka99: You are right, I will do that.

tests/cli/test_schema_generation.py
disrupted: makes sense

tests/cli/test_schema_generation.py
disrupted, line 192: a more Pythonic version of this would be to just
`assert result.stdout` 🙂

disrupted:
> > might be due to different version of `sed`?!
> 
> Yea, it [behaves differently](https://github.com/lmquang/til/issues/18) on mac, I will fix it

now I am getting

```
gen-docs.................................................................Failed
- hook id: gen-docs
- exit code: 1
- files were modified by this hook

Docs saved to: /Users/disrupted/bakdata/kpops/docs/docs/user/references/cli-commands.md
sed: illegal option -- -
usage: sed script [-Ealnru] [-i extension] [file ...]
        sed [-Ealnu] [-i extension] [-e script] ... [-f script_file] ... [file ...]
```

tests/cli/test_schema_generation.py
```diff
@@ -82,26 +84,107 @@ class SubPipelineComponentCorrect(SubPipelineComponent):
...
+            ],
+            catch_exceptions=False,
+        )
+        assert [
+            (
+                "root",
+                logging.WARNING,
+                "No components are provided, no schema is generated.",
+            )
+        ]
```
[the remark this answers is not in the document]
disrupted: I don't think this works tbh. I don't see an actual comparison in this assertion, all this does is evaluate if the list (containing a tuple of values) is not empty...
if I change the string to something else the test also passes.

however, I also do not know how to capture stderr. It's also not super necessary to test if it's tricky.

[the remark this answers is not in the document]
sujuka99: I believe it works right now

tests/cli/test_schema_generation.py
```diff
@@ -82,26 +84,107 @@ class SubPipelineComponentCorrect(SubPipelineComponent):
...
+    )
+
+
 MODULE = EmptyPipelineComponent.__module__
 
 
 @pytest.mark.filterwarnings("ignore:handlers", "ignore:config", "ignore:enrich")
 class TestGenSchema:
-    def test_gen_pipeline_schema_without_custom_module(self, snapshot: SnapshotTest):
+    def test_gen_pipeline_schema_no_modules(self, caplog):
```
disrupted, line 131: TIL about caplog 😄 
please add a type annotation for the fixture

tests/cli/test_schema_generation.py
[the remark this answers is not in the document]
disrupted: yes, it works now! :)

tests/cli/test_schema_generation.py
sujuka99: done

tests/cli/test_schema_generation.py
```diff
@@ -113,16 +196,13 @@ def test_gen_pipeline_schema_with_custom_module(self, snapshot: SnapshotTest):
...
+    def test_gen_config_schema(self):
         result = runner.invoke(
             app,
             ["schema", "config"],
             catch_exceptions=False,
         )
 
         assert result.exit_code == 0
-
-        snapshot.assert_match(result.stdout, "test-schema-generation")
```
disrupted, line 128: we should at least have an `assert result.stdout` here

disrupted (changes requested):
last one, otherwise LGTM

disrupted (approved):

PR #198, opened 2023-04-24
Refactor variable substitution

### LINKED ISSUE #102: Variable substitution does not work on Windows
We use os.environ in order to substitute internal variables such as `pipeline_name`: https://github.com/bakdata/kpops/blob/8f5e1bdc7f4ddf13e633259908fd1d9bcc4b34e0/kpops/pipeline_generator/pipeline.py#L90. However, when calling `dict(os.environ)`, Windows returns a dictionary containing `{"PIPELINE_NAME": ...}`. Therefore, substitution does not work. We should find another way to do the substitution. In my opinion, we should not add our internal variables to os.environ but rather add os.environ to our dictionary of variables.

### LINKED ISSUE #115: Refactor variable substitution
Currently, variables are replaced in predefined fields in the init method: https://github.com/bakdata/kpops/blob/0ca26d6ad5d5b37780de13c646b871fec0a54e7f/kpops/components/base_components/pipeline_component.py#L40. We should consider doing that automatically for any field.

Also have a look at #102

### LINKED ISSUE #119: Refactor input/output types
Currently, we have two types for input topics and extra output topics. However, we could use the same type but just distinguish if role is set or not because role is mandatory for extra types. The same can be applied to input pattern and output type. See https://github.com/bakdata/kpops/blob/aa44fdf6520a4ef9c520b27cfdb94e9fc35cd079/kpops/components/base_components/pipeline_component.py#L117

sujuka99:
closes #115


sujuka99:
### All substitution functions

- [`yaml_loading.substitute`](https://github.com/bakdata/kpops/blob/f6baeeede78cfb7a1cefcb5519c36a7ac9756b8c/kpops/utils/yaml_loading.py#L25)
	- [`PipelineComponent.substitute_component_names`](https://github.com/bakdata/kpops/blob/f6baeeede78cfb7a1cefcb5519c36a7ac9756b8c/kpops/components/base_components/pipeline_component.py#L99)
		- [`PipelineComponent.substitute_component_variables`](https://github.com/bakdata/kpops/blob/ab8355dfe67065f3f5943a35825588dfee4c3b57/kpops/components/base_components/pipeline_component.py#L112)
			- [`PipelineComponent.substitute_output_topic_names`](https://github.com/bakdata/kpops/blob/ab8355dfe67065f3f5943a35825588dfee4c3b57/kpops/components/base_components/pipeline_component.py#L89)
			- [`StreamsApp.__substitute_autoscaling_topic_names`](https://github.com/bakdata/kpops/blob/ab8355dfe67065f3f5943a35825588dfee4c3b57/kpops/components/streams_bootstrap/streams/streams_app.py#L110)
		- [`PipelineComponent.substitute_name`](https://github.com/bakdata/kpops/blob/ab8355dfe67065f3f5943a35825588dfee4c3b57/kpops/components/base_components/pipeline_component.py#L262)
	- [`PipelineComponent.substitute_prefix`](https://github.com/bakdata/kpops/blob/f6baeeede78cfb7a1cefcb5519c36a7ac9756b8c/kpops/components/base_components/pipeline_component.py#L329)
	- [`Pipeline.substitute_component_specific_variables`](https://github.com/bakdata/kpops/blob/f6baeeede78cfb7a1cefcb5519c36a7ac9756b8c/kpops/pipeline_generator/pipeline.py#L233)

_A nested function contains a call to its parent in the list._

actions/kpops-runner/README.md
```diff
@@ -11,7 +11,7 @@ This action runs KPOps with the given config.
 | kpops-version     |    ✅    |       -       | string | KPOps version to use                                                                                                                          |
 | working-directory |    ❌    |       .       | string | root directory used by KPOps to run pipelines                                                                                                 |
 | pipeline-base-dir |    ❌    |       .       | string | directory where relative pipeline variables are initialized from                                                                              |
-| defaults          |    ❌    |   defaults    | string | defaults folder path                                                                                                                          |
+| defaults          |    ❌    |       .       | string | defaults folder path                                                                                                                          |
```
philipp94831, line 14: Unrelated to this PR?

sujuka99: true, I will move it into a separate one

I did it to make manual testing simpler

kpops/components/base_components/pipeline_component.py
```diff
@@ -80,59 +78,9 @@ class Config(CamelCaseConfig, DescConfig):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
-        self.substitute_output_topic_names()
-        self.substitute_name()
-        self.substitute_prefix()
```
disrupted, line 85: 🤩

disrupted (changes requested):

kpops/components/base_components/base_defaults_component.py
```diff
@@ -213,17 +213,23 @@ def deduplicate(seq: Sequence[T]) -> list[T]:
...
+    :param other_dict: Mapping that contains new key-value pairs
     :type other_dict: Mapping
     :return: Updated dictionary
     :rtype: dict
     """
     for key, value in other_dict.items():
         if isinstance(value, Mapping):
             nested_val = original_dict.get(key, {})
-            if nested_val is not None:
+            if nested_val is not None and isinstance(nested_val, dict):
```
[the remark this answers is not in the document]
sujuka99: I will. Note that this does not solve the issue that @raminqaf and @philipp94831 reported with wrong defaults loading, at least not intentionally.

kpops/components/base_components/kubernetes_app.py
```diff
@@ -88,7 +87,6 @@ class Config(CamelCaseConfig, DescConfig):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
-        self.__check_compatible_name()
```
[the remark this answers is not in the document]
sujuka99: The problem is that the substitution happens after at least one init of the component and a placeholder of the form `${foo}` is not a valid Kubernetes name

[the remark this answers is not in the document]
sujuka99: To solve it, I would either have to create a new parameter in the KubernetesApp or figure out a way to deduce from the existing arguments whether it's the final initialization of the component or not

kpops/components/base_components/base_defaults_component.py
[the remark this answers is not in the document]
sujuka99: Added tests, lmk if you think more are needed

kpops/components/base_components/kubernetes_app.py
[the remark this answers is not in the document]
disrupted: I don't have a precise answer for you. About the _final initialization_ problem I can only add that this could be a call  in `enrich_component` after the component is initialized

kpops/components/base_components/base_defaults_component.py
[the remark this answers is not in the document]
disrupted: exactly what I had in mind

kpops/components/base_components/kubernetes_app.py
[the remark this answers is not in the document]
disrupted: the simplest solution I can think of is to add a generic `validate` method to the base component and call it from `enrich_component`. Of course it would be nicer if this could use Pydantic's builtin validation somehow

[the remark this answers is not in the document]
sujuka99: I added a validator, but I honestly don't like the solution much. Unfortunately I cannot think of a better way and simpler way to do it.

One side effect is that now the generated values of the connector contain a useless field `validate_name`. I could combat this by moving the validate_name class attribute from KubernetesApp to PipelineComponent, but then it introduces a KubernetesApp-specific field to all components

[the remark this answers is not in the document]
sujuka99: Currently, I made it so that the name is validated by default and you have to deliberately opt-out. The reason is simple - to avoid introducing bugs in the future by forgetting to request validation.

kpops/pipeline_generator/pipeline.py
```diff
@@ -194,26 +230,58 @@ def enrich_component(
...
-        return component_class(
+        enriched_component = component_class(
             enrich=False,
             config=self.config,
             handlers=self.handlers,
             **component_data,
         )
 
+        return enriched_component
+
```
raminqaf, line 257: Why not inline the return value?

kpops/pipeline_generator/pipeline.py
```diff
@@ -230,44 +298,80 @@ def __str__(self) -> str:
...
+            # "component_name": component.name,
+            # "component_type": component.type,
+            "error_topic_name": component.config.topic_name_config.default_error_topic_name,
+            "output_topic_name": component.config.topic_name_config.default_output_topic_name,
+            "schema_registry_url": component.config.schema_registry_url,
+            "broker": component.config.broker,
+        }
+        substitution = gen_substitution(component, "component", substitution_hardcoded)
+
+        # TODO: Should this be allowed? Probably not, delete after discussion.
```
raminqaf, line 325: Already discussed?

kpops/components/base_components/kubernetes_app.py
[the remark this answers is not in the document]
disrupted: do we have to have the `validate_name` attribute for the `root_validator` to work? the validation doesn't have to be opt-out, it should always be enabled.

kpops/pipeline_generator/pipeline.py
```diff
@@ -276,3 +380,43 @@ def set_pipeline_name_env_vars(base_dir: Path, path: Path) -> None:
         os.environ["pipeline_name"] = pipeline_name
         for level, parent in enumerate(path_without_file):
             os.environ[f"pipeline_name_{level}"] = parent
+
+
+# TODO: Does it belong here? Does it need dedicated tests? Substitution in the
+# pipeline is already covered.
+def gen_substitution(
```
raminqaf, line 387:
```suggestion
def generate_substitution(
```

tests/components/test_pipeline_component.py
```diff
@@ -21,57 +24,58 @@ class PlainPipelineComponent(PipelineComponent):
...
-                        type=OutputTopicTypes.ERROR, partitions_count=10
-                    ),
-                    "${output_topic_name}": TopicConfig(
-                        type=OutputTopicTypes.OUTPUT, partitions_count=10
-                    ),
-                },
-            ),
-        )
+    pass
+    # def test_topic_substitution(self):
```
raminqaf, line 28: What is happening here?

kpops/components/base_components/kubernetes_app.py
```diff
@@ -86,9 +94,10 @@ class KubernetesApp(PipelineComponent):
     class Config(CamelCaseConfig, DescConfig):
         pass
 
+    # When removed, mypy throws some errors about missign arguments in
```
disrupted, line 97: I can't reproduce this locally. are you sure this is still the case?

raminqaf, line 97: When what is removed?

kpops/components/base_components/kubernetes_app.py
```diff
@@ -182,12 +191,17 @@ def get_helm_chart(self) -> str:
...
         )
 
-    def __check_compatible_name(self) -> None:
-        """Check if the component's name `self.name` is valid for Kubernetes"""
-        if not bool(KUBERNETES_NAME_CHECK_PATTERN.match(self.name)):  # TODO: SMARTER
-            raise ValueError(
-                f"The component name {self.name} is invalid for Kubernetes."
-            )
+    @root_validator()
+    def name_must_be_valid_for_kubernetes(cls, values) -> None:
```
disrupted, line 195:
```suggestion
    def name_must_be_valid_for_kubernetes(cls, values: dict) -> dict:
```

raminqaf (changes requested):

kpops/components/base_components/kubernetes_app.py
disrupted: also consider a shorter name

kpops/components/base_components/kubernetes_app.py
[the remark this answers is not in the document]
sujuka99: The problem is that when the name contains a placeholder, it is not valid. Explicitly adding placeholders to the allowed string format does not work, because it is not necessary that the placeholder would be substituted

kpops/components/base_components/kubernetes_app.py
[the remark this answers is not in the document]
sujuka99: Good catch!

kpops/components/base_components/kubernetes_app.py
sujuka99: Forgot to delete the file, was moving tests one by one

kpops/components/base_components/kubernetes_app.py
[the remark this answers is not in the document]
disrupted: ok I see :/ so your proposed workaround would be to set `validate_name=False` if the user wants to use placeholders in the name of a kubernetes-app component?

kpops/components/base_components/kubernetes_app.py
sujuka99: the init method, strangely mypy doesnt mind it now, deleted

sujuka99: It strangely isn't. I removed it.

kpops/components/base_components/kubernetes_app.py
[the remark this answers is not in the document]
sujuka99: > if it's not possible to do it conveniently with Pydantic's builtin validators this could just be a custom method `validate` that checks it

The difference would only be that instead of being decorated with `@root_validator`, it would have to be called in `__init__` on itself. I suppose that would be a bit more efficient?

kpops/pipeline_generator/pipeline.py
```diff
@@ -276,3 +380,43 @@ def set_pipeline_name_env_vars(base_dir: Path, path: Path) -> None:
         os.environ["pipeline_name"] = pipeline_name
         for level, parent in enumerate(path_without_file):
             os.environ[f"pipeline_name_{level}"] = parent
+
+
+# TODO: Does it belong here? Does it need dedicated tests? Substitution in the
```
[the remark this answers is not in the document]
sujuka99: added

sujuka99:
### Overview

#### Substitution before

1. Load `pipeline.yaml`
2. Substitute `os.environ` upon reading file
3. Load `pipeline_env.yaml`
4. Substitute `os.environ` and `component_type` in env-specific definition during the creation of the env-specific component index
5. For each component:
	1. Substitute `${component_type}` and `${component_name}` in env def
	2. Init merged component
		1. Substitute `os.environ` in the defaults and apply them
		2. Substitute `${component_name}`, `${error_topic_name}`, `${output_topic_name}`, `${component_type}` ONLY in `self.to.topics`
		3. Substitute `${component_type}` ONLY `in self.name`
		4. Substitute `os.environ` ONLY in `self.prefix`
		5. IF `KafkaApp` 
			- substitute `${broker}` ONLY in `self.app.streams.brokers` with `self.config.broker`
			- substitute `${schema_registry_url} ONLY in `self.app.streams.schema_registry_url` with `self.config.schema_registry_url`
		6. IF `StreamsApp` substitute `${component_name},` `${error_topic_name},` `${output_topic_name}`, `${component_type}` ONLY in `self.app.autoscaling.topics` and `self.app.autoscaling.consumer_group`
		7. IF `KafkaConnector` substitute  `${component_name}`, `${error_topic_name}`, `${output_topic_name}` ONLY in `self.app`

#### Substitution after

kpops/pipeline_generator/pipeline.py
```diff
@@ -194,26 +230,58 @@ def enrich_component(
...
+        return enriched_component
+
     def print_yaml(self, substitution: dict | None = None) -> None:
+        """Print the generated pipeline definition
+
+        :param substitution: Substitution dictionary, defaults to None
+        :type substitution: dict | None, optional
+        """
+        # Enables cross-referencing between components.
+        # TODO: Discuss whether it is needed
```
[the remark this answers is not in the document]
sujuka99: At the meeting we decided that it would be moved to a separate PR

kpops/components/base_components/kubernetes_app.py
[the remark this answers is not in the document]
irux: isn't it easier to expand the `check_compatible_name` and then filter the place holders and proof validation ?

kpops/pipeline_generator/pipeline.py
```diff
@@ -229,45 +284,71 @@ def __str__(self) -> str:
...
+        # Generate a substitution dict from the pipeline config and combine it
+        # with the component-specifig substitution
+        substitution = generate_substitution(
+            json.loads(config.json()), "", component_substitution
+        )
+        # Substitute all vars in the component, avoid duplicates, prioritize component def
+        substituted_component: dict = json.loads(
+            substitute_nested(
+                json.dumps(component_as_dict),
+                **update_nested_pair(substitution, os.environ),
```
raminqaf, line 321: I am not a fan of having so many comments written in the code. This is not a good practice since the comment can get outdated, which leads to misleading information. If you think your method is too long and hard to understand each part, break each part into, smaller methods with good names, and docs.

kpops/pipeline_generator/pipeline.py
```diff
@@ -276,3 +357,32 @@ def set_pipeline_name_env_vars(base_dir: Path, path: Path) -> None:
...
+    :param substitution: existing substitution to include
+    :type substitution: dict
+    :returns: Substitution dict of all variables related to the model.
+    :rtype: dict
+    """
+    # Check if given existing substitution is a dict
+    if existing_substitution and not isinstance(existing_substitution, dict):
+        raise TypeError("Argument existing_substitution must be a dict")
+    elif not existing_substitution:
+        existing_substitution = {}
```
raminqaf, line 386: For example, here you can do something like that :
```python
def __is_substitution_dict(existing_substitution: dict) -> dict
    if existing_substitution and not isinstance(existing_substitution, dict):
        raise TypeError("Argument existing_substitution must be a dict")
    elif not existing_substitution:
        return {}
```
P.S. haven't checked the syntax properly but you get the point

raminqaf (changes requested):

kpops/pipeline_generator/pipeline.py
sujuka99: I agree with you, I planned on removing them after the reviews. I expected that I wouldn't do too well explaining my solution otherwise.

tests/pipeline/test_pipeline.py
```diff
@@ -198,6 +198,24 @@ def test_substitute_in_component(self, snapshot: SnapshotTest):
 
         snapshot.assert_match(enriched_pipeline, "test-pipeline")
 
+    @pytest.mark.timeout(10)
```
raminqaf, line 201: Why do we need this?

kpops/components/base_components/kubernetes_app.py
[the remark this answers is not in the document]
sujuka99: @irux I am not sure if I understand, what do you mean with expand?

tests/pipeline/test_pipeline.py
sujuka99: Update: the infinite loop is avoided by simply counting the iterations and a Warning is displayed every 10 seconds to notify the user what is happening in case the set maximum repetitions are too high

kpops/pipeline_generator/pipeline.py
sujuka99: done

kpops/pipeline_generator/pipeline.py
sujuka99: Done

kpops/utils/yaml_loading.py
```diff
@@ -37,3 +39,54 @@ def substitute(input: str, substitution: Mapping[str, Any] | None = None) -> str
...
+    :param input: The raw input containing $-placeholders
+    :type input: str
+    :param max_repetitions: The max number of consecutive substitutions to be carried out, defaults to 500
+    :type max_repetitions: int, optional
+    :param **kwargs: Substitutions
+    :raises Exception: Substitution was repeated {counter} times and placeholders still exist, check for loops
+    :return: Substituted input string
+    :rtype: str
+    """
+    if not {**kwargs}:
```
disrupted, line 74: I believe `{**kwargs}` is unnecessary, just use `kwargs` directly

disrupted (changes requested):

kpops/utils/yaml_loading.py
sujuka99: indeed, will fix it

kpops/utils/dict_ops.py
```diff
@@ -0,0 +1,129 @@
...
+    :type ex_msg: str, optional
+    :raises TypeError: Argument must be of type dict or None
+    :returns: unmodified ``d`` or empty ``dict`` if d is ``None``
+    """
+    if d is None:
+        return {}
+    elif isinstance(d, dict):
+        return d
+    else:
+        raise TypeError(ex_msg)
```
disrupted, line 129: why would it not be `dict | None`? could you give an example for such a scenario?

sujuka99: See [here](https://github.com/bakdata/kpops/pull/198/files#diff-a7e11ead0d843c3b75b7427119e68b95eb41970f919eff2cf96e1111ff485822R86-R109)

I [extracted](https://github.com/bakdata/kpops/pull/198/commits/a0e6ffe1bcd027aba99fe8cdf921f0f2afe1d5b9#diff-a7e11ead0d843c3b75b7427119e68b95eb41970f919eff2cf96e1111ff485822L102-L109) the check out of `generate_substitution()` following [this discussion](https://github.com/bakdata/kpops/pull/198/files/1823e6a1be8961046c03af13076be028d09396f4#r1211217053)

disrupted: yeah, I was wondering because there it is defined as `existing_substitution: dict | None = None`. so then how could it be of some other type?

kpops/utils/yaml_loading.py
sujuka99: done

kpops/utils/dict_ops.py
sujuka99: It is a bit of a defensive approach. Tbh I prefer to return the check inside of the function that sits right above it

This function seems a bit too specific to be on its own like that

sujuka99: I simplified it a lot now

kpops/components/base_components/kubernetes_app.py
```diff
@@ -66,6 +68,12 @@ class KubernetesApp(PipelineComponent):
         description=describe_object(__doc__),
         exclude=True,
     )
+    validate_name: bool = Field(
+        default=True,
+        description=describe_attr("check_name", __doc__),
```
disrupted, line 73: outdated name? shouldn't it be the same name as the attribute itself?

kpops/pipeline_generator/pipeline.py
```diff
@@ -229,45 +273,64 @@ def __str__(self) -> str:
...
+
+        :param component_as_dict: Component represented as dict
+        :type component_as_dict: dict
+        :return: Updated component
+        :rtype: dict
+        """
+        config = self.config
+        # Leftover variables that were previously introduced in the component by the substitution
+        # functions, still hardcoded, because of their names.
+        # TODO: Get rid of them with other breaking changes
```
disrupted, line 287: would it be possible to change the config option from default_error_topic_name → error_topic_name and avoid a breaking change by setting alias="default_error_topic_name"?

kpops/pipeline_generator/pipeline.py
```diff
@@ -229,45 +273,64 @@ def __str__(self) -> str:
...
         Moreover, for each sub-path an environment variable is set.
         For example, for a given path ./data/v1/dev/pipeline.yaml the pipeline_name would be
         set to data-v1-dev. Then the sub environment variables are set:
+
         pipeline_name_0 = data
         pipeline_name_1 = v1
         pipeline_name_2 = dev
+
         :param base_dir: Base directory to the pipeline files
+        :type bas_dir: Path
```
disrupted, line 331: typo
```suggestion
        :type base_dir: Path
```

kpops/utils/dict_ops.py
disrupted, line 1: that name 😎

kpops/utils/yaml_loading.py
```diff
@@ -37,3 +39,54 @@ def substitute(input: str, substitution: Mapping[str, Any] | None = None) -> str
...
+    old_str, new_str = "", substitute(input, kwargs)
+    counter = 0
+    starting_time = time.time()
+    while old_str != new_str:
+        old_str, new_str = new_str, substitute(new_str, kwargs)
+        counter += 1
+        current_time = time.time()
+        if current_time - starting_time > 10:
+            logging.warn(
+                f"Substitution has been repeating for 10 seconds, possible loop detected. The program will exit after reaching {max_repetitions} tries."
```
disrupted, line 85: let's have a think once more what are the exact conditions for a doom loop occuring and see if we can come up with a better detection

kpops/components/base_components/kubernetes_app.py
sujuka99: Absolutely shoud

tests/utils/test_dict_ops.py
```diff
@@ -0,0 +1,84 @@
...
+        existing_substitution = {
+            "key1": "Everything",
+            "key2": "work",
+            "key3": "well",
+            "first_half": "${key1}_seems_to_${key2}",
+            "key": "${first_half}_${key3}",
+            "name": "pre-existing-name",
+        }
+        substitution = generate_substitution(model, "prefix", existing_substitution)
+        snapshot.assert_match(substitution, "test-substitution")
```
disrupted, line 84: not necessary to use a snapshot here. I would prefer `assert substitution == { ... }`

tests/utils/test_dict_ops.py
```diff
@@ -0,0 +1,84 @@
...
+                type_="type",
+                field_nested_dict={
+                    "value_is_none": None,
+                    "value_is_str": "str",
+                    "value_is_int": 0,
+                    "value_is_dict": {
+                        "nested_key": "nested_value",
+                    },
+                },
+                problems=99,
```
disrupted, line 72: `problems` is an undefined attribute, is that intentional?

kpops/utils/yaml_loading.py
sujuka99: [related test](https://github.com/bakdata/kpops/pull/198/files#diff-c06c7ddac9f17bd76f0e58bfc3a019ae647f7a305534a96b15e80c83520f777cR167-R185)

[Recognized conditions](https://github.com/bakdata/kpops/pull/198/files#diff-b67554f5da44782699825ea328c85f262297f18162ac18cc10504e8f2e7dc21d)

kpops/components/base_components/kubernetes_app.py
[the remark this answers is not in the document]
irux: What I mean is just add the edge case where it needs to handle place holders and it filters them.

tests/utils/snapshots/snap_test_dict_ops.py
disrupted, line 1: delete this file

disrupted (changes requested):

kpops/components/base_components/kubernetes_app.py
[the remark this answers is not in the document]
sujuka99: The problem is that they are not necessarily going to be substituted and the length of the future name after a successful substitution is also unknown to the component.

kpops/pipeline_generator/pipeline.py
disrupted: ok, then nvm for now

sujuka99: I actually have a proposed work around - the user could alias any variables as they desire. Here is an example:

`export output_topic_name = topic_name_config_default_output_topic_name`
`export error_topic_name = topic_name_config_default_error_topic_name`

sujuka99: #234

kpops/components/base_components/kubernetes_app.py
```diff
@@ -87,8 +95,9 @@ class Config(CamelCaseConfig, DescConfig):
         pass
 
     def __init__(self, **kwargs):
+        if kwargs.get("validate_name", True):
+            self.validate_kubernetes_name(kwargs.get("name", "invalid_name"))
```
disrupted, line 99: are there cases where `name` is not present in `kwargs`?

tests/cli/test_schema_generation.py
```diff
@@ -126,7 +126,9 @@ class SubPipelineComponentCorrectDocstr(SubPipelineComponent):
 MODULE = EmptyPipelineComponent.__module__
 
 
-@pytest.mark.filterwarnings("ignore:handlers", "ignore:config", "ignore:enrich")
+@pytest.mark.filterwarnings(
+    "ignore:handlers", "ignore:config", "ignore:enrich", "ignore:validate_name"
```
disrupted, line 130: what does this do?

tests/pipeline/test_pipeline.py
disrupted: this is technically not needed anymore, right? I would be fine with keeping it in as a fallback in case the tested function is changed, so that pytest automatically fails in those cases. timeout could probably further reduced

tests/pipeline/test_pipeline.py
```diff
@@ -118,12 +120,92 @@ def test_substitute_component_names(self, snapshot: SnapshotTest):
...
+            .relative_to(Path(PIPELINE_BASE_DIR).resolve())
+            .parts[:-1]
+        )
+        if not path_without_file:
+            raise ValueError("The pipeline-base-dir should not equal the pipeline-path")
+        pipeline_name = "-".join(path_without_file)
+
+        assert (
+            enriched_pipeline["components"][0]["name"]
+            == pipeline_name + "-scheduled-producer"
```
disrupted, line 137: can't we assert this using a static string or does it always change?

kpops/pipeline_generator/pipeline.py
```diff
@@ -194,14 +223,22 @@ def enrich_component(
         self,
         component: PipelineComponent,
     ) -> PipelineComponent:
-        env_component_definition = self.env_components_index.get(component.name, {})
-        pair = update_nested_pair(
-            env_component_definition,
+        """Enrich a pipeline component with env-specific config and substitute variables
```
disrupted, line 226: thanks for the docs 🙏🏻

kpops/utils/yaml_loading.py
```diff
@@ -37,3 +37,43 @@ def substitute(input: str, substitution: Mapping[str, Any] | None = None) -> str
...
+    :param **kwargs: Substitutions
+    :raises Exception: Substitution was repeated {counter} times and placeholders still exist, check for loops
+    :return: Substituted input string
+    :rtype: str
+    """
+    if not kwargs:
+        return input
+    old_str, new_str = "", substitute(input, kwargs)
+    steps: set[str] = {old_str}
+    while new_str not in steps:
```
disrupted, line 76: should we raise an exception if the substituted string is detected in the set of seen substitutions? to communicate to the user that their substitution is looping and should be fixed

disrupted: ah no, this is probably not that straightforward

kpops/utils/yaml_loading.py
```diff
@@ -37,3 +37,43 @@ def substitute(input: str, substitution: Mapping[str, Any] | None = None) -> str
     if not substitution:
         return input
     return Template(input).safe_substitute(**substitution)
+
+
+def substitute_nested(input: str, max_repetitions: int = 500, **kwargs) -> str:
```
disrupted, line 42: remove `max_repetitions` since it's not used anymore

disrupted (changes requested):

kpops/utils/yaml_loading.py
disrupted: perhaps you could check if the return value still contains placeholders?

kpops/utils/yaml_loading.py
```diff
@@ -37,3 +37,43 @@ def substitute(input: str, substitution: Mapping[str, Any] | None = None) -> str
...
+    :type max_repetitions: int, optional
+    :param **kwargs: Substitutions
+    :raises Exception: Substitution was repeated {counter} times and placeholders still exist, check for loops
+    :return: Substituted input string
+    :rtype: str
+    """
+    if not kwargs:
+        return input
+    old_str, new_str = "", substitute(input, kwargs)
+    steps: set[str] = {old_str}
```
disrupted, line 75: is the starting value necessary? it's just an empty string at the beginning

sujuka99: Nope, it's a remnant of the previous design, removed it

kpops/utils/yaml_loading.py
sujuka99: I could add a line that compares the last 2 strings after the loop is finished. If they are different, a loop must exist

tests/cli/test_schema_generation.py
sujuka99: Schema generation logs warnings when a class is skipped and I found it annoying to constantly see the warnings in pytest, so I suppressed them.

tests/pipeline/test_pipeline.py
sujuka99: It shouldn't change, will fix it

kpops/utils/yaml_loading.py
sujuka99: done

if it wasn't for that, I could remove the `old_str` var completely

kpops/utils/yaml_loading.py
sujuka99: done

tests/pipeline/test_pipeline.py
sujuka99: done

tests/components/test_pipeline_component.py
```diff
@@ -1,85 +1 @@
...
-                ),
-            ),
-            handlers=ComponentHandlers(
-                schema_handler=MagicMock(),
-                connector_handler=MagicMock(),
-                topic_handler=MagicMock(),
-            ),
-        )
-        assert prefix_to_match + "-" == pipeline_component.prefix
+pass
```
disrupted, line 1: what happened here?

kpops/utils/yaml_loading.py
disrupted: also from the docstring

kpops/components/base_components/kubernetes_app.py
disrupted: then you can just access it like `kwargs["name"]`

tests/components/test_pipeline_component.py
sujuka99: see:
[`56b46dd` (#198)](https://github.com/bakdata/kpops/pull/198/commits/56b46ddd583ad06978a26efab89564e6e3fe7d17)
[`006f16a` (#198)](https://github.com/bakdata/kpops/pull/198/commits/006f16a899ef7bea5d3b6956c7d26b8b7bf281b4)
https://github.com/bakdata/kpops/pull/198#discussion_r1202474325

I moved the tests, should I delete the file?

disrupted: yeah, please delete it

kpops/utils/yaml_loading.py
```diff
@@ -7,6 +8,8 @@
 from cachetools import cached
 from cachetools.keys import hashkey
 
+log = logging.getLogger("")
```
disrupted, line 11: unused?

disrupted (dismissed):

kpops/utils/yaml_loading.py
sujuka99: I think that would happen only if `new_str` is somehow found to be in the empty set `steps`. Is that possible?

sujuka99: I can revert back to assigning the empty string to `old_str`

kpops/utils/yaml_loading.py
sujuka99: removed

tests/components/test_pipeline_component.py
sujuka99: done

kpops/utils/yaml_loading.py
disrupted: I don't think the scenario of it being unbound is possible. We should either silence the diagnostic or init it with an empty string as before.

sujuka99: I will init it like before.

tests/pipeline/test_pipeline.py
sujuka99: It shouldn't be needed currently, but I would, indeed, prefer to keep it in as a fallback. A silent infinite loop is quite the undesirable 'feature'. 😁

I lowered the time significantly, lmk if you think I should adjust it further

kpops/utils/yaml_loading.py
sujuka99: done

tests/pipeline/test_pipeline.py
```diff
@@ -371,3 +443,22 @@ def test_model_serialization(self, snapshot: SnapshotTest):
 
         enriched_pipeline = yaml.safe_load(result.stdout)
         snapshot.assert_match(enriched_pipeline, "test-pipeline")
+
+    def test_kubernetes_app_name_validation(self):
+        with pytest.raises((ValueError, ParsingException)):
```
disrupted, line 448: which of those exceptions are raised?

sujuka99: both

kpops/components/base_components/kubernetes_app.py
```diff
@@ -46,6 +45,9 @@ class KubernetesApp(PipelineComponent):
     :param schema_type: Used for schema generation, same as :param:`type`,
         defaults to "kubernetes-app"
     :type schema_type: Literal["kubernetes-app"], optional
+    :param validate_name: Whether to check if the name of the component is
+        compatible with Kubernetes, defaults to True
+    :type validate_name: bool, optional
```
disrupted, line 50: why is it optional? it even has a default value

kpops/utils/yaml_loading.py
```diff
@@ -37,3 +37,45 @@ def substitute(input: str, substitution: Mapping[str, Any] | None = None) -> str
...
+    """
+    if not kwargs:
+        return input
+    old_str, new_str = "", substitute(input, kwargs)
+    steps = set()
+    while new_str not in steps:
+        steps.add(new_str)
+        old_str, new_str = new_str, substitute(new_str, kwargs)
+    if new_str != old_str:
+        raise Exception(
```
disrupted, line 78: we should avoid raising generic `Exception`, it's an anti-pattern. how about `ValueError`, i.e. input is invalid?

tests/pipeline/test_pipeline.py
```diff
@@ -118,12 +120,82 @@ def test_substitute_component_names(self, snapshot: SnapshotTest):
...
+                "test_placeholder_in_placeholder"
+            ]
+            == "filter-app-filter"
+        )
 
         snapshot.assert_match(enriched_pipeline, "test-pipeline")
 
+    @pytest.mark.timeout(0.5)
+    def test_substitute_in_component_infinite_loop(self):
+        with pytest.raises(Exception):
```
disrupted, line 157: change Exception type to something more precise, see comment above

kpops/components/base_components/kubernetes_app.py
sujuka99: optional in the ReST means that you don't have to give it a value, not the same as `Optional[]`.

[reference](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format)

disrupted: ah ok, thanks for explaining. I find this very.. _unhelpful_ :D like I would expect the `type` attribute to only contain information about the allowed type and therefore understood optional as `... | None` 🤦🏻 
I guess most important is that we find a clear way that works for us

sujuka99: Yeah, I agree that it is kinda counterintuitive. I am okay with abolishing the `optional` keyword in `:type ...:`, it is redundant when `:param ...:` already contains `, defaults to ...`

kpops/pipeline_generator/pipeline.py
raminqaf, line 287: I need help to follow this TODO. What are these breaking changes? 👀

kpops/components/base_components/kubernetes_app.py
```diff
@@ -88,8 +96,9 @@ class Config(CamelCaseConfig, DescConfig):
         pass
 
     def __init__(self, **kwargs):
+        if kwargs.get("validate_name", True):
+            self.validate_kubernetes_name(kwargs["name"])
```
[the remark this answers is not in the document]
sujuka99: A component is initialized more than once before reaching its "final" form.
[first](https://github.com/bakdata/kpops/pull/198/files#diff-bf5691f6b8f231bf35e494f9dbb7d78f8bff5467239c2a3f9a93af4a78e68bd1R185-R190)
[second](https://github.com/bakdata/kpops/pull/198/files#diff-bf5691f6b8f231bf35e494f9dbb7d78f8bff5467239c2a3f9a93af4a78e68bd1R243-R244)

Since substitution happens after at least one init and right before the last one, it is possible that the `name` of a `KubernetesApp` instance contains placeholders. Hence, we can only validate the name during the final init after the substitution is carried out.

kpops/components/base_components/kubernetes_app.py
```diff
@@ -192,12 +201,16 @@ def get_helm_chart(self) -> str:
             f"Please implement the get_helm_chart() method of the {self.__module__} module."
         )
 
-    def __check_compatible_name(self) -> None:
-        """Check if the component's name `self.name` is valid for Kubernetes"""
-        if not bool(KUBERNETES_NAME_CHECK_PATTERN.match(self.name)):  # TODO: SMARTER
-            raise ValueError(
-                f"The component name {self.name} is invalid for Kubernetes."
-            )
+    # TODO: SMARTER
```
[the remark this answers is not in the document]
sujuka99: We do not, [the comment was there before](https://github.com/bakdata/kpops/blob/main/kpops/components/base_components/kubernetes_app.py#L195-L200). I probably shouldn't have moved it from its original position 🤔 

I am not sure what it referred to, but assumed that it's asking for a way to validate the name without a regexp. I explored this a bit and couldn't come up with a faster or a more elegant solution, so I left it there for someone smarter :D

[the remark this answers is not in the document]
sujuka99: I tried to see when it was added and by whom, but it is in the initial commit of the old gh kpops repo :D

I assume there was a gitlab repo before that, soo I guess there could be a hint, but I have no access to it.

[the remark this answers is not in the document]
raminqaf: This was added initially by Victor. This is exactly why we shouldn't keep comments like that. It is not helping anyone now. I would remove it.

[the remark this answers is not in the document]
sujuka99: Agreed, will remove it

kpops/components/base_components/kubernetes_app.py
```diff
@@ -192,12 +194,15 @@ def get_helm_chart(self) -> str:
...
-        if not bool(KUBERNETES_NAME_CHECK_PATTERN.match(self.name)):  # TODO: SMARTER
-            raise ValueError(
-                f"The component name {self.name} is invalid for Kubernetes."
-            )
+    @staticmethod
+    def validate_kubernetes_name(name: str) -> None:
+        """Check if a name is valid for a Kubernetes resource
+
+        :param name: Name that is to be used for the resource
+        :type name: str
```
disrupted, line 202:
```suggestion
        :type name: str
        :raises ValueError: ...
```

disrupted (changes requested):

disrupted (approved):
LGTM

raminqaf (approved):

PR #200, opened 2023-04-26
Add `kpops --version` command

### LINKED ISSUE #184: Print version of installed KPOps with `KPOps --version`
We should add this option to print the installed KPOps version. Currently, the user has to use `pip list` to see the installed `KPOps` version

szymonsadowski3:
Closes #184

The ideas were taken from:

- _version.py file -> https://stackoverflow.com/questions/458550/standard-way-to-embed-version-into-python-package
- Handling the version command -> https://github.com/tiangolo/typer/issues/52

Should be integrated in conjunction with the PR that changes bump action to bump the version in the new _version.py file (will paste it below in a moment)

[edit] the related ci-templates pull request can be found here https://github.com/bakdata/ci-templates/pull/95

### DIFF
--- a/kpops/cli/main.py
+++ b/kpops/cli/main.py
@@ -355,7 +355,12 @@ def version_callback(show_version: bool) -> None:
 @app.callback()
 def main(
     version: bool = typer.Option(
-        False, "--version", "-V", callback=version_callback, is_eager=True
+        False,
+        "--version",
+        "-V",
+        help="Print KPOps version",
+        callback=version_callback,
+        is_eager=True,
     ),
 ):
     return
--- a/kpops/cli/main.py
+++ b/kpops/cli/main.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING, Iterator, Optional
 
 import typer
 
+from kpops import __version__
 from kpops.cli.custom_formatter import CustomFormatter
 from kpops.cli.pipeline_config import ENV_PREFIX, PipelineConfig
 from kpops.cli.registry import Registry
@@ -343,6 +344,21 @@ def clean(
         log_action("Clean", component)
         component.destroy(dry_run)
         component.clean(dry_run)
+
+
+def version_callback(show_version: bool) -> None:
+    if show_version:
+        typer.echo(f"KPOps {__version__}")
+        raise typer.Exit()
+
+
+@app.callback()
+def main(
+    version: bool = typer.Option(
+        False, "--version", "-V", callback=version_callback, is_eager=True
+    ),
+):
+    return
 
 
 if __name__ == "__main__":
--- a/kpops/cli/main.py
+++ b/kpops/cli/main.py
@@ -363,7 +363,7 @@ def main(
         is_eager=True,
     ),
 ):
-    return
+    ...
 
 
 if __name__ == "__main__":
--- a/docs/docs/user/references/cli-commands.md
+++ b/docs/docs/user/references/cli-commands.md
@@ -8,6 +8,7 @@ $ kpops [OPTIONS] COMMAND [ARGS]...
 
 **Options**:
 
+* `-V, --version`: Print KPOps version
 * `--install-completion`: Install completion for the current shell.
 * `--show-completion`: Show completion for the current shell, to copy it or customize the installation.
 * `--help`: Show this message and exit.
--- a/kpops/__init__.py
+++ b/kpops/__init__.py
@@ -1,0 +1,3 @@
+import importlib.metadata
+
+__version__ = importlib.metadata.version("kpops")
--- a/kpops/__init__.py
+++ b/kpops/__init__.py
@@ -1,3 +1,1 @@
-import importlib.metadata
-
-__version__ = importlib.metadata.version("kpops")
+__version__ = "1.1.1"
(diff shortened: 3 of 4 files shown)


raminqaf:
@szymonsadowski3 Thanks for your contribution! It seems that your formatting is not correct, and it breaks the CI pipeline. Could you please install the pre-commit hook for the formatting?
https://github.com/bakdata/kpops/blob/main/.pre-commit-config.yaml

kpops/cli/main.py
```diff
@@ -345,5 +353,13 @@ def clean(
         component.clean(dry_run)
 
 
+@app.callback()
+def main(
+        version: bool = typer.Option(None, "--version", callback=version_callback, is_eager=True),
+):
+    log.debug(f"Version option called: {version}")
```
raminqaf, line 360: I don't think this is necessary. @disrupted WDYT?

kpops/_version.py
```diff
@@ -0,0 +1 @@
+__version__ = "1.1.1"
```
raminqaf, line 1: @disrupted We have to keep in mind that this file needs to be updated whenever we do a release.

kpops/cli/main.py
```diff
@@ -91,6 +93,12 @@
 log = logging.getLogger("")
 
 
+def version_callback(value: bool):
+    if value:
+        typer.echo(f"Kpops CLI Version: {__version__}")
```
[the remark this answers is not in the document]
szymonsadowski3: applied

kpops/_version.py
disrupted: yeah, I would prefer updating it with a Poetry plugin during version bumping, rather than doing it in the CI release pipeline.

szymonsadowski3: Not sure what release process is followed, but the idea was to let python-poetry-bump-version action handle that (ref: https://github.com/bakdata/ci-templates/pull/95)

disrupted: If we use either the [poetry-bumpversion](https://github.com/monim67/poetry-bumpversion) plugin, or read it dynamically using `importlib` we can entirely skip the additional CI step.

```python
__version__= importlib.metadata.version("kpops")
```

raminqaf:
> > @szymonsadowski3 Thanks for your contribution! It seems that your formatting is not correct, and it breaks the CI pipeline. Could you please install the pre-commit hook for the formatting? https://github.com/bakdata/kpops/blob/main/.pre-commit-config.yaml
> 
> I just ran the pre-commit hooks and commited the changes accordingly :)

Thanks! I created an issue for that. We need a guideline on how externals should contribute to KPOps. https://github.com/bakdata/kpops/issues/201

kpops/_version.py
disrupted: also please move this to `kpops/__init__.py` and use an absolute import `from kpops import __version__`

kpops/cli/main.py
```diff
@@ -91,6 +93,12 @@
 log = logging.getLogger("")
 
 
+def version_callback(value: bool):
+    if value:
```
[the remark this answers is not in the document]
disrupted: please also move this function down above the `main` function from which it is called

kpops/cli/main.py
```diff
@@ -345,5 +353,15 @@ def clean(
         component.clean(dry_run)
 
 
+@app.callback()
+def main(
+    version: bool = typer.Option(
+        None, "--version", callback=version_callback, is_eager=True
```
disrupted, line 359:
```suggestion
        False, "--version", "-V", callback=version_callback, is_eager=True
```

kpops/cli/main.py
disrupted: yes, the debug logging is unnecessary

disrupted (changes requested):

kpops/cli/main.py
szymonsadowski3: removed the unnecessary logging statement

kpops/_version.py
szymonsadowski3: > also please move this to `kpops/__init__.py` and use an absolute import `from kpops import __version__`

great suggestions, applied

kpops/cli/main.py
```diff
@@ -91,6 +93,12 @@
 log = logging.getLogger("")
 
 
+def version_callback(value: bool):
+    if value:
+        typer.echo(f"KPOps CLI Version: {__version__}")
```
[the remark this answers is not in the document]
szymonsadowski3: applied

kpops/cli/main.py
szymonsadowski3: applied

kpops/cli/main.py
```diff
@@ -345,5 +346,20 @@ def clean(
         component.clean(dry_run)
 
 
+def version_callback(show_version: bool) -> None:
+    if show_version:
+        typer.echo(f"KPOps Version: {__version__}")
```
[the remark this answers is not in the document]
szymonsadowski3: applied

disrupted:
during testing I've noticed that in local development environments reading the version dynamically from the package using `importlib` doesn't work, i.e. when I would run
```
kpops --version
poetry version patch
kpops --version
```
it would both times print the same version.

I've switched to the poetry bumpversion plugin and not it works as expected.

disrupted (approved):

raminqaf (approved):

PR #205, opened 2023-04-26
Test shell scripts on zsh in the CI

sujuka99:
Avoid situations like [this](https://github.com/bakdata/kpops/pull/197#issuecomment-1523604103)

### DIFF
--- a/.github/actions/pre-commit-zsh/action.yml
+++ b/.github/actions/pre-commit-zsh/action.yml
@@ -1,5 +1,5 @@
 name: "Test the pre-commit shell scripts in zsh"
-description: "Only works when ran on macos"
+description: ""
 
 runs:
   using: "composite"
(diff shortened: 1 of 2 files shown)


.github/actions/pre-commit-zsh/action.yml
```diff
@@ -0,0 +1,21 @@
...
+
+    - name: Install dependencies
+      shell: zsh {0}
+      run: poetry install --no-interaction
+
+    - name: Run pre-commit-hooks
+      shell: zsh {0}
+      run: |
+        poetry run pre-commit run mypy --all-files --show-diff-on-failure
+        poetry run pre-commit run gen-docs --all-files --show-diff-on-failure
```
sujuka99, line 21: I could also do it like that to avoid needing changes after each new added hook, but it would add redundancy.

```suggestion
        poetry run pre-commit run --all-files --show-diff-on-failure
```

.github/actions/pre-commit-zsh/action.yml
```diff
@@ -0,0 +1,21 @@
+name: "Test the pre-commit shell scripts in zsh"
+description: "Only works when ran on macos"
```
disrupted, line 2: isn't the whole idea that it should be compatible cross-platform?

sujuka99: Unfortunately, no. Perhaps we could install zsh in the ubuntu tests as a step?

sujuka99: zsh comes preinstalled only on mac, not on the ubuntu runner, hence if you run the action on ubuntu it will return an error

sujuka99: https://github.com/actions/runner-images/issues/264

PR #207, opened 2023-04-28
Rewrite bash pre-commit hooks in Python

sujuka99:
- [x] gen_schema
- [x] gen_docs
- [x] run_mypy


sujuka99:
TODO: delete zsh-compatibility branch upon merging this one

hooks/run_mypy.py
disrupted, line 1: this MyPy wrapper is part of our Python template. Should we leave it as-is or replace the template as well? is the change of converting it to Python necessary?

.pre-commit-config.yaml
```diff
@@ -33,13 +25,19 @@ repos:
     hooks:
       - id: gen-schema
         name: gen-schema
-        entry: ./hooks/gen_schema.sh
+        entry: python hooks/gen_schema.py
         language: system
+        types: [python]
         require_serial: true
```
disrupted, line 31: consider adding `pass_filenames: false`

.pre-commit-config.yaml
```diff
@@ -33,13 +25,19 @@ repos:
     hooks:
       - id: gen-schema
         name: gen-schema
-        entry: ./hooks/gen_schema.sh
+        entry: python hooks/gen_schema.py
         language: system
+        types: [python]
```
disrupted, line 30: good improvement! 👍🏻

.pre-commit-config.yaml
```diff
@@ -33,13 +25,19 @@ repos:
...
-  - repo: local
-    hooks:
       - id: gen-docs
         name: gen-docs
-        entry: ./hooks/gen_docs.sh
+        entry: python hooks/gen_docs.py
         language: system
+        types: [python]
         require_serial: true
+      - id: mypy
```
disrupted, line 38: any reason for changing the order of execution?

hooks/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+from contextlib import contextmanager
+from pathlib import Path
+
+PATH_ROOT = Path(__file__).parents[1]
+
+# Taken from https://stackoverflow.com/a/24176022/11610149
+
+
+@contextmanager
+def cd(newdir: str | Path = PATH_ROOT):
```
disrupted, line 12: clever use of a contextmanager :) why not stick to pathlib, is the `os` module really necessary?

hooks/gen_docs.py
```diff
@@ -0,0 +1,22 @@
...
+from hooks import PATH_ROOT
+
+PATH_KPOPS_MAIN = PATH_ROOT / "kpops/cli/main.py"
+PATH_CLI_COMMANDS_DOC = PATH_ROOT / "docs/docs/user/references/cli-commands.md"
+
+# Run typer-cli on kpops to generate doc on CLI usage
+# TODO: try to use typer_cli.main.docs here instead
+typer_args = (
+    f"typer {PATH_KPOPS_MAIN} utils docs --name kpops --output {PATH_CLI_COMMANDS_DOC}"
+)
```
disrupted, line 13: you could convert this to a `list[str]`

hooks/gen_docs.py
```diff
@@ -0,0 +1,22 @@
...
+subprocess.run(typer_args, shell=True)
+
+# Replace wrong title in CLI Usage doc
+text: list[str] = []
```
disrupted, line 17: is it necessary to instantiate an empty list here before actually reading the file into it?

hooks/gen_schema.py
```diff
@@ -0,0 +1,23 @@
+"""Generates the stock KPOps editor integration schemas"""
+from contextlib import redirect_stdout
+from io import StringIO
+
+from hooks import PATH_ROOT
+from kpops.utils.gen_schema import gen_config_schema, gen_pipeline_schema
+
+PATH_TO_SCHEMA = PATH_ROOT / "docs/docs/schema"
+
+schema = ""
```
disrupted, line 10: necessary?

hooks/gen_schema.py
```diff
@@ -0,0 +1,23 @@
...
+with redirect_stdout(StringIO()) as f:
+    gen_pipeline_schema()
+    schema = f.getvalue()
+
+with open(PATH_TO_SCHEMA / "pipeline.json", "w") as file:
+    file.write(schema)
```
disrupted, line 16: with pathlib you could consider just doing `path.write_text("...")`

hooks/gen_schema.py
```diff
@@ -0,0 +1,23 @@
...
+
+with redirect_stdout(StringIO()) as f:
+    gen_config_schema()
+    schema = f.getvalue()
+
+with open(PATH_TO_SCHEMA / "config.json", "w") as file:
+    file.write(schema)
```
disrupted, line 23: perhaps extract this to a function? the two are very similar

hooks/run_mypy.py
```diff
@@ -0,0 +1,22 @@
...
+from sys import stderr, stdout
+
+from mypy import api
+
+from hooks import cd
+
+# Args should contain the args that the equivalent
+# call to mypy from the shell would contain
+args = ["--pretty", "kpops", "tests", "hooks"]
+result: tuple = tuple()
```
disrupted, line 14: necessary?

disrupted (changes requested):

hooks/run_mypy.py
sujuka99: Since we talked bout rewriting all hooks, I just went for it. Aside from consistency language-wise in the hooks, I don't see any pros or cons of converting it to Python. I will revert it if you think it's unnecessary.

If we replace the template, it would have to also include the `__init__` or at least the `cd()` function from it. I am not sure if that's desirable.

.pre-commit-config.yaml
sujuka99: Not really, it wasn't intentional, will change it back

hooks/gen_docs.py
sujuka99: No, I had assumed (wrongly) that the context manager creates a scope.

hooks/gen_schema.py
sujuka99: no

hooks/gen_schema.py
sujuka99: very cool, will do

hooks/run_mypy.py
disrupted: hm, I see. I don't really have a preference. what do you think makes more sense?

hooks/run_mypy.py
sujuka99: nop

hooks/__init__.py
sujuka99: I couldn't find how to change dir without `os`

hooks/gen_schema.py
sujuka99: done

hooks/run_mypy.py
sujuka99: I would personally prefer the Python script, although I don't have a good explanation why. I guess it's a bit more platform-independent? Not sure about that though. I like having all scripts in one language. Since it's a **Python** template, I suppose it suits it to have the hooks written in Python, too.

On the other hand, the bash script has fewer lines, is probably easier to understand and it's just one file, so I see why it may be better to stick with it.

disrupted (changes requested):

.pre-commit-config.yaml
```diff
@@ -31,15 +23,22 @@ repos:
         exclude: ^tests/.*snapshots/
   - repo: local
     hooks:
+      - id: mypy
```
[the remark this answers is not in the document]
sujuka99: sry, thought I did in the feedback commit, must be in the right place now.

disrupted (approved):

PR #209, opened 2023-05-03
Collapse pip install output for GitHub action

disrupted:
usually we only want to see the output of `kpops <operation>` in the CI. This effectively hides the output of `pip install` by grouping the log https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-grouping-log-lines


sujuka99 (commented):
Could you also update the README with a newer version of `kpops`?
Currently the example version is `0.3.0`

yannick-roeder (approved):

yannick-roeder (approved):

PR #211, opened 2023-05-04
Set up Helm inside composite action

disrupted:
- Set up Helm for composite action
- Install latest version of kpops by default

### DIFF
--- a/actions/kpops-runner/action.yaml
+++ b/actions/kpops-runner/action.yaml
@@ -26,10 +26,19 @@ inputs:
   kpops-version:
     description: "KPOps version"
     required: true
+  helm-version:
+    description: "Helm version"
+    required: false
+    default: "latest"
 
 runs:
   using: "composite"
   steps:
+    - name: Set up Helm
+      uses: azure/setup-helm@v3
+      with:
+        version: ${{ inputs.helm-version }}
+
     - name: Install KPOps
       shell: bash
       run: |
--- a/actions/kpops-runner/action.yaml
+++ b/actions/kpops-runner/action.yaml
@@ -25,7 +25,8 @@ inputs:
     required: false
   kpops-version:
     description: "KPOps version"
-    required: true
+    required: false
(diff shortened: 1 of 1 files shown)


actions/kpops-runner/action.yaml
```diff
@@ -25,16 +25,30 @@ inputs:
...
   kpops-version:
     description: "KPOps version"
-    required: true
+    required: false
+    default: "latest"
+  helm-version:
+    description: "Helm version"
+    required: false
+    default: "latest"
+  token:
```
philipp94831, line 34: Is there a README for the action (I guess not)? We should create one. This PR or issue?

sujuka99: There is, I was just going to ask for an update to it

disrupted: oh, there is, I just forgot to update it

philipp94831: I am actually surprised 😂

actions/kpops-runner/action.yaml
```diff
@@ -25,16 +25,30 @@ inputs:
...
+    required: false
 
 runs:
   using: "composite"
   steps:
+    - name: Set up Helm
+      uses: azure/setup-helm@v3
+      with:
+        version: ${{ inputs.helm-version }}
+        token: ${{ inputs.token }}
```
philipp94831, line 45: Do you know why a token is needed? Seems quite unintuitive

disrupted: idk, it's written in their readme
actually I had a look in their code just now and it seems the actual latest version is hardcoded https://github.com/Azure/setup-helm/blob/ac5ee1fca8fba06d82bfc39af726f19ffe037a49/src/run.ts#L16

philipp94831 (approved):

sujuka99 (approved):
LGTM

PR #214, opened 2023-05-09
Replace native os exit with exceptions

irux:


kpops/component_handlers/topic/handler.py
```diff
@@ -150,7 +150,7 @@ def __check_partition_count(
             log.error(
                 f"Topic Creation: partition count of topic {topic_name} changed! Partitions count of topic {topic_name} is {partition_count}. The given partitions count {topic_spec.partitions_count}."
             )
-            exit(1)  # FIXME raise instead https://github.com/bakdata/kpops/issues/101
+            sys.exit(1)  # FIXME raise instead https://github.com/bakdata/kpops/issues/101
```
raminqaf, line 153: How about removing the sys.exit(1) and raising an exception instead? As described in the comment/issue

raminqaf (changes requested):

kpops/component_handlers/topic/handler.py
raminqaf: Replacing these exit(1)s with exceptions is a better solution, and it is not dependent on the OS.

kpops/component_handlers/kafka_connect/kafka_connect_handler.py
```diff
@@ -120,9 +120,7 @@ def __dry_run_connector_creation(
                 f"Connector Creation: validating the connector config for connector {connector_name} resulted in the following errors:"
             )
             log.error("\n".join(errors))
-            sys.exit(
-                1
-            )  # FIXME raise instead https://github.com/bakdata/kpops/issues/101
+            raise ValueError(f"Error creating connector {connector_name}")
```
raminqaf, line 123: Don't log and throw
https://stackoverflow.com/questions/6639963/why-is-log-and-throw-considered-an-anti-pattern

kpops/component_handlers/topic/handler.py
```diff
@@ -150,7 +149,7 @@ def __check_partition_count(
             log.error(
                 f"Topic Creation: partition count of topic {topic_name} changed! Partitions count of topic {topic_name} is {partition_count}. The given partitions count {topic_spec.partitions_count}."
             )
-            sys.exit(1)  # FIXME raise instead https://github.com/bakdata/kpops/issues/101
+            raise ValueError(f"Error creating the topic {topic_name}")
```
raminqaf, line 152: Use the message of the log in the Error and remove the log

kpops/component_handlers/topic/handler.py
```diff
@@ -171,7 +170,7 @@ def __check_replication_factor(
             log.error(
                 f"Topic Creation: replication factor of topic {topic_name} changed! Replication factor of topic {topic_name} is {replication_factor}. The given replication count {topic_spec.replication_factor}."
             )
-            sys.exit(1)  # FIXME raise instead https://github.com/bakdata/kpops/issues/101
+            raise ValueError(f"Error creating the topic {topic_name}")
```
raminqaf, line 173: same

raminqaf (changes requested):

kpops/component_handlers/topic/handler.py
disrupted: additionally we should consider introducing custom Exception types

raminqaf: @disrupted Agree! We should create a Issue for that to track it!

irux: maybe now is better (?)

disrupted: perhaps you could resolve #215 in this PR if it's not too much effort?

irux: You mean all the exceptions? But let me do it in a separé PR I think it makes more sense. This will merged on the windows part

disrupted (approved):

raminqaf:
Please update the PRs title!

raminqaf (approved):

Issue #218, opened 2023-05-22
Example pipeline should be copied to a stable branch in another repo


sujuka99:
**Problem**
Currently, the docs depend on examples in the main branch of [kpops](https://github.com/bakdata/kpops).

**Task**
Move or copy the example to the [kpops-examples](https://github.com/bakdata/kpops-examples) repo. 

A possible solution could be to copy from [main](https://github.com/bakdata/kpops/tree/main) to the examples repo in the [release](https://github.com/bakdata/kpops/blob/main/.github/workflows/release.yaml) workflow?

PR #216, opened 2023-05-12
Update example pipeline

sujuka99:

PR #219, opened 2023-05-22
Fix links to ATM-fraud defaults

### LINKED ISSUE #220: Test docs generation with mkdocs
To avoid problems with publishing such as [this one](https://github.com/bakdata/kpops/actions/runs/5044422510/jobs/9047562651), we should test docs generation with mkdocs as part of the CI or a pre-commit hook.

Could be as simple as a line in the CI `just serve-docs`.

sujuka99:
Closes #220 
#216 updated the example pipeline file structure, but this was not reflected in the links to the docs.

See #218

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -59,7 +59,7 @@ jobs:
         run: poetry install --with docs
 
       - name: Test docs build (mkdocs)
-        run: poetry run mkdocs build
+        run: poetry run mkdocs build -f docs/mkdocs.yml
 
   publish:
     name: Publish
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -55,6 +55,12 @@ jobs:
       - name: Test
         run: poetry run pytest tests
 
+      - name: Install docs dependencies
+        run: poetry install --with docs
```
disrupted, line 59: any reason for not adding this to the initial poetry install above?

sujuka99: to save time if anything fails fefore that

disrupted: ah, yeah that makes sense. I don't have a preference, you can decide yourself what you think is most appropriate

sujuka99: I prefer to keep it like it is currently, it seems that it is only installing the doc dependencies after all

disrupted (approved):

PR #221, opened 2023-05-23
Fix environment variables on windows

irux:


kpops/components/base_components/base_defaults_component.py
```diff
@@ -158,7 +158,7 @@ def defaults_from_yaml(path: Path, key: str) -> dict:
 
     kafka_app_defaults = defaults_from_yaml(Path("/path/to/defaults.yaml"), "kafka-app")
     """
-    content = load_yaml_file(path, substitution=dict(os.environ))
+    content = load_yaml_file(path, substitution=dict(environ))
```
disrupted, line 161: is casting to dict still necessary?

tests/components/test_pipeline_component.py
```diff
@@ -55,9 +54,9 @@ def test_topic_substitution(self):
         assert "output-plain-pipeline-component" in pipeline_component.to.topics
         assert len(pipeline_component.to.topics.keys()) == 2
 
-    def test_prefix_substitution(self, monkeypatch: MonkeyPatch):
+    def test_prefix_substitution(self):
         prefix = "my-fake-prefix"
-        monkeypatch.setenv("pipeline_name", prefix)
+        environ["pipeline_name"] = prefix
```
disrupted, line 59: does this need to be cleared on teardown of the test?

tests/pipeline/snapshots/snap_test_pipeline.py
```diff
@@ -563,7 +563,7 @@
             "prefix": "resources-pipeline-with-paths-",
             "repoConfig": {
                 "repoAuthFlags": {
-                    "caFile": "/my-cert.cert",
+                    "caFile": str(Path("/my-cert.cert")),
```
disrupted, line 566: did you manually edit the snapshot? it should only be generated

disrupted (changes requested):

kpops/utils/environment.py
```diff
@@ -0,0 +1,33 @@
+import os
+import platform
+from typing import Callable
+
+
+class Environment(dict):
```
disrupted, line 6: perhaps we should subclass from `collections.UserDict`

kpops/utils/environment.py
```diff
@@ -0,0 +1,33 @@
...
+            mapping = {transformation(key): value for key, value in mapping.items()}
+        else:
+            mapping = {}
+        if kwargs:
+            mapping.update(
+                {transformation(key): value for key, value in kwargs.items()}
+            )
+        super().__init__(mapping)
+
+    def _key_camel_case_transform(self, key):
```
disrupted, line 20: please add missing type annotations

tests/pipeline/snapshots/snap_test_pipeline.py
irux: I updated it manually, we don't have any other way to do that for the snapshots. At least one I am aware of. Any ideas ?

kpops/components/base_components/base_defaults_component.py
irux: I would prefer to this because it creates a copy

kpops/utils/environment.py
```diff
@@ -0,0 +1,34 @@
...
+            mapping = {transformation(key): value for key, value in mapping.items()}
+        else:
+            mapping = {}
+        if kwargs:
+            mapping.update(
+                {transformation(key): value for key, value in kwargs.items()}
+            )
+        super().__init__(mapping)
+
+    def _key_camel_case_transform(self, key: str) -> str:
```
disrupted, line 21: mark as `@staticmethod`

tests/pipeline/snapshots/snap_test_pipeline.py
disrupted: how about changing the path to a relative one in tests/pipeline/resources/pipeline-with-paths/pipeline.yaml, would that work?

kpops/components/base_components/base_defaults_component.py
disrupted: but is it necessary to create a copy in memory?

kpops/utils/environment.py
raminqaf, line 21: Are these protected methods or private methods?
```
Protected methods use one underscore as their prefix, 
while private methods use two underscores as their prefix
```

kpops/utils/environment.py
```diff
@@ -0,0 +1,34 @@
...
+        return str(key).lower()
+
+    def _key_identity_transform(self, key: str) -> str:
+        return str(key)
```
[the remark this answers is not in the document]
irux: Well it depends... right now could be redundant but a dict key can be anything. So, we use it for os.environ and that's why we can say for sure it is a string, but it can multiple things. I can change it if you want.

kpops/components/base_components/base_defaults_component.py
irux: I mean, I understood the code as doing it because of the same thing. os.environ is already a dict too, so I really thought that it was because of this reason. I will remove it then

kpops/utils/environment.py
```diff
@@ -0,0 +1,36 @@
...
+
+    @staticmethod
+    def __key_identity_transform(key: str) -> str:
+        return str(key)
+
+    def __get_transformation(self) -> Callable[[str], str]:
+        if platform.system() == "Windows":
+            return Environment.__key_camel_case_transform
+        else:
+            return Environment.__key_identity_transform
```
raminqaf, line 33: I am confused about this syntax here 😄 Where are you passing the `key` parameter?

kpops/utils/environment.py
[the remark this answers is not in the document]
disrupted: imo it doesn't have to be general purpose, as we only need it for environment. Also the parameter type annotation says it's always a `str`

kpops/utils/environment.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+environ = Environment(os.environ)
```
disrupted, line 36: what do you think about `ENV`? I was never a fan of the environ name in the os module 😄

tests/components/test_pipeline_component.py
```diff
@@ -55,9 +64,7 @@ def test_topic_substitution(self):
         assert "output-plain-pipeline-component" in pipeline_component.to.topics
         assert len(pipeline_component.to.topics.keys()) == 2
 
-    def test_prefix_substitution(self, monkeypatch: MonkeyPatch):
-        prefix = "my-fake-prefix"
-        monkeypatch.setenv("pipeline_name", prefix)
+    def test_prefix_substitution(self, prefix_to_match):
```
disrupted, line 67:
```suggestion
    def test_prefix_substitution(self, prefix_to_match: str):
```

tests/utils/test_environment.py
```diff
@@ -0,0 +1,117 @@
...
+
+    assert environment["my"] == "fake"
+    assert environment["environment"] == "here"
+
+
+@patch("platform.system")
+def test_normal_behaviour_keys_transformation(system, fake_environment_linux):
+    system.return_value = "Linux"
+    environment = Environment(fake_environment_linux)
+    keys = list(environment.keys())
```
disrupted, line 40: `set` instead of `list`?

kpops/utils/environment.py
disrupted: this is just the transformer function (see method return type annotation)

irux: > I am confused about this syntax here 😄 Where are you passing the `key` parameter?

It is a factory function

sujuka99:
> (note to self about failing CI): we should exclude the pytest snapshots from the pre-commit checks

#222

disrupted (approved):

irux:
> please change the `substitution` parameter in `load_yaml_file` (and `generate_hashkey`) functions from `dict | None = None` to `Mapping[str, Any] | None = None`. same as `substitute` function. Otherwise `UserDict` (i.e. `Environment` instances) is incompatible

weird... shouldn't have mypy give a warning on this ?

raminqaf (approved):
LGTM! 🚀

disrupted:
> > please change the `substitution` parameter in `load_yaml_file` (and `generate_hashkey`) functions from `dict | None = None` to `Mapping[str, Any] | None = None`. same as `substitute` function. Otherwise `UserDict` (i.e. `Environment` instances) is incompatible
> 
> 
> 
> weird... shouldn't have mypy give a warning on this ? 

yeah, one would imagine it should. Pyright does detect it for me locally. Do you have a clue why MyPy misses it?

PR #225, opened 2023-05-25
Enable cross-component substitution

### LINKED ISSUE #224: Enable cross-component substitution
Make it possible to use variables in the pipeline def that refer to other component's attributes.

Example: `type: ${otherComponent_type}` where 'otherComponent' is the name of the other component

sujuka99:
Closes #224


sujuka99:
Functionality is not needed

PR #226, opened 2023-05-26
Exclude pytest snapshots from pre-commit hook

### LINKED ISSUE #222: Exclude pytest snapshots from pre-commit hooks
Currently the pre-commit hooks also cover the pytest snapshots. This is inconvenient as every time a snapshot is updated/generated it fails the checks.

sujuka99:
closes #222

### DIFF
--- a/hooks/gen_docs.py
+++ b/hooks/gen_docs.py
@@ -18,7 +18,7 @@ typer_args: list[str] = [
     "--output",
     str(PATH_CLI_COMMANDS_DOC),
 ]
-subprocess.run(typer_args, shell=True)
+subprocess.run(" ".join(typer_args), shell=True)
 
 # Replace wrong title in CLI Usage doc
 with open(PATH_CLI_COMMANDS_DOC, "r") as f:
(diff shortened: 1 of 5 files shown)


hooks/gen_docs.py
```diff
@@ -18,7 +18,7 @@
     "--output",
     str(PATH_CLI_COMMANDS_DOC),
 ]
-subprocess.run(typer_args, shell=True)
+subprocess.run(" ".join(typer_args), shell=True)
```
sujuka99, line 21: The docs weren't being generated, it's fixed now.

disrupted: this change shouldn't be necessary. `list[str]` is totally fine for `args`

disrupted: ah, it's probably because you're setting `shell=True`. is this necessary?

sujuka99: Apparently not :D

fixed.

disrupted (approved):

Issue #194, opened 2023-04-14
Update streams-bootstrap models to match v2.12.0


torbsto:
streams-bootstrap v2.11.0 includes a couple of additions to the Helm chart that should be part of KPOps' streams-bootstrap models. 
Relevant changes:
- https://github.com/bakdata/streams-bootstrap/commit/737f34d9ef0a5821a44f726bd7ce2d3aa6037fb9
- https://github.com/bakdata/streams-bootstrap/commit/6d5a489852f75ed36ec564c4b2b374005103542d
- https://github.com/bakdata/streams-bootstrap/commit/e424f0bef95051d52f5bf539fd626ec780cd94b0

sujuka99:
Edit: Updated title as streams-bootstrap v2.12.0 was released

disrupted:
also `optimizeLeaveGroupBehavior` is currently missing

disrupted:
streams-bootstrap v2 models are deprecated. If anything needs to be added to the models it should be done for v3

PR #230, opened 2023-06-05
Fix missing extra input topics

disrupted:
Replace Pydantic's `exclude_unset` option with `exclude_defaults`, or remove entirely in most places, as it was likely used incorrectly


disrupted:
> pls add this to the atm fraud defaults under `streams-app`
> image

this option should also be added, I've commented in https://github.com/bakdata/kpops/issues/194 as these seem related

sujuka99 (approved):
LGTM

philipp94831 (approved):

PR #232, opened 2023-06-07
Refactor input/output types

### LINKED ISSUE #119: Refactor input/output types

sujuka99:
closes #119


docs/docs/schema/pipeline.json
```diff
@@ -31,7 +31,7 @@
             "description": "Input topic",
             "properties": {
                 "role": {
-                    "description": "Custom identifier belonging to a topic, provide only if `type` is `extra` or `extra-pattern`",
+                    "description": "Custom identifier belonging to a topic, provide only if `type` is `extra`, `pattern` or `extra-pattern`",
```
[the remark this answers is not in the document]
philipp94831: There is no need for type extra and extra-pattern anymore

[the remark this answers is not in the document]
sujuka99: I updated the way it works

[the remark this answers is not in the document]
sujuka99: i would prefer to keep the old functionality at least until v2 is released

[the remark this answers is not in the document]
sujuka99: WDYT?

tests/pipeline/test_pipeline.py
```diff
@@ -444,6 +444,43 @@ def test_model_serialization(self, snapshot: SnapshotTest):
...
+                str(RESOURCE_PATH / "pipeline-with-short-topics"),
+            ],
+            catch_exceptions=False,
+        )
+
+        assert result.exit_code == 0
+
+        enriched_pipeline = yaml.safe_load(result.stdout)
+
+        assert isinstance(enriched_pipeline, dict)
```
disrupted, line 465: we should change this everywhere to use proper type annotations instead of runtime checks/narrowing
```suggestion
        enriched_pipeline: dict = yaml.safe_load(result.stdout)
```
but should be a followup

disrupted (changes requested):

kpops/components/base_components/models/to_section.py
```diff
@@ -93,5 +110,11 @@ class ToSection(BaseModel):
         ..., description=describe_attr("topics", __doc__)
     )
 
+    def __init__(self, **kwargs):
+        for key, value in kwargs.get("topics", {}).items():
+            if value is None:
+                kwargs["topics"][key] = TopicConfig()
```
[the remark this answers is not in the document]
sujuka99: Hmm that't true, I will do `for key, value in kwargs.get["topics"].items():` instead

Are you referring to the value of `topics` or of a topic inside?

docs/docs/schema/pipeline.json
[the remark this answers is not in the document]
disrupted: I get that this is shorter (saves 1 line each), but I am not a fan personally. Not only does it look a bit odd without the value, it would also mean that we have to change our data model to an optional: `TopicConfig | None` unless there's a way to mitigate it with a Pydantic root validator. Either way, it adds complexity which is unnecessary imo for such small gain

kpops/components/base_components/models/to_section.py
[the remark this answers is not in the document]
sujuka99: I was wondering, what does the [models](https://github.com/bakdata/kpops/pull/232/files#diff-b2c3190c16395403f228cee3b22277b4152257bf40cbbe0b108fea4659eff3b0R106-R108) attribute do?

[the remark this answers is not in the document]
disrupted: I mean that the topic value is defined as `TopicConfig`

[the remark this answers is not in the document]
sujuka99: > I mean that the topic value is defined as `TopicConfig`

The thing is that now it's possible to just mention the name of a topic and the type is implied. WIthout that line I would get errors that `None` is not allowed, didn't know how else to fix it.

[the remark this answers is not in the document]
sujuka99: > I mean that the topic value is defined as `TopicConfig`

Since I changed that to `TopicConfig | None`, I had to also handle the `None` case.

kpops/components/base_components/models/from_section.py
```diff
@@ -16,26 +16,53 @@ class InputTopicTypes(str, Enum):
...
     class Config(DescConfig):
         extra = Extra.forbid
         use_enum_values = True
 
+    def __init__(self, **kwargs):
+        kwargs["type"] = self.__assign_type(kwargs)
+        super().__init__(**kwargs)
+
+    @staticmethod
+    def __assign_type(kwargs: dict) -> str:
```
disrupted, line 48: please return `InputTopicTypes`, not `str`

disrupted, line 48: also extract the kwargs
```suggestion
    def __assign_type(type: InputTopicTypes | None, role: str | None) -> InputTopicTypes:
```

kpops/components/base_components/models/from_section.py
```diff
@@ -16,26 +16,53 @@ class InputTopicTypes(str, Enum):
...
+                else:
+                    type_ = InputTopicTypes.EXTRA.value
+            case InputTopicTypes.PATTERN:
+                if role is None:
+                    type_ = InputTopicTypes.INPUT_PATTERN.value
+                else:
+                    type_ = InputTopicTypes.EXTRA_PATTERN.value
+            case _:
+                pass
+        return type_
```
disrupted, line 64: structural pattern matching is really powerful, no need for nesting

```suggestion
        match type, role:
            case None, None:
                return InputTopicTypes.INPUT
            case None, _:
                return InputTopicTypes.EXTRA
            case InputTopicTypes.PATTERN, None:
                return InputTopicTypes.INPUT_PATTERN
            case InputTopicTypes.PATTERN, _:
                return InputTopicTypes.EXTRA_PATTERN
            case _, _:
                return type
```

disrupted (changes requested):

docs/docs/schema/pipeline.json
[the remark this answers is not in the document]
sujuka99: Fixed I think

kpops/components/base_components/models/to_section.py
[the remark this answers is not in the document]
sujuka99: fixed

kpops/components/base_components/models/from_section.py
```diff
@@ -16,26 +16,53 @@ class InputTopicTypes(str, Enum):
...
+            case None, None:
+                type_ = InputTopicTypes.INPUT
+            case None, _:
+                type_ = InputTopicTypes.EXTRA
+            case InputTopicTypes.PATTERN, None:
+                type_ = InputTopicTypes.INPUT_PATTERN
+            case InputTopicTypes.PATTERN, _:
+                type_ = InputTopicTypes.EXTRA_PATTERN
+            case _, _:
+                return type_  # type: ignore[return-value]
```
sujuka99, line 63: Had to silence mypy here, it was concerned about returning `Optional[InputTopicTypes]` instead of `InputTopicTypes`, which I'm pretty sure is not the case as the cases narrow the type down.

disrupted: I think it's because you're reassigning `type_` instead of directly returning

kpops/components/base_components/models/from_section.py
```diff
@@ -16,26 +16,53 @@ class InputTopicTypes(str, Enum):
...
 
     class Config(DescConfig):
         extra = Extra.forbid
         use_enum_values = True
 
+    def __init__(self, **kwargs):
+        kwargs["type"] = self.__assign_type(
+            kwargs.get("type", None), kwargs.get("role", None)
+        )
+        super().__init__(**kwargs)
```
disrupted, line 47: I would also suggest to extract the kwargs. this helps the type checker do its job. `**kwargs` on it's own sadly doesn't carry any types

```suggestion
    def __init__(self, type: InputTopicTypes | None = None, role: str | None = None, **kwargs) -> None:
        type = self.__assign_type(type, role)
        super().__init__(type=type, role=role)
```

kpops/components/base_components/models/from_section.py
```diff
@@ -16,26 +16,53 @@ class InputTopicTypes(str, Enum):
...
+
+    @staticmethod
+    def __assign_type(
+        type_: InputTopicTypes | None, role: str | None
+    ) -> InputTopicTypes:
+        match type_, role:
+            case None, None:
+                type_ = InputTopicTypes.INPUT
```
disrupted, line 55:
```suggestion
                return InputTopicTypes.INPUT
```

kpops/components/base_components/models/from_section.py
sujuka99: Didn't fix it unfortunately

kpops/components/base_components/models/to_section.py
```diff
@@ -74,6 +76,23 @@ def extra_topic_role(cls, values):
...
+    @staticmethod
+    def __assign_type(
+        type_: OutputTopicTypes | None, role: str | None
+    ) -> OutputTopicTypes:
+        if type_ is None:
+            if role is None:
+                type_ = OutputTopicTypes.OUTPUT
+            else:
+                type_ = OutputTopicTypes.EXTRA
+        return type_
```
disrupted, line 94: please refactor this similar to FromSection

kpops/components/base_components/models/from_section.py
disrupted: ok :( then it's not doing such a good job of type narrowing. Pyright gets it

kpops/components/base_components/models/to_section.py
sujuka99: done

kpops/components/base_components/models/from_section.py
disrupted: perhaps it's the same problem that we encountered here
https://github.com/bakdata/kpops/pull/197#discussion_r1177656235

disrupted: perhaps we should replace MyPy with Pyright in the CI, it seems a lot more powerful

kpops/components/base_components/models/from_section.py
```diff
@@ -16,26 +16,53 @@ class InputTopicTypes(str, Enum):
...
+    )
     role: str | None = Field(default=None, description=describe_attr("role", __doc__))
 
     class Config(DescConfig):
         extra = Extra.forbid
         use_enum_values = True
 
+    def __init__(
+        self, type: InputTopicTypes | None = None, role: str | None = None, **kwargs
+    ):
```
disrupted, line 45:
```suggestion
    ) -> None:
```

kpops/components/base_components/models/to_section.py
```diff
@@ -74,6 +76,23 @@ def extra_topic_role(cls, values):
             )
         return values
 
+    def __init__(
+        self, type: OutputTopicTypes | None = None, role: str | None = None, **kwargs
+    ):
```
disrupted, line 81:
```suggestion
    ) -> None:
```

tests/pipeline/resources/pipeline-with-short-topics/defaults.yaml
```diff
@@ -0,0 +1,35 @@
...
+      input-topic:
+        role: ~
+      extra-topic:
+        role: role
+      input-pattern:
+        type: pattern
+      extra-pattern:
+        type: pattern
+        role: role
+
```
disrupted, line 35: please fix the last line

disrupted (changes requested):

tests/pipeline/resources/pipeline-with-short-topics/defaults.yaml
```diff
@@ -0,0 +1,35 @@
...
+  namespace: example-namespace
+
+kafka-app:
+  app:
+    streams:
+      brokers: "${broker}"
+      schema_registry_url: "${schema_registry_url}"
+  version: "2.4.2"
+
+streams-app: # inherits from kafka
```
disrupted, line 12:
```suggestion
streams-app: # inherits from kafka-app
```

kpops/components/base_components/models/from_section.py
sujuka99: I read a bit and the MyPy vs Pyright comparison seems a bit one-sided in favor of the latter. I support the replacement. Should I create an issue?

examples/bakdata/atm-fraud-detection/pipeline.yaml
```diff
@@ -62,9 +62,8 @@
   from:
     components:
       fraud-detector:
-        type: input
+        type: ~
```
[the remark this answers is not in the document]
disrupted: like it's cool that it works implicitly but it's not something I would actually use tbh. imo explicit is better in this case

kpops/components/base_components/models/from_section.py
sujuka99: #250

docs/docs/resources/pipeline-components/kafka-app.yaml
```diff
@@ -18,14 +18,18 @@
   to:
     topics: # required
       ${pipeline_name}-output-topic:
-        type: output # required
-        # role: topic-role # only used if type is `extra`
+        type: output
+      ${pipeline_name}-extra-topic:
+        # type: extra # Implied when role is specified
+        role: topic-role # only used if type is `extra`
```
disrupted, line 24: should we remove the comment now? it reads a little bit odd with the new line above

docs/docs/resources/pipeline-components/kafka-connector.yaml
```diff
@@ -10,13 +10,37 @@
...
       ${pipeline_name}-input-topic:
-        type: input # required
-        # role: topic-role # only used if type is `extra`
+        type: input
+      ${pipeline_name}-extra-topic:
+        # type: extra # Implied when role is specified
+        role: topic-role # only used if type is `extra` or `extra-pattern`
+      ${pipeline_name}-input-pattern-topic:
+        type: pattern # Implied that type is input-pattern when role is NOT specified
+      ${pipeline_name}-input-pattern-topic:
```
disrupted, line 19:
```suggestion
      ${pipeline_name}-extra-pattern-topic:
```

docs/docs/resources/pipeline-components/kafka-sink-connector.yaml
```diff
@@ -11,15 +11,37 @@
...
+        type: pattern # Implied that type is extra-pattern when role is specified
+        role: some-role
     components: # read from specific component
-      - producer
+      account-producer:
+        type: output # Implied when role is NOT specified
+      other-producer:
+        # type: extra # Implied when role is specified
+        role: some-role # only used if type is `extra`
+      component-as-pattern:
```
disrupted, line 29:
```suggestion
      component-as-input-pattern:
```

docs/docs/schema/pipeline.json
```diff
@@ -91,6 +88,7 @@
             "enum": [
                 "input",
                 "extra",
+                "pattern",
```
philipp94831, line 91: Why is there still extra? And why is there pattern and input-pattern?

docs/docs/resources/pipeline-components/kubernetes-app.yaml
```diff
@@ -12,16 +12,37 @@
...
+        type: pattern # Implied that type is extra-pattern when role is specified
+        role: some-role
     components: # read from specific component
-      - producer
+      account-producer:
+        # type: output # Implied when role is NOT specified
+      other-producer:
+        # type: extra # Implied when role is specified
+        role: some-role # only used if type is `extra`
+      component-as-pattern:
```
disrupted, line 30:
```suggestion
      component-as-input-pattern:
```

philipp94831 (changes requested):

docs/docs/resources/pipeline-components/streams-app.yaml
```diff
@@ -63,21 +63,42 @@
...
       ${pipeline_name}-input-topic:
-        type: input # required
-        # role: topic-role # only used if type is `extra` or `extra-pattern`
+        type: input
+      ${pipeline_name}-extra-topic:
+        # type: extra # Implied when role is specified
+        role: topic-role # only used if type is `extra` or `extra-pattern`
+      ${pipeline_name}-input-pattern-topic:
+        type: pattern # Implied that type is input-pattern when role is NOT specified
+      ${pipeline_name}-input-pattern-topic:
```
disrupted, line 72:
```suggestion
      ${pipeline_name}-extra-pattern-topic:
```

docs/docs/resources/pipeline-components/streams-app.yaml
```diff
@@ -63,21 +63,42 @@
...
+        type: pattern # Implied that type is extra-pattern when role is specified
+        role: some-role
     components: # read from specific component
-      - producer
+      account-producer:
+        type: output # Implied when role is NOT specified
+      other-producer:
+        # type: extra # Implied when role is specified
+        role: some-role # only used if type is `extra`
+      component-as-pattern:
```
disrupted, line 81:
```suggestion
      component-as-input-pattern:
```

docs/docs/resources/pipeline-defaults/defaults-kafka-connector.yaml
```diff
@@ -11,15 +11,39 @@ kafka-connector:
...
+    topics: # required
       ${pipeline_name}-input-topic:
         type: input
-        # role: topic-role # only used if type is `extra`
+      ${pipeline_name}-extra-topic:
+        # type: extra # Implied when role is specified
+        role: topic-role # only used if type is `extra` or `extra-pattern`
+      ${pipeline_name}-input-pattern-topic:
+        type: pattern # Implied that type is input-pattern when role is NOT specified
+      ${pipeline_name}-input-pattern-topic:
```
disrupted, line 22:
```suggestion
      ${pipeline_name}-extra-pattern-topic:
```

docs/docs/resources/pipeline-defaults/defaults-kafka-connector.yaml
```diff
@@ -11,15 +11,39 @@ kafka-connector:
...
+        type: pattern # Implied that type is extra-pattern when role is specified
+        role: some-role
+    components: # read from specific component
+      account-producer:
+        type: output # Implied when role is NOT specified
+      other-producer:
+        # type: extra # Implied when role is specified
+        role: some-role # only used if type is `extra`
+      component-as-pattern:
```
disrupted, line 31:
```suggestion
      component-as-input-pattern:
```

docs/docs/resources/pipeline-defaults/defaults-kubernetes-app.yaml
```diff
@@ -13,16 +13,39 @@ kubernetes-app:
...
+      ${pipeline_name}-input-pattern-topic:
+        type: pattern # Implied that type is extra-pattern when role is specified
+        role: some-role
+    components: # read from specific component
+      account-producer:
+        type: output # Implied when role is NOT specified
+      other-producer:
+        # type: extra # Implied when role is specified
+        role: some-role # only used if type is `extra`
+      component-as-pattern:
```
disrupted, line 33:
```suggestion
      component-as-input-pattern:
```

docs/docs/resources/pipeline-defaults/defaults.yaml
```diff
@@ -143,15 +148,39 @@ kafka-connector:
...
+    topics: # required
       ${pipeline_name}-input-topic:
         type: input
-        # role: topic-role # only used if type is `extra`
+      ${pipeline_name}-extra-topic:
+        # type: extra # Implied when role is specified
+        role: topic-role # only used if type is `extra` or `extra-pattern`
+      ${pipeline_name}-input-pattern-topic:
+        type: pattern # Implied that type is input-pattern when role is NOT specified
+      ${pipeline_name}-input-pattern-topic:
```
disrupted, line 159:
```suggestion
      ${pipeline_name}-extra-pattern-topic:
```

docs/docs/resources/pipeline-defaults/defaults.yaml
```diff
@@ -143,15 +148,39 @@ kafka-connector:
...
+        type: pattern # Implied that type is extra-pattern when role is specified
+        role: some-role
+    components: # read from specific component
+      account-producer:
+        type: output # Implied when role is NOT specified
+      other-producer:
+        # type: extra # Implied when role is specified
+        role: some-role # only used if type is `extra`
+      component-as-pattern:
```
disrupted, line 168:
```suggestion
      component-as-input-pattern:
```

kpops/components/base_components/models/from_section.py
```diff
@@ -16,26 +16,53 @@ class InputTopicTypes(str, Enum):
...
     """Input topic
 
-    :param type: Topic type
-    :type type: InputTopicTypes
-    :param role: Custom identifier belonging to a topic, provide only if `type` is `extra` or `extra-pattern`
-    :type role: str | None
+    :param type: Topic type, defaults to None
+    :type type: InputTopicTypes | None, optional
+    :param role: Custom identifier belonging to a topic, provide only if `type` is `extra` or `extra-pattern`.
+        When `role` is not `None`, `type: pattern` is equal to `type: extra-pattern`, defaults to None
```
disrupted, line 30: perhaps easier to understand?
```suggestion
        When `role` is defined, `type: pattern` is equal to `type: extra-pattern`, defaults to None
```

kpops/components/base_components/models/from_section.py
```diff
@@ -16,26 +16,53 @@ class InputTopicTypes(str, Enum):
...
     role: str | None = Field(default=None, description=describe_attr("role", __doc__))
 
     class Config(DescConfig):
         extra = Extra.forbid
         use_enum_values = True
 
+    def __init__(
+        self, type: InputTopicTypes | None = None, role: str | None = None, **kwargs
+    ) -> None:
+        type = self.__assign_type(type, role)
```
disrupted, line 46: would it actually be possible to use a Pydantic `root_validator` instead?

docs/docs/resources/pipeline-components/kafka-app.yaml
sujuka99: yeah, but it must be reworded somehow to make it clear that role should not be defined for any type other than extra

tests/components/test_streams_app.py
```diff
@@ -178,6 +178,27 @@ def test_should_validate(self, config: PipelineConfig, handlers: ComponentHandle
                 },
             )
 
+        with pytest.raises(ValueError):
```
disrupted, line 181: add a comment why this fails. I am assuming because role is not allowed for type input

disrupted (changes requested):

docs/docs/schema/pipeline.json
sujuka99: Otherwise there would be breaking changes

docs/docs/resources/pipeline-components/kafka-source-connector.yaml
```diff
@@ -12,8 +12,13 @@
...
       ${pipeline_name}-output-topic:
-        type: output # required
-        # role: topic-role # only used if type is `extra`
+        type: output
+      ${pipeline_name}-extra-topic:
+        # type: extra # Implied when role is specified
+        role: topic-role # only used if type is `extra`
+      ${pipeline_name}-error-topic:
+        type: error
+        # Currently KPOps only supports Avro schemas.
```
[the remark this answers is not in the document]
sujuka99: Hmm, it's an outdated comment that I didn't bother to fact check and copied everywhere. Will fix.

tests/components/test_streams_app.py
sujuka99: You are right in your assumption, I will.

docs/docs/resources/pipeline-components/kafka-app.yaml
sujuka99: done

kpops/components/base_components/models/from_section.py
sujuka99: Yeah, I added it inside. Should I split it into 2 separate ones?

kpops/components/base_components/models/from_section.py
```diff
@@ -38,19 +42,39 @@ class Config(DescConfig):
...
+                return values
+            case None, True:
+                values["type"] = InputTopicTypes.EXTRA
+                return values
+            case InputTopicTypes.PATTERN, False:
+                values["type"] = InputTopicTypes.INPUT_PATTERN
+                return values
+            case InputTopicTypes.PATTERN, True:
+                values["type"] = InputTopicTypes.EXTRA_PATTERN
+                return values
```
disrupted, line 61: just to confirm: the check below is unnecessary in this case, right? hence the early return you're doing here

kpops/components/base_components/models/from_section.py
```diff
@@ -38,19 +42,39 @@ class Config(DescConfig):
...
-            raise ValueError(
-                "If you define an extra input component, extra input topic, or extra input pattern, you have to define a role."
-            )
-        if not is_extra_topic and values.get("role"):
-            raise ValueError(
-                "If you do not define an input component, input topic, or input pattern, the role is unnecessary."
-            )
+        """Ensure that cls.role is used correctly, assign type if needed"""
+        type_ = values.get("type", None)
+        role = values.get("role", None) is not None
```
disrupted, line 47: why are you using `.get(...)`? they are always defined

kpops/components/base_components/models/from_section.py
```diff
@@ -38,19 +42,39 @@ class Config(DescConfig):
...
+                values["type"] = InputTopicTypes.INPUT_PATTERN
+                return values
+            case InputTopicTypes.PATTERN, True:
+                values["type"] = InputTopicTypes.EXTRA_PATTERN
+                return values
+            case _, _:
+                is_input_topic = values["type"] not in (
+                    InputTopicTypes.EXTRA,
+                    InputTopicTypes.EXTRA_PATTERN,
+                )
```
disrupted, line 66: you can dedent this and remove the fallthrough case

kpops/components/base_components/models/from_section.py
```diff
@@ -38,19 +42,39 @@ class Config(DescConfig):
...
+        match is_input_topic, role:
+            case True, True:
+                raise ValueError(
+                    "If you define an extra input component, extra input topic, or extra input pattern, you have to define a role."
+                )
```
disrupted, line 71: shouldn't it be `True`, `False`?

disrupted (changes requested):

kpops/components/base_components/models/from_section.py
```diff
@@ -38,19 +42,39 @@ class Config(DescConfig):
...
+            case False, False:
+                raise ValueError(
+                    "If you do not define an extra input component, input topic, or input pattern, the role is unnecessary."
+                )
+            case _, _:
+                return values
```
disrupted, line 77: unnecessary since you have a return below

docs/docs/schema/pipeline.json
philipp94831: I think that would be fine, user base is small. What do you think @disrupted @raminqaf ?

kpops/components/base_components/models/from_section.py
disrupted: ah, I see you reversed the `is_extra_topic` to `is_input_topic`. may I ask why?

docs/docs/schema/pipeline.json
philipp94831: Otherwise we keep carrying technical debt because decision in the beginning weren't ideal

kpops/components/base_components/models/from_section.py
sujuka99: Ah, I was wondering if it is okay not to have a default case and read that all cases should be covered. Will fix

docs/docs/schema/pipeline.json
disrupted: I am fine with it, but keep in mind we would have to do a major version bump for the next release, aka kpops 2.0

kpops/components/base_components/models/from_section.py
sujuka99: I think I got confused being defined as None with being undefined. Will fix

kpops/components/base_components/models/from_section.py
disrupted: is that possible?

kpops/components/base_components/models/from_section.py
sujuka99: No particular reason, I kept mishandling the cases and just ended up on that while trying to make it work. I guess it was easier to follow at the time. I will switch it around though, no need for the negation.

kpops/components/base_components/models/from_section.py
sujuka99: yeah, if `pattern` and `role`, `type=extra-pattern`

docs/docs/schema/pipeline.json
disrupted: our current plan was to make changes in a backwards compatible manner and then once those changes pile up make a breaking change for v2

kpops/components/base_components/models/from_section.py
sujuka99: The check below exists, because right now the user is free to set the type to `extra-pattern` or `extra`

docs/docs/schema/pipeline.json
philipp94831: I consider KPOps still beta wrt to API and I think it is much more beneficial to release often and document breaking changes if needed. KPOps is not ages old :-)

disrupted: yeah, but we're not pre-release (0.x.x) anymore. Since we use semantic versioning we should follow its rules.

philipp94831: I mean it is also fine to have v2 and quite soon v3

disrupted: yep, absolutely!

sujuka99: Sounds good. Will carry out the breaking changes in this PR

kpops/components/base_components/models/from_section.py
sujuka99: it is possible, but now with the breaking changes, I don't think it will be needed

kpops/components/base_components/pipeline_component.py
```diff
@@ -154,14 +154,16 @@ def apply_from_inputs(self, name: str, topic: FromTopic) -> None:
         :type topic: FromTopic
         """
         match topic.type:
+            case None:
+                self.add_extra_input_topic(
+                    topic.role, [name]  # pyright: ignore [reportGeneralTypeIssues]
```
disrupted, line 159: don't we need to check if `topic.role` is set?

sujuka99: It's covered in the `from` [validator](https://github.com/bakdata/kpops/blob/refactor/input-output-types/kpops/components/base_components/models/from_section.py#L45-L47)

Now that I see it, perhaps the validator doesn't need to assign a value to `type` and instead should only validate?
Same happens with the `to` section.

disrupted: yeah, I guess you could technically skip the assignment if you're covering `None` here anyways.

docs/docs/schema/pipeline.json
sujuka99: @philipp94831 I think I'm done with the changes to the code, what do you think?

docs/docs/resources/pipeline-components/kafka-sink-connector.yaml
```diff
@@ -11,15 +11,34 @@
...
     components: # read from specific component
-      - producer
+      account-producer:
+        type: output # Implied when role is NOT specified, alternatives of `type: null` or `role: null` also work
+      other-producer:
+        role: some-role # Implies `type` to be extra
+      component-as-input-pattern:
+        type: pattern # Implied to be an input pattern if `role` is undefined
+      component-as-extra-pattern:
+        type: pattern # Implied to be an extra pattern if `role` undefined
```
disrupted, line 30:
```suggestion
        type: pattern # Implied to be an extra pattern if `role` is defined
```

disrupted (changes requested):

disrupted (changes requested):

docs/docs/resources/pipeline-components/kafka-app.yaml
```diff
@@ -5,21 +5,37 @@
   # Pipeline prefix that will prefix every component name. If you wish to not
   # have any prefix you can specify an empty string.
   prefix: ${pipeline_name}-
-  # Topic(s) from which the component will read input
   from: # Must not be null
     topics: # read from topic
       ${pipeline_name}-input-topic:
-        type: input # required
-        # role: topic-role # only used if type is `extra` or `extra-pattern`
+        type: input # Implied when role is NOT specified, alternatives of `type: null` or `role: null` also work
```
philipp94831, line 11: I don't quite understand the comment. I also think that while it works with type or role null, that is more a side-effect of pydantic or whatever and should not really be documented as a feature

tests/pipeline/resources/pipeline-with-short-topics/defaults.yaml
```diff
@@ -0,0 +1,34 @@
...
+      output-topic:
+        type: 
+      error-topic:
+        type: error
+      extra-topic:
+        role: role
+  from:
+    topics:
+      input-topic:
+        role: ~
```
philipp94831, line 27: what kind of role is that?

philipp94831 (changes requested):
I really like the way the result turned out

tests/pipeline/resources/pipeline-with-short-topics/defaults.yaml
sujuka99: Equal to `null`

philipp94831: Hm, then I would just put `type: input` there because this is quite confusing

kpops/components/base_components/models/from_section.py
```diff
@@ -39,19 +37,9 @@ class Config(DescConfig):
 
     @root_validator
     def extra_topic_role(cls, values: dict) -> dict:
```
disrupted, line 39:
```suggestion
    def extra_topic_role(cls, values: dict[str, Any]) -> dict[str, Any]:
```

kpops/components/base_components/models/to_section.py
```diff
@@ -64,17 +64,9 @@ class Config(DescConfig):
 
     @root_validator
     def extra_topic_role(cls, values):
```
disrupted, line 66:
```suggestion
    def extra_topic_role(cls, values: dict[str, Any]) -> dict[str, Any]:
```

disrupted (changes requested):

docs/docs/resources/pipeline-components/kafka-app.yaml
sujuka99: I intended to say with the comment that the following are equal:

```
some-topic:
	type:
```
```
some-topic:
	type: ~
```
```
some-topic:
	type: null
```
```
some-topic:
	role:
```
```
some-topic:
	role: ~
```
```
some-topic:
	role: null
```

It would work the same with any attribute of `some-topic`


I originally wanted to allow default definitions of topics like the example bellow, but the added complexity wasn't worth it:
```
# Equivalent to specifying `type: input`
some-topic:
```

tests/pipeline/resources/pipeline-with-short-topics/defaults.yaml
sujuka99: I did it to ensure that if such definitions ever stop working we would know. Should I comment an explanation or replace with `type: input`

docs/docs/resources/pipeline-components/kafka-app.yaml
philipp94831: I still think all this null stuff is more confusing than helpful. So I'd rather skip documenting it and treat it as a side-effect

tests/pipeline/resources/pipeline-with-short-topics/defaults.yaml
philipp94831: Because it is a side-effect, I wouldn't even have it tested

sujuka99: Done

docs/docs/resources/pipeline-components/kafka-app.yaml
sujuka99: Done

philipp94831 (approved):

docs/docs/resources/pipeline-components/kafka-connector.yaml
```diff
@@ -3,21 +3,37 @@
...
       ${pipeline_name}-input-topic:
-        type: input # required
-        # role: topic-role # only used if type is `extra` or `extra-pattern`
+        type: input # Implied when role is NOT specified
+      ${pipeline_name}-extra-topic:
+        role: topic-role # Implies `type` to be extra
+      ${pipeline_name}-input-pattern-topic:
+        type: pattern # Implied to be an input pattern if `role` is undefined
+      ${pipeline_name}-extra-pattern-topic:
+        type: pattern # Implied to be an extra pattern if `role` undefined
```
disrupted, line 15: same as in other places
```suggestion
        type: pattern # Implied to be an extra pattern if `role` is defined
```

disrupted (changes requested):

docs/docs/resources/pipeline-components/kafka-connector.yaml
sujuka99: Ah, I missed that one. BTW [`sections/_from.yaml`](https://github.com/bakdata/kpops/blob/refactor/input-output-types/docs/docs/resources/pipeline-components/sections/from_.yaml#L10) is the only place where a manual change is needed. Such a relief!

PR #239, opened 2023-06-08
Add helm repo update <repo-name> for helm >3.7

### LINKED ISSUE #18: Use `helm repo update <repo_name>` instead of updating every repo
Currently, when adding a new helm repo in the Helm wrapper, the helm repo update causes to update all Helm repositories. With Helm 3.7 one can now specify the repo name to update. Introducing this requires that every project that uses KPOps uses helm > 3.7, which is not the case right now.

raminqaf:
fixes #18

### DIFF
--- a/kpops/component_handlers/helm_wrapper/helm.py
+++ b/kpops/component_handlers/helm_wrapper/helm.py
@@ -68,7 +68,13 @@ class Helm:
             else:
                 raise e
 
-        self.__execute(["helm", "repo", "update"])
+        version = self.get_version()
+        _, minor, _ = map(int, version.split("."))
+
+        if minor > 7:
+            self.__execute(["helm", "repo", "update", repository_name])
+        else:
+            self.__execute(["helm", "repo", "update"])
 
     def upgrade_install(
         self,
@@ -177,6 +183,15 @@ class Helm:
             return Helm.load_manifest(stdout)
         except ReleaseNotFoundException:
             return ()
+
+    def get_version(self) -> str:
+        command = ["helm", "version", "--short"]
+        short_version = self.__execute(command)
+        version_match = re.search(r"v(\d+\.\d+\.\d+)", short_version)
+        if version_match is None:
+            raise RuntimeError("Could not parse the Helm version.")
+
+        return version_match.group(1)
 
     @staticmethod
     def load_manifest(yaml_contents: str) -> Iterator[HelmTemplate]:
--- a/tests/component_handlers/helm_wrapper/test_helm_wrapper.py
+++ b/tests/component_handlers/helm_wrapper/test_helm_wrapper.py
@@ -62,10 +62,12 @@ class TestHelmWrapper:
             ],
         )
 
-    def test_should_include_configured_tls_parameters_on_add(
+    def test_should_include_configured_tls_parameters_on_add_when_version_is_old(
         self, run_command: MagicMock
     ):
         helm = Helm(HelmConfig())
+        run_command.return_value = "v3.6.0+gc9f554d"
+
         helm.add_repo(
             "test-repository",
             "fake",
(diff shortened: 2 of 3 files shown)


kpops/component_handlers/helm_wrapper/helm.py
```diff
@@ -68,7 +68,13 @@ def add_repo(
             else:
                 raise e
 
-        self.__execute(["helm", "repo", "update"])
+        version = self.get_version()
+        _, minor, _ = map(int, version.split("."))
+
+        if minor > 7:
```
disrupted, line 74: shouldn't we also check that major >= 3?

kpops/component_handlers/helm_wrapper/helm.py
```diff
@@ -178,6 +184,15 @@ def get_manifest(self, release_name: str, namespace: str) -> Iterable[HelmTempla
...
             return ()
 
+    def get_version(self) -> str:
+        command = ["helm", "version", "--short"]
+        short_version = self.__execute(command)
+        version_match = re.search(r"v(\d+\.\d+\.\d+)", short_version)
+        if version_match is None:
+            raise RuntimeError("Could not parse the Helm version.")
+
+        return version_match.group(1)
```
disrupted, line 194: consider returning a tuple with major, minor, patch (+ sha or whatever) directly to make it easier to process in the consuming function

kpops/component_handlers/helm_wrapper/helm.py
raminqaf: I was not really sure about that. But I think it is not a bad idea. Maybe I can push the check of the major a bit further up in the method.

kpops/component_handlers/helm_wrapper/helm.py
```diff
@@ -26,6 +27,11 @@ class Helm:
     def __init__(self, helm_config: HelmConfig) -> None:
         self._context = helm_config.context
         self._debug = helm_config.debug
+        self._version = self.get_version()
+        if self._version.major < 3:
+            raise RuntimeError(
+                f"The supported Helm version is >3.x.x The current Helm version {self._version.major}.{self._version.minor}.{self._version.patch}"
```
sujuka99, line 33:
```suggestion
                f"The supported Helm version is >3.x.x The current Helm version is {self._version.major}.{self._version.minor}.{self._version.patch}"
```

kpops/component_handlers/helm_wrapper/helm.py
sujuka99: Do major versions support backwards compatible syntax, i.e. could major > 3 introduce a breaking change?

sujuka99 (changes requested):

kpops/component_handlers/helm_wrapper/helm.py
disrupted: also `>=` instead of `>`

raminqaf: @disrupted fixed

disrupted: @raminqaf: @sujuka99's suggestion is not fixed

kpops/component_handlers/helm_wrapper/helm.py
disrupted: good point! I would also prefer pinning the major version to 3. Once Helm 4 becomes a thing we get a chance to confirm it is compatible.

disrupted (changes requested):
please address @sujuka99 feedback, otherwise LGTM

kpops/component_handlers/helm_wrapper/helm.py
```diff
@@ -26,6 +27,11 @@ class Helm:
     def __init__(self, helm_config: HelmConfig) -> None:
         self._context = helm_config.context
         self._debug = helm_config.debug
+        self._version = self.get_version()
+        if self._version.major != 3:
+            raise RuntimeError(
+                f"The supported Helm version is 3.x.x The current Helm version is {self._version.major}.{self._version.minor}.{self._version.patch}"
```
disrupted, line 33: There is also a period missing in between the sentences 😬
```suggestion
                f"The supported Helm version is 3.x.x. The current Helm version is {self._version.major}.{self._version.minor}.{self._version.patch}"
```

raminqaf: Done

disrupted (approved):

sujuka99 (approved):

PR #247, opened 2023-06-13
Update Poetry version in CI

sujuka99:


.github/actions/update-docs/action.yml
```diff
@@ -24,7 +24,7 @@ runs:
     - name: Install python and set up Poetry
```
disrupted, line 24: please capitalize Python

sujuka99: done

.github/workflows/ci.yaml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: Install python and set up Poetry
```
disrupted, line 24: here as well

disrupted (changes requested):

.github/workflows/ci.yaml
sujuka99: done

disrupted (approved):

PR #252, opened 2023-06-15
Remove MyPy

### LINKED ISSUE #250: Replace MyPy with Pyright
Pyright seems more powerful than MyPy, so the project should benefit from the replacement. It is also well-supported and in active development.

I also encountered reports that features consistently come out in Pyright first.

disrupted:
Closes #250
followup of #251 (transition from MyPy to Pyright)


sujuka99 (approved):
LGTM

PR #257, opened 2023-06-21
Fix release & publish workflows

irux:


disrupted (changes requested):
please also update the publish workflow with the fix for PyPI

.github/workflows/release.yaml
```diff
@@ -15,7 +15,7 @@ on:
 
 jobs:
   create-github-release-push-tag:
-    uses: bakdata/ci-templates/.github/workflows/python-poetry-release.yaml@v1.23.0
+    uses: bakdata/ci-templates/.github/workflows/python-poetry-release.yaml@1.25.2
```
disrupted, line 18: shouldn't it be 1.25.3?

disrupted (changes requested):

disrupted (approved):

PR #265, opened 2023-06-21
Refactor custom component validation & hide field from kpops output

### LINKED ISSUE #263: Internal `validate_name` field should be hidden in kpops output
currently the output of `kpops generate` and dry-run shows the [`validate_name`](https://github.com/bakdata/kpops/blob/fd19c6b706012d130e9591b8ab7a8ff63a737933/kpops/components/base_components/kubernetes_app.py#L65) field causing some snapshots to break. This field should be hidden. Additionally, we should consider refactoring this validation feature to be a bit more general purpose, not just tied specifically to the name of Kubernetes components.

disrupted:
Fixes #263

### DIFF
--- a/kpops/components/base_components/base_defaults_component.py
+++ b/kpops/components/base_components/base_defaults_component.py
@@ -34,6 +34,7 @@ class BaseDefaultsComponent(BaseModel):
     :type config: PipelineConfig
     :param handlers: Component handlers to be accessed by this component
     :type handlers: ComponentHandlers
+    :param validate: Whether to run custom validation on the component, defaults to True
     """
 
     type: str = Field(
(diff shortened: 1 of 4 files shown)


kpops/pipeline_generator/pipeline.py
```diff
@@ -235,8 +235,8 @@ def enrich_component(
             # HACK: Pydantic .dict() doesn't create jsonable dict
             json.loads(component.json(by_alias=True)),
         )
-        if "validate_name" in env_component_as_dict:
-            del env_component_as_dict["validate_name"]
+        if "validate" in env_component_as_dict:
+            del env_component_as_dict["validate"]
```
disrupted, line 239: @sujuka99 why do we delete it for the substitution + final initialization?

kpops/components/base_components/base_defaults_component.py
```diff
@@ -66,6 +75,8 @@ def __init__(self, **kwargs) -> None:
         if kwargs.get("enrich", True):
             kwargs = self.extend_with_defaults(**kwargs)
         super().__init__(**kwargs)
+        if kwargs.get("validate", True):
```
disrupted, line 78: todo: check using debugger that it is passed as `validate` not `validate_`

sujuka99 (commented):
LGTM!

Soo the problem was that classes not inheriting from `KubernetesApp` weren't instructed to hide the field?

sujuka99 (approved):

PR #267, opened 2023-06-23
Modularize and autogenerate examples for the documentation

### LINKED ISSUE #248: Concatenate examples for the docs in the pre-commit hook
The example [pipeline definition](https://github.com/bakdata/kpops/blob/main/docs/docs/resources/pipeline-components/pipeline.yaml) is created by concatenating the individual examples in its folder.
Same for the [defaults](https://github.com/bakdata/kpops/blob/main/docs/docs/resources/pipeline-defaults/defaults.yaml)

Automate this via the [gen_docs](https://github.com/bakdata/kpops/blob/main/hooks/gen_docs.py) hook.

sujuka99:
closes #248

Additional features:
- Generates component-specific example definitions and defaults from manually written section configs.
- Each unique section needs to be defined only once in a separate `yaml` file.  
	Example: `namespace` is specified once and used for all components
- Changes to the KPOps components' structure and hierarchy are detected and trigger an update to the dependency files. 
- Changes to the filenames in `docs/docs/pipeline-components/resources/sections` are detected and trigger an update to the dependency files.


hooks/gen_docs.py
```diff
@@ -1,10 +1,169 @@
...
+    "resetter_values",
+    "offset_topic",
+]
+
+#####################
+# EXAMPLES          #
+#####################
+
+
+def get_sections(
```
irux, line 42: After looking at the PR I think one improvement we can do is to move the dependencies as a static file where we define the possible sections of each component. In that way we can save time and reprocessing (like sorting, etc)

sujuka99: I did that, but added a check if the dependencies are up to date

sujuka99:
:( I don't get why the [checks fail](https://github.com/bakdata/kpops/actions/runs/5422933819/jobs/9860204060?pr=267) like that if it works locally

docs/docs/schema/pipeline.json
```diff
@@ -288,7 +288,7 @@
                 },
                 "type": {
                     "default": "kafka-connector",
-                    "description": "Base class for all Kafka connectors\nShould only be used to set defaults",
+                    "description": "Used for schema generation, same as :param:`type`",
```
disrupted, line 291: isn't that the description of the `schema_type` attribute?

kpops/components/base_components/kafka_connector.py
```diff
@@ -58,11 +58,13 @@ class KafkaConnector(PipelineComponent, ABC):
...
     """
 
-    type: str = Field(default="kafka-connector", description="Component type")
+    type: str = Field(
+        default="kafka-connector", description=describe_attr("type", __doc__)
+    )
     schema_type: Literal["kafka-connector"] = Field(
         default="kafka-connector",
         title="Component type",
-        description=describe_object(__doc__),
```
sujuka99, line 65: revert

docs/docs/schema/pipeline.json
sujuka99: Oof, yeah, idk what I was thinking...

[noted](https://github.com/bakdata/kpops/pull/267#discussion_r1252237204)

sujuka99:
TODO: Detect changes in the sections dir

disrupted:
> TODO: Detect changes in the sections dir

thinking out loud here: I am wondering whether we could use the actual config of the pre-commit hook for the change detection locally. And in the CI it could run and verify regardless if there were changes or not. wdyt?

hooks/gen_docs.py
```diff
@@ -1,10 +1,263 @@
...
+PATH_DOCS_COMPONENTS_DEPENDENCIES = (
+    PATH_DOCS_COMPONENTS / "dependencies/pipeline_component_dependencies.yaml"
+)
+PATH_DOCS_COMPONENTS_DEPENDENCIES_DEFAULTS = (
+    PATH_DOCS_COMPONENTS / "dependencies/defaults_pipeline_component_dependencies.yaml"
+)
+PATH_DOCS_KPOPS_STRUCTURE = PATH_DOCS_COMPONENTS / "dependencies/kpops_structure.yaml"
+KPOPS_COMPONENTS = tuple(_find_classes("kpops.components", PipelineComponent))
+KPOPS_COMPONENTS_INHERITANCE_REF: dict = {
+    component.get_component_type(): component.__base__.get_component_type()  # type: ignore[reportGeneralTypeIssues]
```
disrupted, line 26: does this dict hold all kpops component classes with their direct parent/base class? what if it would inherit from multiple bases? (hypothetically, not sure if that's the case for any of the builtin components)

disrupted: we have something similar here, just FYI: https://github.com/bakdata/kpops/blob/2804fbec0e4fef86f0a0f65f4ec5843fd2fc69c8/kpops/components/base_components/base_defaults_component.py#L143
this could also be extracted into an independent function in case you'd wanna reuse it

hooks/gen_docs.py
```diff
@@ -1,10 +1,263 @@
...
+    for component in KPOPS_COMPONENTS
+}
+KPOPS_COMPONENTS_FIELDS = {
+    # `set` to make it pickleable
+    component.get_component_type(): set(component.__fields__.keys())
+    for component in KPOPS_COMPONENTS
+}
+SECTIONS_ORDER = [
```
disrupted, line 34: 👍🏻 for fixed ordering

hooks/gen_docs.py
```diff
@@ -1,10 +1,263 @@
...
+    """
+    component_sections: list[str] = []
+    for target_section in KPOPS_COMPONENTS_FIELDS[component_name]:
+        component = component_name
+        section = target_section + "-" + component + ".yaml"
+        if section in sections:
+            component_sections.append(section)
+        elif (
+            not include_inherited
+            and KPOPS_COMPONENTS_INHERITANCE_REF[component] == "pipeline-component"
```
disrupted, line 72:
```suggestion
            and KPOPS_COMPONENTS_INHERITANCE_REF[component] == PipelineComponent.type
```

hooks/gen_docs.py
```diff
@@ -1,10 +1,263 @@
...
+
+    :param component_name: Component name
+    :param sections: Available section files, names with extension, no path
+    :param include_inherited: Whether to include sections that can be defined in
+        a parent component defaults
+    :returns: The applicable sections
+    """
+    component_sections: list[str] = []
+    for target_section in KPOPS_COMPONENTS_FIELDS[component_name]:
+        component = component_name
```
disrupted, line 66: why not leave it as `component_name` if it's just a str, not the actual component class?

hooks/gen_docs.py
```diff
@@ -1,10 +1,263 @@
...
+            if section in sections:
+                component_sections.append(section)
+        elif include_inherited:
+            while component != "pipeline-component":
+                component = KPOPS_COMPONENTS_INHERITANCE_REF[component]
+                section = target_section + "-" + component + ".yaml"
+                if section in sections:
+                    component_sections.append(section)
+                    break
+                elif component == "pipeline-component":
```
disrupted, line 84: same as above: replace the hardcoded string

hooks/gen_docs.py
```diff
@@ -1,10 +1,263 @@
...
+        retained in the result
+    :param target: File to store the result in. Note that any previous contents
+        will be deleted.
+    """
+    res = ""
+    for file in file_paths:
+        with open(file, "r") as f:
+            res += f.read()
+    with open(target, "w+") as f:
+        f.write(res)
```
disrupted, line 106: I would prefer skipping the intermediate string assignment. you could write directly to the file in append mode
something like this 
```suggestion
    with target.open("a") as f:
        for source in file_paths:
            f.write(source.read_text())
```

hooks/gen_docs.py
```diff
@@ -1,10 +1,263 @@
...
+                elif component == "pipeline-component":
+                    section = target_section + ".yaml"
+                    if section in sections:
+                        component_sections.append(section)
+                        break
+    component_sections.sort(key=component_section_position_in_definition)
+    return component_sections
+
+
+def concatenate_text_files(*file_paths: Path, target: Path) -> None:
```
disrupted, line 93: nice use of the star operator for unpacking! 👍🏻 
make `target` a kw-only argument, it's slightly nicer :)
```suggestion
def concatenate_text_files(*file_paths: Path, *, target: Path) -> None:
```

disrupted: also consider renaming `file_paths` to `sources` (or `source_paths` and `target_path` if you prefer the longer names)

hooks/gen_docs.py
```diff
@@ -1,10 +1,263 @@
...
+        "kpops_components_inheritance_ref": KPOPS_COMPONENTS_INHERITANCE_REF,
+        "kpops_components_fields": KPOPS_COMPONENTS_FIELDS,
+    }
+    if kpops_new_structure != kpops_structure:
+        with open(PATH_DOCS_KPOPS_STRUCTURE, "w+") as f:
+            yaml.dump(kpops_new_structure, f)
+        with open(PATH_DOCS_COMPONENTS_DEPENDENCIES, "w+"):
+            pass
+        with open(PATH_DOCS_COMPONENTS_DEPENDENCIES_DEFAULTS, "w+"):
+            pass
```
disrupted, line 145: what does this do?

hooks/gen_docs.py
```diff
@@ -1,10 +1,263 @@
...
+        print("KPOps components' structure has changed, updating dependencies.")
+        return True
+    return False
+
+
+def get_sections(exist_changes: bool) -> tuple[list[str], list[str]]:
```
disrupted, line 151: consider returning a `NamedTuple`

disrupted (changes requested):

disrupted:
the title is underselling the value of this PR :D you could indicate we're moving to better modularization and auto-generation of the docs with this

hooks/gen_docs.py
disrupted: if you're going for the current approach you could help type checker out here and get rid of the pragma by using
```python
cast(type[PipelineComponent], component.__base__).get_component_type()
```

hooks/gen_docs.py
```diff
@@ -1,10 +1,263 @@
...
+PATH_DOCS_COMPONENTS = PATH_DOCS_RESOURCES / "pipeline-components"
+PATH_DOCS_COMPONENTS_DEPENDENCIES = (
+    PATH_DOCS_COMPONENTS / "dependencies/pipeline_component_dependencies.yaml"
+)
+PATH_DOCS_COMPONENTS_DEPENDENCIES_DEFAULTS = (
+    PATH_DOCS_COMPONENTS / "dependencies/defaults_pipeline_component_dependencies.yaml"
+)
+PATH_DOCS_KPOPS_STRUCTURE = PATH_DOCS_COMPONENTS / "dependencies/kpops_structure.yaml"
+KPOPS_COMPONENTS = tuple(_find_classes("kpops.components", PipelineComponent))
+KPOPS_COMPONENTS_INHERITANCE_REF: dict = {
```
disrupted, line 25: either add complete type annotations or remove it and let the type checker infer it (latter is preferred if possible)

hooks/gen_docs.py
```diff
@@ -1,10 +1,263 @@
...
+    concatenate_text_files(
+        *(sections_paths),
+        target=PATH_DOCS_RESOURCES / "pipeline-components/" / component_file_name,
+    )
+
+concatenate_text_files(
+    *(
+        PATH_DOCS_COMPONENTS / component
+        for component in pipeline_component_file_names
+        if "kafka-connector" not in component  # Shouldn't be used in the pipeline def
```
disrupted, line 246:
```suggestion
        if KafkaConnector.type not in component  # Shouldn't be used in the pipeline def
```

hooks/gen_docs.py
sujuka99: Empties the files

hooks/gen_docs.py
sujuka99: Yes, it currently supports only single inheritance. If it were to encounter a class with multiple parents, it just [picks the first one](https://note.nkmk.me/en/python-issubclass-mro-bases-subclasses/).

The thing is,  `inspect.getmro()` returns `component.__mro__` which is a tuple of `component` + all parent classes up to `object` in method resolution order.
IMO if we wanna support multiple inheritance, easier would be `component.__bases__` which returns a tuple of all immediate parents.

I have to see how to solve the dependencies in such a scenario.

hooks/gen_docs.py
sujuka99: The only drawback is we would have to maintain it manually. I think the dev docs are becoming more and more important.

hooks/gen_docs.py
sujuka99: I just needed a new name to do [this](https://github.com/bakdata/kpops/pull/267/files/d127dec5a420ff87cf3f3c9f0934773ab7519b87#diff-69ba40347f1f5ed3ff5374d0a0252fed381b703ca07b32ecd11089956c6ff4a5R79)

hooks/gen_docs.py
sujuka99: Good idea! I will have to erase the contents of `target` before the loop then.

hooks/gen_docs.py
```diff
@@ -1,10 +1,263 @@
...
+    sections_paths = [
+        PATH_DOCS_RESOURCES / "pipeline-components/headers" / component_file_name
+    ] + [PATH_DOCS_COMPONENTS / "sections" / section for section in component_sections]
+    concatenate_text_files(
+        *(defaults_sections_paths),
+        target=PATH_DOCS_RESOURCES / "pipeline-defaults/" / component_defaults_name,
+    )
+    concatenate_text_files(
+        *(sections_paths),
+        target=PATH_DOCS_RESOURCES / "pipeline-components/" / component_file_name,
```
disrupted, line 239: why the duplicate slash?

hooks/gen_docs.py
disrupted: ah, gotcha! you could also simply delete the file using `PATH.unlink()`

sujuka99: Nice! Would look cleaner, will do

hooks/gen_docs.py
disrupted: sorry, I still don't get it. If I remove this line and rename later uses of `component` to `component_name` it seems to work as well

hooks/gen_docs.py
sujuka99: typo

hooks/gen_docs.py
sujuka99: The idea is that in [this branch](https://github.com/bakdata/kpops/blob/ci/concatenate-examples/hooks/gen_docs.py#L78-L88) I reassign the component name to the parent while it is not `PipelineComponent` in search for the relevant section to inherit. I was worried that if I don't use a "burner" variable, the function would no longer be solving for the same component in further iterations.

sujuka99: If I assign a different value to `component_name`, would it not remain change for the rest of the function's life?

disrupted: but at this point the dict access in the surrounding for loop is already evaluated, no?

sujuka99: [Here](https://github.com/bakdata/kpops/blob/ci/concatenate-examples/hooks/gen_docs.py#L67C25-L67C25) I use it to assign a name to a file

sujuka99: I think I may have some general misunderstanding how variables behave in such loops haha. The whole idea is to reset the component name to its original value for each iteration.

hooks/gen_docs.py
sujuka99: Hmm, it seems I forgot to put the component name as a parameter, didn't even know this was valid haha

sujuka99:
> > TODO: Detect changes in the sections dir

> thinking out loud here: I am wondering whether we could use the actual config of the pre-commit hook for the change detection locally. And in the CI it could run and verify regardless if there were changes or not. wdyt? 

I suppose we could call the hook with certain args if certain files are changed, but I don't think `pre-commit` supports conditional args, so maybe we would need to have it twice in the config?

hooks/gen_docs.py
sujuka99: Did I do it the way you had in mind?

hooks/gen_docs.py
sujuka99: > nice use of the star operator for unpacking! 👍🏻 make `target` a kw-only argument, it's slightly safer :)

Thanks! It isn't allowed, the `*` arg makes it impossible to use `target` different from a keyword.
image

sujuka99: > 

done! reads much better

hooks/gen_docs.py
```diff
@@ -1,10 +1,277 @@
...
+                component = KPOPS_COMPONENTS_INHERITANCE_REF[component]
+                section = target_section + "-" + component + ".yaml"
+                if section in sections:
+                    component_sections.append(section)
+                    break
+                elif component == PipelineComponent.get_component_type():
+                    section = target_section + ".yaml"
+                    if section in sections:
+                        component_sections.append(section)
+                        break
```
disrupted, line 129:
```suggestion
            while (
                component := KPOPS_COMPONENTS_INHERITANCE_REF[component]
            ) != PipelineComponent.get_component_type():
                section = target_section + "-" + component_name
                if section in sections:
                    component_sections.append(section)
                    break
```

this seems to yield the same result for me, even though the `elif` block is missing. I am wondering what's its purpose (it never seems to get executed?)

hooks/gen_docs.py
```diff
@@ -1,10 +1,277 @@
...
+
+COMPONENTS_DEFINITION_SECTIONS = os.listdir(PATH_DOCS_COMPONENTS / "sections")
+PIPELINE_COMPONENT_FILE_NAMES = sorted(os.listdir(PATH_DOCS_COMPONENTS / "headers"))
+PIPELINE_COMPONENT_DEFAULTS = sorted(
+    os.listdir(PATH_DOCS_RESOURCES / "pipeline-defaults/headers")
+)
+
+
+class KpopsComponent(NamedTuple):
+    # name: str
```
disrupted, line 88: ?

hooks/gen_docs.py
disrupted: precisely! :)

hooks/gen_docs.py
```diff
@@ -1,10 +1,277 @@
...
+    :param key: A value to look for in the list of sections
+    :returns: The corresponding position or 0 if not found
+    """
+    for section in SECTIONS_ORDER:
+        if key.startswith((section + "-", section + ".")):
+            return SECTIONS_ORDER.index(section)
+    return 0
+
+
+def check_for_changes_in_kpops_component_structure() -> bool:
```
disrupted, line 160: do we still need this? Now that the pre-commit hook is triggered by specific files changing, we could just assume it's necessary to re-generate the docs, no?

hooks/gen_docs.py
```diff
@@ -1,10 +1,277 @@
...
+    try:
+        kpops_structure = load_yaml_file(PATH_DOCS_KPOPS_STRUCTURE)
+    except OSError:
+        kpops_structure = None
+    kpops_new_structure = {
+        "kpops_components_inheritance_ref": KPOPS_COMPONENTS_INHERITANCE_REF,
+        "kpops_components_fields": KPOPS_COMPONENTS_FIELDS,
+    }
+    if kpops_new_structure != kpops_structure:
+        Path(PATH_DOCS_COMPONENTS / "dependencies").mkdir(parents=True, exist_ok=True)
```
disrupted, line 178:
```suggestion
        (PATH_DOCS_COMPONENTS / "dependencies").mkdir(parents=True, exist_ok=True)
```
it's already a `Path` instance

hooks/__init__.py
```diff
@@ -1,4 +1,4 @@
 """KPOps pre-commit hooks"""
 from pathlib import Path
 
-PATH_ROOT = Path(__file__).parents[1]
+PATH_ROOT: Path = Path(__file__).parents[1]
```
disrupted, line 4: why is the annotation necessary?

hooks/gen_docs.py
sujuka99: This decides whether to regenerate the dependencies. If it returns `False`, the docs are generated using the dependency yamls. On `True` they are regenerated first. The intention is to save some time and resources.

hooks/gen_docs.py
sujuka99: Added `name` at first, but decided that it isn't needed afterwards.

hooks/gen_docs.py
sujuka99: worked without the `".yaml"` suffix to the section?? :0

It doesn't work for me. Note that this function is only used when `check_for_changes_in_kpops_component_structure()` notices a change.

hooks/__init__.py
sujuka99: Otherwise `pyright` wasn't recognizing that all the constants in `gem_docs.py` are of type `Path`

sujuka99:
@disrupted Is there a nice way to catch all exceptions in the examples-part of the script without indenting into a `try-except`?

disrupted:
> @disrupted Is there a nice way to catch all exceptions in the examples-part of the script without indenting into a `try-except`?

there is the `contextlib.suppress` but it's generally a common pattern in Python to use try-except if necessary. Which exceptions do you have to ignore and why?

sujuka99:
> > @disrupted Is there a nice way to catch all exceptions in the examples-part of the script without indenting into a `try-except`?
> 
> there is `contextlib.suppress` but it's generally a common pattern in Python to use try-except if necessary, it's part of the _better to ask for forgiveness than permission_ principle. Which exceptions do you have to ignore and why?

None, I solved it in a different way :) There was a problem if the user manually edits some of the dependency files

hooks/gen_docs.py
sujuka99: @disrupted Let me know if you still see a problem, I would be happy to simplify the function :)

Else, please mark as resolved

hooks/gen_docs.py
sujuka99: I looked and while it is certainly possible and not hard to implement, I would prefer not to for now. I cannot find a way to avoid adding even more nested loops in `filter_sections()` or `filter_section()` and I doubt that we would need such a feature soon. I don't want to add complexity and worsen readability.

If the need arises, I'd be happy to add support for multiple inheritance, but I suspect that by then we would have moved to examples generation without the need for manually-maintained yamls. At that point this script would need to be heavily modified.

.pre-commit-config.yaml
```diff
@@ -47,6 +47,14 @@ repos:
         name: gen-docs
         entry: python hooks/gen_docs.py
         language: system
-        types: [python]
+        files: > # contains `.gitignore` as an indicator whether the script is ran with `--all-files`
```
disrupted, line 50: clever, any side-effects if we purposefully make changes to the gitignore?

.pre-commit-config.yaml
```diff
@@ -47,6 +47,14 @@ repos:
...
+          (?x)^(
+            kpops/cli/main\.py
```
disrupted, line 52: shouldn't the docs be updated on any Python source file changing (except for tests), i.e. components?

.pre-commit-config.yaml
sujuka99: The warnings wouldn't appear, nothing more.

disrupted: great!

hooks/gen_docs.py
```diff
@@ -1,10 +1,307 @@
...
+        )
+        with PATH_DOCS_COMPONENTS_DEPENDENCIES.open("a") as f:
+            yaml.dump({component_file_name: component_sections}, f)
+        with PATH_DOCS_COMPONENTS_DEPENDENCIES_DEFAULTS.open("a") as f:
+            yaml.dump({component_file_name: component_sections_not_inheritted}, f)
+    else:
+        component_sections: list[str] = PIPELINE_COMPONENT_DEPENDENCIES[  # type: ignore [reportGeneralTypeIssues]
+            component_file_name
+        ]
+        component_sections_not_inheritted: list[
```
disrupted, line 238:
```suggestion
        component_sections_not_inherited: list[
```

hooks/gen_docs.py
disrupted: how about getting the fields in the order they are defined?

```python
KPOPS_COMPONENTS_FIELDS = {
    component.get_component_type(): [
        field_name
        for field_name, model in component.__fields__.items()
        if not model.field_info.exclude
    ]
    for component in KPOPS_COMPONENTS
}
SECTION_ORDER = deduplicate(
    [field for fields in KPOPS_COMPONENTS_FIELDS.values() for field in fields]
)
```

I am currently tweaking the order in #290

hooks/gen_docs.py
```diff
@@ -1,10 +1,317 @@
...
+#####################
+
+DANGEROUS_FILES_TO_CHANGE = {
+    PATH_DOCS_COMPONENTS_DEPENDENCIES,
+    PATH_DOCS_COMPONENTS_DEPENDENCIES_DEFAULTS,
+    PATH_DOCS_KPOPS_STRUCTURE,
+}
+# All args provided to the script
+# Pre-commit passes changed files as args
+SYS_ARGV = set(sys.argv)
```
disrupted, line 66:
```suggestion
CHANGED_FILES = set(sys.argv)
```
?

disrupted (changes requested):

hooks/gen_docs.py
sujuka99: I am not sure if that would be correct. Anything could be passed to the script as an arg and it would end up in that constant.

How about `SCRIPT_ARGUMENTS`?

.pre-commit-config.yaml
sujuka99: WDYT now?

hooks/gen_docs.py
sujuka99: Would be a great improvement :)

hooks/gen_docs.py
```diff
@@ -1,10 +1,308 @@
...
+    )
+except OSError:
+    is_change_present = True
+
+for component_file in PIPELINE_COMPONENT_HEADER_FILES:
+    component_file_name = component_file.name
+    component_type = component_file.stem
+    component_defaults_name = "defaults-" + component_file_name
+
+    component_sections, component_sections_not_inheritted = get_sections(
```
disrupted, line 264:
```suggestion
    component_sections, component_sections_not_inherited = get_sections(
```

hooks/gen_docs.py
```diff
@@ -1,10 +1,308 @@
...
+    concatenate_text_files(
+        *(sections_paths),
+        target=PATH_DOCS_RESOURCES / "pipeline-components" / component_file_name,
+    )
+concatenate_text_files(
+    *(
+        component_file.parents[1] / component_file.name
+        for component_file in PIPELINE_COMPONENT_HEADER_FILES
+        if KafkaConnector.get_component_type()
+        != component_file.stem  # Shouldn't be used in the pipeline def
```
disrupted, line 291: I would prefer to swap the operands in the condition, because it's the same way in other places aka reads nicer :)
```suggestion
        if component_file.stem
        != KafkaConnector.get_component_type() # shouldn't be used in the pipeline def
```

hooks/gen_docs.py
```diff
@@ -1,10 +1,308 @@
...
+KPOPS_COMPONENTS_SECTIONS = {
+    component.get_component_type(): [
+        field_name
+        for field_name, model in component.__fields__.items()
+        if not model.field_info.exclude
+    ]
+    for component in KPOPS_COMPONENTS
+}
+KPOPS_COMPONENTS_SECTIONS_ORDER = deduplicate(
+    [field for fields in KPOPS_COMPONENTS_SECTIONS.values() for field in fields]
```
disrupted, line 44: could you extract a `flatten` function from this list comprehension? it's a little hard to grasp

disrupted: it could live in the same location as the `deduplicate` function and ideally that's not the current location `base_defaults_component.py` 😄

disrupted (dismissed):

hooks/gen_docs.py
sujuka99: Sure

hooks/gen_docs.py
sujuka99: agreed, will do

hooks/gen_docs.py
sujuka99: ah, missed those, will fix

hooks/gen_docs.py
```diff
@@ -1,10 +1,316 @@
...
+                temp_component_name := KPOPS_COMPONENTS_INHERITANCE_REF[
+                    temp_component_name
+                ]
+            ) != PipelineComponent.get_component_type():
+                if section := filter_section(
+                    temp_component_name, sections, target_section
+                ):
+                    component_sections.append(section)
+                    break
+    component_sections.sort(key=component_section_position_in_definition)
```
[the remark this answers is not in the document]
sujuka99: Because this function is used to rewrite the yaml file.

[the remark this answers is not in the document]
sujuka99: I have to check whether I have to change the order in which the sections are collected tho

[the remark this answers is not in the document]
sujuka99: You are correct that we shouldn't have to sort the sections, `dict` is essentially equivalent to `OrderedDict` in modern Python.

I am pretty sure that I don't mess with the order of the sections anywhere in the script, so I'm removing the sorting func.

(I also tested it a bit and it seems to be working well without explicit sorting.)

hooks/gen_docs.py
sujuka99: I ended up removing it instead. The sort no longer made sense without any sets and with your reordering PR

hooks/gen_docs.py
raminqaf, line 1: Is it possible to split this file into smaller modules and call them in this file? It is getting bigger and bigger and harder to understand

raminqaf (changes requested):
It would be awesome to have a README where you **<i>briefly</i>** explain and define `defaults_pipeline_componenet_dependencies.yaml`, `kpops_structure.yaml`, `pipeline_component_dependencies.yaml`. It makes it easier to understand in the future. Besides, it is good to know when we need to update them whenever we introduce a new change. Later we can use this documentation for the contributor page.

hooks/gen_docs.py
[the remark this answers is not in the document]
irux: ok, yeah, I think remove it is better

sujuka99:
> It would be awesome to have a README where you **briefly** explain and define `defaults_pipeline_componenet_dependencies.yaml`, `kpops_structure.yaml`, `pipeline_component_dependencies.yaml`. It makes it easier to understand in the future. Besides, it is good to know when we need to update them whenever we introduce a new change. Later we can use this documentation for the contributor page.

That's a good idea, I will write it. Perhaps I could just start the dev docs with this? If I don't index the pages, no one would see them, so it doesn't matter if they are unfinished.

hooks/gen_docs.py
sujuka99: I agree that it's getting hard to grasp, I will break it down.

sujuka99: I am not sure how to break it down more, please take a look and let me know what you think.

disrupted (dismissed):
can we remove the v2 milestone from this PR?

docs/docs/resources/pipeline-components/sections/repo_config-kubernetes-app.yaml
```diff
@@ -0,0 +1,10 @@
+  # Helm repository configuration (optional)
+  # If not set the helm repo add will not be called. Useful when using local Helm charts
+  repoConfig:
+    repositoryName: bakdata-streams-bootstrap # required
```
raminqaf, line 4: Didn't we agree on snake_case for KPOps models?

disrupted: that's moved to v2

raminqaf (dismissed):

disrupted (approved):

PR #278, opened 2023-07-07
Plural broker field in pipeline config

raminqaf:


kpops/cli/pipeline_config.py
```diff
@@ -38,10 +38,11 @@ class PipelineConfig(BaseSettings):
...
         "Suffix your environment files with this value (e.g. defaults_development.yaml for environment=development).",
     )
-    broker: str = Field(
+    brokers: str = Field(
         default=...,
-        env=f"{ENV_PREFIX}KAFKA_BROKER",
-        description="The Kafka broker address.",
+        env=f"{ENV_PREFIX}KAFKA_BROKERS",
+        description="The comma separated Kafka brokers address.",
+        example="broker1:9092,broker2:9092,broker3:9092",
```
sujuka99, line 45: Awesome 🔥

sujuka99 (commented):
Why does that work?

kpops/cli/pipeline_config.py
raminqaf: This works because [streams-bootstrap](https://github.com/bakdata/streams-bootstrap/tree/master/charts/streams-app#streams) and the [Kafa Connect resetter](https://github.com/bakdata/kafka-connect-resetter#source-resetter) work in that way.

sujuka99: Ah, I assumed, but I see that they've had this feature for a while, right?

raminqaf: They had it from the beginning. KPOps only reads the string value and passes it downstream to streams-bootstrap and Kafka Connect resetter. They are handling the rest (for example, parsing the string with the comma into a list)

sujuka99: Why was it singular in KPOps before 🤔

raminqaf: I think back then, we just added it and didn't pay attention to it 🙃

sujuka99 (approved):
LGTM

PR #292, opened 2023-07-12
Validate unique step names

### LINKED ISSUE #134: KPOps should fail if there are two components with the same name
Currently, you can have two components with the same name although this should not be possible

disrupted:
Fixes #134


irux:
Please update :)

irux (approved):
LGTM!

raminqaf (approved):
🚀

PR #293, opened 2023-07-12
Exclude Helm tests from dry-run diff

### LINKED ISSUE #131: Exclude Helm tests from dry-run diff
Currently, [Helm tests](https://helm.sh/docs/topics/chart_tests/) are rendered in the diff as additions. They should be ignored instead

raminqaf:
Fixes #131

### DIFF
--- a/tests/component_handlers/helm_wrapper/test_helm_wrapper.py
+++ b/tests/component_handlers/helm_wrapper/test_helm_wrapper.py
@@ -320,10 +320,9 @@ foo: bar
         stdout = """---
             # Resource: chart/templates/test1.yaml
             """
-        with pytest.raises(ParseError) as parse_error:
+        with pytest.raises(ParseError, match="Not a valid Helm template source"):
             helm_template = list(Helm.load_manifest(stdout))
             assert len(helm_template) == 0
-            assert str(parse_error.value) == "Not a valid Helm template source"
 
     def test_load_manifest(self):
 
@@ -377,6 +376,7 @@ NOTES:
             isinstance(helm_template, HelmTemplate) for helm_template in helm_templates
         )
         assert helm_templates[0].filepath == "chart/templates/test3a.yaml"
+        assert helm_templates[0].template == {"data": [{"a": 1}, {"b": 2}]}
         assert helm_templates[1].filepath == "chart/templates/test3b.yaml"
         assert helm_templates[1].template == {"foo": "bar"}
 
--- a/tests/component_handlers/helm_wrapper/test_helm_wrapper.py
+++ b/tests/component_handlers/helm_wrapper/test_helm_wrapper.py
@@ -380,11 +380,11 @@ NOTES:
         assert helm_templates[1].filepath == "chart/templates/test3b.yaml"
         assert helm_templates[1].template == {"foo": "bar"}
 
-    def test_get_manifest(self, run_command: MagicMock, mock_get_version: MagicMock):
+    def test_helm_get_manifest(
+        self, run_command: MagicMock, mock_get_version: MagicMock
+    ):
         helm_wrapper = Helm(helm_config=HelmConfig())
-        run_command.return_value = """Release "test-release" has been upgraded. Happy Helming!
-NAME: test-release
-MANIFEST:
+        run_command.return_value = """
 ---
 # Source: chart/templates/test.yaml
 data:
--- a/tests/component_handlers/helm_wrapper/test_helm_wrapper.py
+++ b/tests/component_handlers/helm_wrapper/test_helm_wrapper.py
@@ -320,9 +320,10 @@ foo: bar
         stdout = """---
             # Resource: chart/templates/test1.yaml
             """
-        with pytest.raises(ParseError) as value_error:
-            list(Helm.load_manifest(stdout))
-        assert str(value_error.value) == f"Failed to parse Helm stdout:\n {stdout}"
+        with pytest.raises(ParseError) as parse_error:
+            helm_template = list(Helm.load_manifest(stdout))
+            assert len(helm_template) == 0
+            assert str(parse_error.value) == "Not a valid Helm template source"
 
     def test_load_manifest(self):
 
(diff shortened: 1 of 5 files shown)


disrupted (changes requested):

tests/components/test_kubernetes_app.py
```diff
@@ -69,6 +69,7 @@ def test_should_lazy_load_helm_wrapper_and_not_repo_add(
         mocker: MockerFixture,
         helm_mock: MagicMock,
         app_value: KubernetesTestValue,
+        dry_run_handler: MagicMock,
```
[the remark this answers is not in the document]
raminqaf: It is needed in this case because the dry-run is set to try when calling the `deploy`, and the dry-run handler is executed. When passing the MagicMock, the actual execution is skipped. I can add a validation `assert_called_once`.
Another way would be to remove it and call deploy with dry-run Flase.

tests/component_handlers/helm_wrapper/test_helm_wrapper.py
```diff
@@ -320,9 +320,10 @@ def test_raise_parse_error_when_helm_content_is_invalid(self):
         stdout = """---
             # Resource: chart/templates/test1.yaml
             """
-        with pytest.raises(ParseError) as value_error:
-            list(Helm.load_manifest(stdout))
-        assert str(value_error.value) == f"Failed to parse Helm stdout:\n {stdout}"
+        with pytest.raises(ParseError) as parse_error:
+            helm_template = list(Helm.load_manifest(stdout))
+            assert len(helm_template) == 0
+            assert str(parse_error.value) == "Not a valid Helm template source"
```
disrupted, line 326: assertion needs to be outside the context manager scope. this is a common mistake we used to make throughout the tests. see https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

disrupted: or you could rewrite it as
```python
with raises(ParseError, match="Not a valid Helm template source"):
    ...
```

tests/component_handlers/helm_wrapper/test_helm_wrapper.py
```diff
@@ -362,14 +377,14 @@ def test_load_manifest(self):
             isinstance(helm_template, HelmTemplate) for helm_template in helm_templates
         )
         assert helm_templates[0].filepath == "chart/templates/test3a.yaml"
-        assert helm_templates[0].template == {"data": [{"a": 1}, {"b": 2}]}
```
disrupted, line 365: why did you delete this line?

tests/components/test_kubernetes_app.py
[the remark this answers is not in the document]
raminqaf: IMO the correct way would be to split this into two tests. Once with dry_run true and once with false. But this is outside of the scope of this PR. So maybe setting it to false is the better approach for now

tests/component_handlers/helm_wrapper/test_helm_wrapper.py
raminqaf: ups

tests/components/test_kubernetes_app.py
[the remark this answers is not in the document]
raminqaf: Changed

tests/component_handlers/helm_wrapper/test_helm_wrapper.py
```diff
@@ -362,14 +377,14 @@ def test_load_manifest(self):
...
         assert helm_templates[1].filepath == "chart/templates/test3b.yaml"
         assert helm_templates[1].template == {"foo": "bar"}
 
     def test_get_manifest(self, run_command: MagicMock, mock_get_version: MagicMock):
         helm_wrapper = Helm(helm_config=HelmConfig())
         run_command.return_value = """Release "test-release" has been upgraded. Happy Helming!
 NAME: test-release
+MANIFEST:
```
disrupted, line 387: why did you add this line, is it guaranteed that the Helm template always includes this line?
otherwise, I'd say we should leave the test as-is.

raminqaf: I added this because this message represents the stdout of `helm upgrade --install,` and this command outputs these headers, HOOKS, MANIFEST, and NOTES (optional).
On the other hand, the command `helm get manifest` only prints the manifests without any header. That is why I added the check if the `MANIFEST:` header exists, otherwise the start should be `---`

raminqaf: Actually this is the output of `helm get manifest` I'll correct the string :D

disrupted: I would say revert this line or even remove the whole header if that's the output format of `helm get manifest`

tests/component_handlers/helm_wrapper/test_helm_wrapper.py
```diff
@@ -278,14 +279,17 @@ def test_validate_console_output(self):
...
     def test_helm_template_load(self):
-        stdout = """---
-# Source: chart/templates/test2.yaml
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-    labels:
-        foo: bar
-"""
+        stdout = dedent(
```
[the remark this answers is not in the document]
raminqaf: These disrupted multi-line strings always annoyed me! But now it is clean 🧼

tests/component_handlers/helm_wrapper/test_helm_wrapper.py
```diff
@@ -331,6 +316,16 @@ def test_load_manifest(self):
...
+    def test_raise_parse_error_when_helm_content_is_invalid(self):
+        stdout = """---
+            # Resource: chart/templates/test1.yaml
+            """
+        with pytest.raises(ParseError, match="Not a valid Helm template source"):
+            helm_template = list(Helm.load_manifest(stdout))
+            assert len(helm_template) == 0
+
+    def test_load_manifest(self):
+
```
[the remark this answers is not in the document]
disrupted: #294

PR #296, opened 2023-07-13
Camel case extra fields in app section

disrupted:
defining Helm values in snake case should work automatically and yield the same result as using camel case. e.g. `image_tag` → `imageTag`

### DIFF
--- a/examples/bakdata/atm-fraud-detection/defaults.yaml
+++ b/examples/bakdata/atm-fraud-detection/defaults.yaml
@@ -3,6 +3,9 @@ pipeline-component:
 
 kubernetes-app:
   namespace: ${NAMESPACE}
+  app:
+    image_tag: "1.0.0"
+    replica_count: 1
 
 kafka-connector:
   namespace: ${NAMESPACE}
@@ -13,8 +16,14 @@ kafka-app:
       brokers: ${brokers}
       schemaRegistryUrl: ${schema_registry_url}
       optimizeLeaveGroupBehavior: false
+    prometheus:
+      jmx:
+        enabled: false
+    debug: true
 
 producer:
+  app:
+    suspend: true
   to:
     topics:
       ${output_topic_name}:
@@ -25,8 +34,8 @@ streams-app:
   app:
     labels:
       pipeline: ${pipeline_name}
-    streams:
-      optimizeLeaveGroupBehavior: false
+    command_line:
+      PRODUCTIVE: false
   to:
     topics:
       ${error_topic_name}:
--- a/examples/bakdata/atm-fraud-detection/defaults.yaml
+++ b/examples/bakdata/atm-fraud-detection/defaults.yaml
@@ -14,8 +14,8 @@ kafka-app:
   app:
     streams:
       brokers: ${brokers}
-      schemaRegistryUrl: ${schema_registry_url}
-      optimizeLeaveGroupBehavior: false
+      schema_registry_url: ${schema_registry_url}
+      optimize_leave_group_behavior: false
     prometheus:
       jmx:
         enabled: false
--- a/examples/bakdata/atm-fraud-detection/pipeline.yaml
+++ b/examples/bakdata/atm-fraud-detection/pipeline.yaml
@@ -1,9 +1,9 @@
 - type: producer
   name: account-producer
   app:
-    replicaCount: 1
+    replica_count: 1
     image: ${DOCKER_REGISTRY}/atm-demo-accountproducer
-    imageTag: 1.0.0
+    image_tag: 1.0.0
     schedule: 0 12 * * *
     suspend: true
     prometheus:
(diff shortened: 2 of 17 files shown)


irux:
So, as I understand. This PR allow to use camel case or snake case interchangeably, right ?

disrupted:
> So, as I understand. This PR allow to use camel case or snake case interchangeably, right ? 

that's right!

irux:
should you maybe have a test where you mix both cases and everything works as expected ?

examples/bakdata/atm-fraud-detection/defaults.yaml
```diff
@@ -25,8 +34,8 @@ streams-app:
   app:
     labels:
       pipeline: ${pipeline_name}
-    streams:
-      optimizeLeaveGroupBehavior: false
+    command_line:
```
raminqaf, line 37:
```suggestion
    commandLine:
```

examples/bakdata/atm-fraud-detection/pipeline.yaml
```diff
@@ -1,61 +1,31 @@
...
-    debug: true
 
 - type: producer
   name: transaction-avro-producer
   app:
-    replicaCount: 1
     image: ${DOCKER_REGISTRY}/atm-demo-transactionavroproducer
-    imageTag: 1.0.0
-    commandLine:
+    command_line:
```
raminqaf, line 11:
```suggestion
    commandLine:
```

disrupted:
> should you maybe have a test where you mix both cases and everything works as expected ?

good idea

examples/bakdata/atm-fraud-detection/defaults.yaml
```diff
@@ -3,6 +3,9 @@ pipeline-component:
 
 kubernetes-app:
   namespace: ${NAMESPACE}
+  app:
+    image_tag: "1.0.0"
+    replica_count: 1
```
raminqaf, line 8:
```suggestion
    imageTag: "1.0.0"
    replicaCount: 1
```

kpops/components/base_components/models/to_section.py
```diff
@@ -36,26 +38,36 @@ class TopicConfig(BaseModel):
     :type role: str | None
     """
 
-    type: OutputTopicTypes = Field(..., description="Topic type")
+    type: OutputTopicTypes = Field(..., description=describe_attr("type", __doc__))
     key_schema: str | None = Field(
-        default=None, alias="keySchema", description="Key schema class name"
+        default=None,
+        title="Key schema",
```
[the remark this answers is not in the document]
disrupted: it's for the schema

examples/bakdata/atm-fraud-detection/defaults.yaml
disrupted: both works :)

examples/bakdata/atm-fraud-detection/defaults.yaml
disrupted: why do you want me to revert it?

examples/bakdata/atm-fraud-detection/defaults.yaml
raminqaf: I know :) I was referring to the conversation we had. Especially for the examples and docs, we should stick to only one naming convention

examples/bakdata/atm-fraud-detection/defaults.yaml
raminqaf: For the examples, I would stick to only one naming convention. Since there is camelCase here already I would stick to camel case.

disrupted: ah ok, there are only two occurrences of camel case left in this example. would it be ok if I change these to snake case as well? 😄 or does it have to be camel case?

raminqaf: I am totally fine with the snake case but also pay attention to the other files here:
https://github.com/bakdata/kpops/blob/0b882c7acddacfe2ce037bf7563183d5af3ead28/examples/bakdata/atm-fraud-detection/pipeline.yaml#L42

raminqaf:
Should we add to the documentation that we are supporting both (camel and snake) cases?

examples/bakdata/atm-fraud-detection/defaults.yaml
disrupted: that's an annotation, they don't adhere to any convention 🤷🏻‍♂️

disrupted: that's why we can't convert it

raminqaf: ah damn 😭 didn't see that.

disrupted:
> Should we add to the documentation that we are supporting both (camel and snake) cases?

imo we could at least mention it, I just don't know where it would make sense. pinging our docs hero @sujuka99 🧙🏻 – what's your opinion?

PR #306, opened 2023-07-31
Refactor CLI using dtyper

disrupted:
Typer doesn't allow calling commands from outside its context. `dtyper` acts as a slim wrapper to patch this drawback, making it possible to import the commands from an external Python module. This is the first step towards providing a developer API that allows to hook into the kpops operations.

### DIFF
--- a/kpops/cli/main.py
+++ b/kpops/cli/main.py
@@ -163,7 +163,9 @@ def get_steps_to_apply(
     return list(pipeline)
 
 
-def reverse_pipeline_steps(pipeline, steps) -> Iterator[PipelineComponent]:
+def reverse_pipeline_steps(
+    pipeline: Pipeline, steps: str | None
+) -> Iterator[PipelineComponent]:
     return reversed(get_steps_to_apply(pipeline, steps))
 
 
--- a/kpops/cli/main.py
+++ b/kpops/cli/main.py
@@ -4,6 +4,7 @@ import logging
 from pathlib import Path
 from typing import TYPE_CHECKING, Iterator, Optional
 
+import dtyper
 import typer
 
 from kpops import __version__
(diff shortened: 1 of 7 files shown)


kpops/cli/main.py
```diff
@@ -185,7 +190,7 @@ def create_pipeline_config(
     return pipeline_config
 
 
-@app.command(
+@app.command(  # pyright: ignore[reportGeneralTypeIssues]
```
disrupted, line 193: opened an issue https://github.com/rec/dtyper/issues/8

irux: I think it would be better to put the link on the code too :)

tests/pipeline/test_pipeline.py
```diff
@@ -5,23 +5,33 @@
...
+        assert len(pipeline) == 3
+
     def test_load_pipeline(self, snapshot: SnapshotTest):
         result = runner.invoke(
             app,
             [
                 "generate",
                 "--pipeline-base-dir",
-                PIPELINE_BASE_DIR,
+                str(PIPELINE_BASE_DIR_PATH),
```
irux, line 34: Any reason to move the str to the reference and remove it from the definition ?

disrupted: for the new test which needs a `Path` object. besides, it looks a little more uniform with the rest of the arguments

tests/pipeline/test_pipeline.py
```diff
@@ -5,23 +5,33 @@
 from snapshottest.module import SnapshotTest
 from typer.testing import CliRunner
 
+import kpops
 from kpops.cli.main import app
 from kpops.pipeline_generator.pipeline import ParsingException
 
 runner = CliRunner()
 
 RESOURCE_PATH = Path(__file__).parent / "resources"
```
irux, line 14: os.join or use Path for joining it already ?

raminqaf:
Closing PR because this can now be achieved through https://github.com/bakdata/kpops/pull/306 and doesn't need to be natively built in KPOps

tests/pipeline/test_pipeline.py
disrupted: what do you mean? the `/` operator works the same as `Path(...).join(...)`

raminqaf (approved):

irux (approved):

PR #307, opened 2023-08-01
Automatically support schema generation for custom components

disrupted:
manually defining component `schema_type` is no longer necessary

component type can be set without using `pydantic.Field`
```python
class CustomComponent(PipelineComponent):
    type = "custom-component"
```

### DIFF
--- a/kpops/utils/gen_schema.py
+++ b/kpops/utils/gen_schema.py
@@ -1,6 +1,6 @@
 import logging
 from enum import Enum
-from typing import Annotated, Any, Literal, Sequence, Union, get_args, get_origin
+from typing import Annotated, Any, Literal, Sequence, Union
 
 from pydantic import Field, schema, schema_json_of
 from pydantic.fields import ModelField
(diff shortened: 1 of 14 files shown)


tests/cli/test_schema_generation.py
disrupted, line 1: this might need some updated test cases

kpops/utils/gen_schema.py
```diff
@@ -96,9 +75,9 @@ def _add_components(
         components = tuple()
     # Set of existing types, against which to check the new ones
     defined_component_types: set[str] = {
-        component.__fields__["schema_type"].default
+        component.__fields__["type"].default
         for component in components
-        if component.__fields__.get("schema_type")
+        if component.__fields__.get("type")
```
sujuka99: Would using the walrus operator be an improvement here?

```
defined_component_types: set[str] = {
    component_type.default
    for component in components
    if (component_type := component.__fields__.get("type"))
}

Comment (disrupted) [reply kpops/utils/gen_schema.py]: absolutely, good point!

Comment (disrupted) [reply kpops/utils/gen_schema.py]: actually it can just be
```python
defined_component_types = {
    component.get_component_type() for component in components
}
```
at this point every component has to have a type defined

tests/cli/test_schema_generation.py
disrupted: seems fine with the updated test cases

sujuka99 (dismissed):
LGTM

sujuka99 (approved):

Issue #66, opened 2023-01-18
Camel case all the component, config, and default variables


raminqaf:
Currently, some of the variables like the `TopicSpec` are not camel-cased when we write the `pipeline.yaml` file:
```python
class TopicSpec(BaseModel):
    topic_name: str
    partitions_count: int | None
    replication_factor: int | None
    configs: list[dict[str, str]] | None

    class ModelConfigs(BaseConfig):
        extra = Extra.forbid
```
Also, the fields in the [PipelineConfig](https://github.com/bakdata/kpops/blob/main/kpops/cli/pipeline_config.py) class are not camel-cased. We should identify all these variables and standardize them.

disrupted:
since this is a Python project, we should consider using snake case for everything instead of camel case. hot take but imo it's also easier to read

disrupted:
yeah, I believe we're already consistent with snake_case. In addition, some models allow for camelCase aliases for easier copy-paste workflows. IMO this can be closed.

raminqaf:
Closing this ticket

PR #308, opened 2023-08-02
Remove camel case conversion of internal models

disrupted:
Related to #66 
only Helm/Kubernetes values require camel casing

### DIFF
--- a/docs/docs/resources/pipeline-components/kafka-source-connector.yaml
+++ b/docs/docs/resources/pipeline-components/kafka-source-connector.yaml
@@ -14,22 +14,22 @@
       ${pipeline_name}-output-topic:
         type: output # required
         # role: topic-role # only used if type is `extra`
-        keySchema: key-schema # must implement SchemaProvider to use
-        valueSchema: value-schema
+        key_schema: key-schema # must implement SchemaProvider to use
+        value_schema: value-schema
         partitions_count: 1
         replication_factor: 1
         configs: # https://kafka.apache.org/documentation/#topicconfigs
           cleanup.policy: compact
     models: # SchemaProvider is initiated with the values given here
       model: model
-  # Pipeline prefix that will prefix every component name. If you wish to not 
+  # Pipeline prefix that will prefix every component name. If you wish to not
   # have any prefix you can specify an empty string.
   prefix: ${pipeline_name}-
   # Helm repository configuration
   repoConfig:
     repositoryName: my-repo # required
     url: https://bakdata.github.io/kafka-connect-resetter/ # required
-    repoAuthFlags: 
+    repoAuthFlags:
       username: user
       password: pass
       ca_file: /home/user/path/to/ca-file
--- a/docs/docs/resources/pipeline-components/kafka-source-connector.yaml
+++ b/docs/docs/resources/pipeline-components/kafka-source-connector.yaml
@@ -26,10 +26,10 @@
   # have any prefix you can specify an empty string.
   prefix: ${pipeline_name}-
   # Helm repository configuration
-  repoConfig:
-    repositoryName: my-repo # required
+  repo_config:
+    repository_name: my-repo # required
     url: https://bakdata.github.io/kafka-connect-resetter/ # required
-    repoAuthFlags:
+    repo_auth_flags:
       username: user
       password: pass
       ca_file: /home/user/path/to/ca-file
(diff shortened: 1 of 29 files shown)


raminqaf:
Should we also update it in the docs pipelines?
https://github.com/search?q=repo%3Abakdata%2Fkpops+%28partitions_count+OR+replication_factor%29+language%3AYAML&type=code

raminqaf:
For case consistency, it would make sense to add the `CamelCaseConfig` to the `PipelineConfig` too. Then the KPOps variables are also camel cased.
https://github.com/bakdata/kpops/blob/main/kpops/cli/pipeline_config.py

kpops/component_handlers/topic/model.py
```diff
@@ -29,9 +26,6 @@ class TopicResponse(BaseModel):
         # https://github.com/confluentinc/kafka-rest/issues/1085
     )
 
-    class ModelConfigs(BaseConfig):
-        extra = Extra.forbid
-
```
disrupted, line 34: this never worked because the class was named correctly (should have been `Config`). when fixing the mistake the dry-run fails, because the model is incomplete, e.g. missing `authorized_operations`. I believe we can just ignore the extra fields which is the default behavior anyways

docs/docs/resources/pipeline-components/kafka-source-connector.yaml
```diff
@@ -14,31 +14,31 @@
...
   version: "1.0.4" # Helm chart version
   # Overriding Kafka Connect Resetter Helm values. E.g. to override the
   # Image Tag etc.
-  resetterValues:
+  resetter_values:
     imageTag: "1.2.3"
   # offset.storage.topic
   # https://kafka.apache.org/documentation/#connect_running
-  offsetTopic: offset_topic
+  offset_topic: offset_topic
```
raminqaf, line 44: If this is part of the Helm values, please keep the camelCase format.
https://github.com/bakdata/kafka-connect-resetter/blob/fc838e560d6b29302ed1e04c0b59748ea109d0f2/charts/kafka-connect-resetter/values.yaml#L23

disrupted: hm, I wonder why this is not part of the `resetter_values` then?

disrupted: at least it's not part of the Helm values in this context

raminqaf: Shouldn't this be formatted like this:
```yaml
  resetter_values:
    imageTag: "1.2.3"
    config:
        offsetTopic: offset_topic
```
I found this `offset_topic`: https://github.com/bakdata/kpops/blob/85d4d048946753dabb2e5aa17052aafcb6d9b7e2/kpops/components/base_components/kafka_connector.py#L322
I am not quite sure if this is needed 🤔 
Becuase we have the one that represents the helm value here: https://github.com/bakdata/kpops/blob/85d4d048946753dabb2e5aa17052aafcb6d9b7e2/kpops/component_handlers/kafka_connect/model.py#L63

disrupted: this is assigned automatically by KPOps though: https://github.com/bakdata/kpops/blob/599980add1eb8237b36b0a5b7102075ed7ebbe09/kpops/components/base_components/kafka_connector.py#L366

raminqaf: So now I can write something like this. Correct? 
```yaml
  resetter_values:
    imageTag: "1.2.3"
    config:
        offsetTopic: offset_topic
  offset_topic: offset_topic <--------- this seems redundant
```
Wouldn't it make sense to have only one definition?

disrupted: first and foremost the offset topic is a Source connector config actually, see `offset.storage.topic` https://kafka.apache.org/documentation/#connectconfigs_offset.storage.topic
as such, it could be written as

```yaml
- type: kafka-source-connector
  name: example
  app:
    "offset.storage.topic": offsets
```

I guess we made it a top-level attribute to make it more clear that this is required for source connectors only. If we decide to change something about it, I think it should be a separate issue/PR

sujuka99 (approved):
LGTM!

PR #309, opened 2023-08-03
Derive component type automatically from class name

disrupted:
Depends on #307

### DIFF
--- a/tests/cli/snapshots/snap_test_schema_generation.py
+++ b/tests/cli/snapshots/snap_test_schema_generation.py
@@ -360,19 +360,19 @@ snapshots['TestGenSchema.test_gen_pipeline_schema_only_custom_module test-schema
                     "title": "Configs",
                     "type": "object"
                 },
-                "keySchema": {
+                "key_schema": {
                     "description": "Key schema class name",
-                    "title": "Keyschema",
+                    "title": "Key schema",
                     "type": "string"
                 },
                 "partitions_count": {
                     "description": "Number of partitions into which the topic is divided",
-                    "title": "Partitions Count",
+                    "title": "Partitions count",
                     "type": "integer"
                 },
                 "replication_factor": {
-                    "description": "Replication topic of the topic",
-                    "title": "Replication Factor",
+                    "description": "Replication factor of the topic",
+                    "title": "Replication factor",
                     "type": "integer"
                 },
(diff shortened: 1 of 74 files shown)


sujuka99 (changes requested):
The reference docs need an update #310

tests/cli/snapshots/snap_test_schema_generation.py
```diff
@@ -259,7 +259,7 @@
                     "enum": [
                         "sub-pipeline-component-correct-docstr"
                     ],
-                    "title": "Parameter description looks correct and it is not included in the class description, terminates here",
```
sujuka99, line 262: Could you keep this check somehow? It is not automated enough, but I'd prefer to have it in some form at least.

[Related change](https://github.com/bakdata/kpops/pull/309/files#diff-f98effd3c9895a821b65f6c97e709d3e9114cf52d41c454c5a87364187fe35dfL52-L59)

tests/cli/test_schema_generation.py
```diff
@@ -49,19 +49,9 @@ class SubPipelineComponentCorrectDocstr(SubPipelineComponent):
 
     ,
 
-    :param type: Parameter description looks correct and it is not included in
-        the class description,
-        terminates here ,
-        defaults to anything really, this here should not be included as it follows
```
sujuka99, line 55: See above

tests/utils/test_pydantic.py
sujuka99, line 1: Cool! I should add tests for the util functions here.

sujuka99 (changes requested):

raminqaf (dismissed):
LGTM

tests/cli/snapshots/snap_test_schema_generation.py
disrupted: you're right, good thing you noticed. I've added an example attribute back to keep this check

tests/utils/test_pydantic.py
disrupted: I agree we could add tests for `describe_object` and `describe_attr` here, but this would be better suited to a follow-up PR, what do you think?

sujuka99 (commented):
LGTM

sujuka99 (approved):

PR #312, opened 2023-08-08
Add support for pipeline steps parallelization 

### LINKED ISSUE #177: Parallelize component steps execution
Imagine a pipeline.yaml file contains 30+ Kafka sink connectors. If you run the `kpops clean` command, KPOps goes through each step and, for the sink connectors, deploys a Kafka connect resetter. On every resetter deployment, KPOps waits until the job is finished and then deploys the other resetter. This takes a lot of time to finish. We can (and should) parallelize the execution of this type of execution.
The first step would be to detect whether the parallelization of the steps would affect the next steps. To do so, we must create an execution plan through a DAG. Having a DAG will help us parallelize the execution.

## Plan

- [x] #282 
- [x] #283
- [x] #284 
- [x] #285

### LINKED ISSUE #283: Refactor builtin components for AsyncIO
- KPOps operations (deploy, destroy, reset, clean) should be async functions, rather than synchronous
- check existing dependencies for AsyncIO support
- dependencies that do not support AsyncIO natively should be wrapped in thread executor

irux:
closes #177


kpops/component_handlers/kafka_connect/kafka_connect_handler.py
```diff
@@ -44,57 +44,77 @@ def create_connector(
...
                     secs=self._timeout,
                 )
 
-                timeout(
-                    lambda: self._connect_wrapper.delete_connector(connector_name),
+                async def delete_connector_locally():
+                    await self._connect_wrapper.delete_connector(connector_name)
+
+                await timeout(
+                    delete_connector_locally(),
```
disrupted, line 104: why does it need to be wrapped in another async function, can't you pass `self._connect_wrapper.delete_connector(connector_name)`?

disrupted (changes requested):

kpops/components/base_components/pipeline_component.py
```diff
@@ -67,6 +71,25 @@ def __init__(self, **kwargs) -> None:
...
+    def get_extra_input_topics(self) -> dict[str, list[str]]:
+        """Get extra input topics list from config."""
+        return {}
+
+    def get_output_topic(self) -> str | None:
+        """Get output topic from config."""
+
+    def get_extra_output_topics(self) -> dict[str, str]:
+        """Get extra output topics list from config."""
+        return {}
```
disrupted, line 87: how about we make these properties?

tests/pipeline/resources/pipelines-with-graphs/simple-pipeline/pipeline.yaml
```diff
@@ -0,0 +1,13 @@
+- type: producer-app
+  name: app1
+  app:
+    resources:
+      limits:
+        memory: 2G
+      requests:
+        memory: 2G
```
disrupted, line 8: what is this needed for?

kpops/components/base_components/pipeline_component.py
irux: But this should be possible to override. Maybe is not clear because it doesn't have the override keyword. Should I put it ?

disrupted: `@property` can be overridden, too

kpops/components/base_components/pipeline_component.py
```diff
@@ -67,6 +71,30 @@ def __init__(self, **kwargs) -> None:
...
+        return None
+
+    @property
+    def extra_output_topics(self) -> dict[str, str]:
+        """Get extra output topics list from config."""
+        return {}
+
+    @property
+    def id(self) -> str:
+        return f"component-{self.full_name}"
```
disrupted, line 96: do we still need the prefix?

kpops/components/streams_bootstrap/producer/producer_app.py
```diff
@@ -68,6 +68,14 @@ def apply_to_outputs(self, name: str, topic: TopicConfig) -> None:
             case _:
                 super().apply_to_outputs(name, topic)
 
+    @property
```
disrupted, line 71: please mark as override

disrupted (changes requested):

kpops/components/streams_bootstrap/producer/producer_app.py
irux: I needed to remove it because pyright doesn't want to have it.

disrupted: did you set it in the correct order?
```python
@property
@override
def ...
```

tests/component_handlers/kafka_connect/test_connect_wrapper.py
```diff
@@ -425,18 +459,22 @@ def test_should_raise_rebalance_in_progress_when_deleting_connector(
             status_code=409,
         )
 
-        timeout(
-            lambda: self.connect_wrapper.delete_connector(connector_name),
+        async def delete_connector_locally():
+            await self.connect_wrapper.delete_connector(connector_name)
+
+        await timeout(
+            delete_connector_locally(),
```
disrupted, line 466: wrapper necessary?

disrupted (changes requested):

kpops/components/streams_bootstrap/producer/producer_app.py
irux: this should be all. I hope!

disrupted (changes requested):
please always label your PRs.

Also add a section about the new feature being in preview and not ready for production use in the docs.

kpops/components/streams_bootstrap/producer/producer_app.py
irux: well it seems that regardless of the order, pyright doesn't like it

irux: should I remove it ?

disrupted: these are not in the correct order https://github.com/bakdata/kpops/pull/312/commits/58cc85ffc4c4837479f72698bfcaab051d2d9a4f

irux:
> Also add a section about the new feature being in preview and not ready for production use in the docs.

Where do you think I should put it ?

kpops/components/streams_bootstrap/producer/producer_app.py
irux: https://github.com/bakdata/kpops/actions/runs/7623207572/job/20762773438#step:8:20 the order is still not the problem https://github.com/bakdata/kpops/blob/feature/parallelization/kpops/components/base_components/pipeline_component.py#L75

kpops/components/base_components/pipeline_component.py
irux: yes! this is because topic and component can have the same name.

kpops/components/streams_bootstrap/producer/producer_app.py
disrupted: ah, check the import, there is a mistake. right now it only gets imported in the except clause

disrupted:
> > Also add a section about the new feature being in preview and not ready for production use in the docs.
> 
> Where do you think I should put it ?

good point, I am not entirely sure. There's no dedicated section I can think of, but perhaps add it to the `kpops --help`? what do you think?

disrupted (changes requested):

tests/components/test_producer_app.py
```diff
@@ -274,3 +278,37 @@ def test_should_clean_producer_app_and_deploy_clean_up_job_and_delete_clean_up_w
...
+                        ),
+                    }
+                },
+            },
+        )
+        assert producer_app.output_topic == "${output_topic_name}"
+        assert producer_app.extra_output_topics == {
+            "first-extra-topic": "extra-topic-1"
+        }
+        assert producer_app.input_topics == []
```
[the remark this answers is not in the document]
irux: I just added the assertion on their respective tests.

irux:
> > > Also add a section about the new feature being in preview and not ready for production use in the docs.
> > 
> > 
> > Where do you think I should put it ?
> 
> good point, I am not entirely sure. There's no dedicated section I can think of, but perhaps add it to the `kpops --help`? what do you think?

but isn't already there?

disrupted:
> > > > Also add a section about the new feature being in preview and not ready for production use in the docs.
> > > 
> > > 
> > > Where do you think I should put it ?
> > 
> > 
> > good point, I am not entirely sure. There's no dedicated section I can think of, but perhaps add it to the `kpops --help`? what do you think?
> 
> but isn't already there?

where do you see that?
To me it only shows


* `--parallel / --no-parallel`: Enable or disable parallel execution of pipeline steps. If enabled, multiple steps can be processed concurrently. If disabled, steps will be processed sequentially.  [default: no-parallel]

tests/components/test_producer_app.py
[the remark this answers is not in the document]
disrupted: for the producer, the assertion for `inputs` is missing
similarly, streams-app should have an assertion for `outputs`

disrupted (approved):
LGTM

PR #313, opened 2023-08-08
Set default for ToSection topics

disrupted:
allow configuring empty ToSection more easily without having to specify topics attribute, same as FromSection

```yaml
- type: kubernetes-app
  name: example
  from: {}
  to: {}
```

### DIFF
--- a/kpops/components/base_components/models/to_section.py
+++ b/kpops/components/base_components/models/to_section.py
@@ -79,10 +79,8 @@ class TopicConfig(BaseModel):
 class ToSection(BaseModel):
     """Holds multiple output topics
 
-    :param models: Data models
-    :type models: dict[str, Any]
     :param topics: Output topics
-    :type topics: dict[str, TopicConfig]
+    :param models: Data models
     """
 
     # TODO: really multiple models?
--- a/kpops/components/base_components/models/to_section.py
+++ b/kpops/components/base_components/models/to_section.py
@@ -87,11 +87,11 @@ class ToSection(BaseModel):
 
     # TODO: really multiple models?
     # any because snapshot versions must be supported
+    topics: dict[TopicName, TopicConfig] = Field(
+        default={}, description=describe_attr("topics", __doc__)
+    )
     models: dict[str, Any] = Field(
         default={}, description=describe_attr("models", __doc__)
-    )
-    topics: dict[TopicName, TopicConfig] = Field(
-        ..., description=describe_attr("topics", __doc__)
     )
 
     class Config(DescConfig):
--- a/kpops/components/base_components/models/to_section.py
+++ b/kpops/components/base_components/models/to_section.py
@@ -83,11 +83,11 @@ class ToSection(BaseModel):
     :param models: Data models
     """
 
-    # TODO: really multiple models?
-    # any because snapshot versions must be supported
     topics: dict[TopicName, TopicConfig] = Field(
         default={}, description=describe_attr("topics", __doc__)
     )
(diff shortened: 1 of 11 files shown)


kpops/components/base_components/models/to_section.py
```diff
@@ -79,20 +79,18 @@ def extra_topic_role(cls, values):
...
-    :type models: dict[str, Any]
     :param topics: Output topics
-    :type topics: dict[str, TopicConfig]
+    :param models: Data models
     """
 
+    topics: dict[TopicName, TopicConfig] = Field(
+        default={}, description=describe_attr("topics", __doc__)
+    )
     # TODO: really multiple models?
```
sujuka99, line 89: Can we address this TODO somehow? I think it is very old.

sujuka99 (changes requested):
Please also update the docs.

[Example](https://bakdata.github.io/kpops/1.4/user/references/components/#__codelineno-1-19)

disrupted:
> Please also update the docs.
> 
> [Example](https://bakdata.github.io/kpops/1.4/user/references/components/#__codelineno-1-19)

do I just have to remove the `required` comment or are there any further changes that you have in mind?

sujuka99:
> > Please also update the docs.
> > [Example](https://bakdata.github.io/kpops/1.4/user/references/components/#__codelineno-1-19)
> 
> do I just have to remove the `required` comment or are there any further changes that you have in mind?

I think it would be nice to also add a comment that tells the user how to define the section as empty.

kpops/components/base_components/models/to_section.py
disrupted: absolutely! will do in a follow up PR :)

kpops/components/base_components/models/to_section.py
```diff
@@ -79,20 +79,18 @@ def extra_topic_role(cls, values):
...
     """Holds multiple output topics
 
-    :param models: Data models
-    :type models: dict[str, Any]
     :param topics: Output topics
-    :type topics: dict[str, TopicConfig]
+    :param models: Data models
     """
 
+    topics: dict[TopicName, TopicConfig] = Field(
```
raminqaf, line 86: Add a snapshot test for this please 🧪

disrupted: there is no new behavior here, the only thing is that `ToSection` can be instantiated without arguments now, same as `FromSection`

disrupted:
> I think it would be nice to also add a comment that tells the user how to define the section as empty.

isn't that the same as for any other value? you'd just set `to: null`

sujuka99:
> > I think it would be nice to also add a comment that tells the user how to define the section as empty.
> 
> isn't that the same as for any other value? you'd just set `to: null`

Hm, maybe we could say somewhere in the docs that you can set `to: {}` or `to: null` and the difference between the 2? I don't think that this is explained anywhere and it may not be intuitive for new users. Same for the other sections that behave the same.

disrupted:
> Hm, maybe we could say somewhere in the docs that you can set `to: {}` or `to: null` and the difference between the 2? I don't think that this is explained anywhere and it may not be intuitive for new users. Same for the other sections that behave the same.

this should be part of the [defaults documentation page](https://bakdata.github.io/kpops/1.4/user/references/defaults/) that explains how the enrichment works. It's more of a broader effort, since this applies to any pipeline key.

sujuka99 (approved):
LGTM :)

raminqaf (approved):

PR #322, opened 2023-08-15
Refactor Kafka Connect handler

disrupted:


kpops/components/base_components/kafka_connector.py
```diff
@@ -94,6 +94,18 @@ class KafkaConnector(PipelineComponent, ABC):
...
         pass
 
+    @validator("app", pre=True)
+    def connector_config_should_have_component_name(
+        cls, v: KafkaConnectorConfig | dict[str, str], values: dict[str, Any]
+    ) -> dict[str, str]:
+        if isinstance(v, KafkaConnectorConfig):
+            v = v.dict()
+        connector_name: str | None = v.get("name")
+        if connector_name and connector_name != values["name"]:
```
irux, line 104: wouldn't be better to use `if connector_name is not None` ?

sujuka99: What if the value is `""`?

kpops/component_handlers/kafka_connect/model.py
```diff
@@ -1,30 +1,47 @@
...
         extra = Extra.allow
+        alias_generator = to_dot
 
         @override
-        @staticmethod
-        def schema_extra(schema: dict[str, Any], model: type[BaseModel]) -> None:
-            schema["description"] = describe_object(model.__doc__)
+        @classmethod
+        def schema_extra(cls, schema: dict[str, Any], model: type[BaseModel]) -> None:
+            super().schema_extra(schema, model)
```
irux, line 28: Isn't it weird to use super() for a class method ? super is for instances. Shouldn't be `cls.schema_extra`?

sujuka99: Wouldn't that be a recursive call?

disrupted: hm, I am not aware of any implications using `super()`. essentially it would be something like `super(cls, cls)` but we can omit the arguments

kpops/components/base_components/kafka_connector.py
disrupted: I kinda prefer checking if the str is not empty as well, as we know by convention that every component has to have a name – so I don't think it matters

disrupted: oh no, you're right

disrupted: thanks! I've changed it and added a test

kpops/component_handlers/kafka_connect/model.py
sujuka99: @irux 

If I understood your suggestion correctly, `cls.schema_extra` should work equivalently to the example bellow:

```
>>> class A:
...     @classmethod
...     def print_class(cls):
...             print("Class method of A")
...
>>> A.print_class()
Class method of A
>>> class B(A):
...     @classmethod
...     def print_class(cls):
...             print("Class method of B")
...             cls.print_class()
...
>>> B.print_class
<bound method B.print_class of <class '__main__.B'>>
>>> B.print_class()
Class method of B
Class method of B
Class method of B
Class method of B
Class method of B
Class method of B
```

kpops/components/base_components/kafka_connector.py
```diff
@@ -94,6 +94,18 @@ class KafkaConnector(PipelineComponent, ABC):
     class Config(CamelCaseConfig):
         pass
 
+    @validator("app", pre=True)
+    def connector_config_should_have_component_name(
+        cls, v: KafkaConnectorConfig | dict[str, str], values: dict[str, Any]
```
raminqaf, line 99:
```suggestion
        cls, kafka_connect_config: KafkaConnectorConfig | dict[str, str], values: dict[str, Any]
```

disrupted: wouldn't it be better to rename it to `app` since that's the name of the attribute?

raminqaf (approved):

PR #326, opened 2023-08-17
Refactor component prefix & name

disrupted:
Don't concat prefix and name to override component name anymore. Instead, they are now separate properties


tests/components/test_producer_app.py
```diff
@@ -19,7 +19,7 @@
 
 class TestProducerApp:
     PRODUCER_APP_NAME = "test-producer-app-with-long-name-0123456789abcdefghijklmnop"
-    PRODUCER_APP_CLEAN_NAME = "test-producer-app-with-long-name-0123456789abc-clean"
+    PRODUCER_APP_CLEAN_NAME = "test-producer-app-with-long-n-clean"
```
raminqaf, line 22: Why did this change?

disrupted: It's trimmed a bit shorter now because of the `${pipeline_name}-` prefix, see below. Does that answer your question?

sujuka99 (approved):
Great change :)

LGTM!

tests/components/test_producer_app.py
raminqaf: Aha now I understand it 👍

raminqaf (approved):

irux (approved):
LGTM!

PR #332, opened 2023-08-22
Exclude abstract components from pipeline schema

disrupted:
- Check if component is abstract (has abstractmethods or directly inherits from ABC), if so exclude it
- Throw runtime error if no valid components are found
- Mark PipelineComponent & KafkaApp as abstract to exclude them from schema


irux (commented):
Do you have a test where it shows that this indeed crash when it inherits from abc/has abstract methods?

disrupted:
> Do you have a test where it shows that this indeed crash when it inherits from abc/has abstract methods?

it doesn't crash, we just don't want to endorse the usage of base components inside the pipeline.yaml. From the built-in components only streams-app, producer-app, kubernetes-app, and kafka-sink/source-connector should be used. The base pipeline-component, kafka-app and kafka-connector should only be used in defaults.yaml for which we can't generate a schema (yet).

disrupted:
> But I mean a test when you only have abstract components and raise the explained exception in the text of the pr, because it will exclude them. Or am I getting something wrong?

ah, I get what you mean. Yeah, I can add a test that the exception is raised if there aren't any valid components.

irux (approved):
LGTM

PR #340, opened 2023-08-23
Fix GitHub action error in non-Python projects

disrupted:
Fixes error when running in a deployment repo that doesn't contain a Python project

> No file in /home/runner/work/repo matched to [**/requirements.txt or **/pyproject.toml], make sure you have checked out the target repository

~~downside of removing the `setup-python` action is that it will only work on GitHub runners that have Python and pipx pre-installed, e.g. the standard ubuntu runners. I don't see a reason why someone would want to use a different runner for deploying with KPOps~~

contains first steps towards caching kpops package installation but is currently blocked by https://github.com/actions/setup-python/issues/361

### DIFF
--- a/actions/kpops-runner/action.yaml
+++ b/actions/kpops-runner/action.yaml
@@ -51,7 +51,7 @@ runs:
         version: ${{ inputs.helm-version }}
         token: ${{ inputs.token }}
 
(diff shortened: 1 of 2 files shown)


sujuka99 (approved):
LGTM

PR #346, opened 2023-08-30
Fix versioning of docs when releasing

raminqaf:


disrupted (commented):
looks good, please also clean the version selection dropdown manually to remove the outlier for 2.0.4

PR #353, opened 2023-09-05
Environment variables override config.yaml

sujuka99:


kpops/cli/pipeline_config.py
```diff
@@ -107,8 +107,8 @@ def customise_sources(
         ):
             return (
                 init_settings,
-                yaml_config_settings_source,
                 env_settings,
```
disrupted, line 110: shouldn't `env_settings` be at the very top

kpops/cli/pipeline_config.py
```diff
@@ -107,8 +107,8 @@ def customise_sources(
         ):
```
disrupted, line 107: add the return type annotation please

tests/pipeline/test_pipeline.py
```diff
@@ -458,8 +459,33 @@ def test_default_config(self, snapshot: SnapshotTest):
 
         snapshot.assert_match(enriched_pipeline, "test-pipeline")
 
+    def test_env_vars_precedence_over_config(
+        self,
+        snapshot: SnapshotTest,
+    ):
+        os.environ["KPOPS_KAFKA_BROKERS"] = "env_broker"
```
disrupted, line 466: consider using `monkeypatch.setenv` in tests. not sure if `os.environ` has any downsides tbh

disrupted (changes requested):
I would consider this a bugfix for something that should have been the default behavior. Therefore I would also remove the `breaking-change` label

disrupted:
Please rephrase the PR title and start with "Fix ..." so that it's easy to spot bugfixes 🙂

sujuka99:
> I would consider this a bugfix for something that should have been the default behavior. Therefore I would also remove the `breaking-change` label

I am not sure about that, [this has always been KPOps' behavior](https://github.com/bakdata/kpops/blob/c17dfc28ea335ce0e000ec60bde76c8425c6f5c0/kpops/cli/pipeline_config.py#L140).

kpops/cli/pipeline_config.py
sujuka99: Done, is it correct?

kpops/cli/pipeline_config.py
sujuka99: Moved it.

disrupted:
> > I would consider this a bugfix for something that should have been the default behavior. Therefore I would also remove the `breaking-change` label
> 
> I am not sure about that, [this has always been KPOps' behavior](https://github.com/bakdata/kpops/blob/c17dfc28ea335ce0e000ec60bde76c8425c6f5c0/kpops/cli/pipeline_config.py#L140).
> 
> The behavior is consistent and likely expected for some if not most of KPOps' users, so I wouldn't necessarily call it a bug.
> 
> If you are certain that it shouldn't be labeled breaking, I will remove it. I think you have a better grasp at the userbase 😁

I totally understand your concern. In this case I strongly believe that the expected way would be that environment variables take precedence, hence why I was so surprised myself that it's not the case. Otherwise we would also have to do a major release, which – in this case – is unnecessary imo.

kpops/cli/pipeline_config.py
```diff
@@ -104,11 +105,16 @@ def customise_sources(
             init_settings: SettingsSourceCallable,
             env_settings: SettingsSourceCallable,
             file_secret_settings: SettingsSourceCallable,
-        ):
+        ) -> tuple[
+            SettingsSourceCallable,
+            SettingsSourceCallable,
+            Callable[["PipelineConfig"], dict | list],
```
raminqaf, line 111: This is a crazy return type 😮 
```python
Callable[["PipelineConfig"], dict | list],
```
@disrupted is this really the return type?

kpops/cli/pipeline_config.py
disrupted: and you also need to change the return value of the `yaml_config_settings_source` function 😛

kpops/cli/pipeline_config.py
disrupted: see https://github.com/bakdata/kpops/pull/353#discussion_r1316071748

raminqaf: I think my main question is if we need to pas PipelineConfig as a string? 🧐

kpops/cli/pipeline_config.py
sujuka99: This throws an error:

```
Expression of type "tuple[SettingsSourceCallable, SettingsSourceCallable, (settings: PipelineConfig) -> (dict[Unknown, Unknown] | list[Unknown]), SettingsSourceCallable]" cannot be assigned to return type "tuple[SettingsSourceCallable, ...]"
  Type "(settings: PipelineConfig) -> (dict[Unknown, Unknown] | list[Unknown])" cannot be assigned to type "SettingsSourceCallable"
    Parameter 1: type "BaseSettings" cannot be assigned to type "PipelineConfig"
      "BaseSettings" is incompatible with "PipelineConfig"
    Function return type "dict[Unknown, Unknown] | list[Unknown]" is incompatible with type "Dict[str, Any]"
      Type "dict[Unknown, Unknown] | list[Unknown]" cannot be assigned to type "Dict[str, Any]"
        "list[Unknown]" is incompatible with "Dict[str, Any]"Pylance[reportGeneralTypeIssues](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportGeneralTypeIssues)
```

kpops/cli/pipeline_config.py
sujuka99: https://github.com/bakdata/kpops/pull/353#discussion_r1316106194

kpops/cli/pipeline_config.py
sujuka99: ah, I never saw your answers from 30 minutes ago, should've refreshed the page before answering :|

kpops/cli/pipeline_config.py
raminqaf: @sujuka99 can you change the string to the actual object?

kpops/cli/pipeline_config.py
```diff
@@ -1,11 +1,18 @@
...
 from pathlib import Path
+from typing import TYPE_CHECKING, Any
 
 from pydantic import BaseConfig, BaseSettings, Field
-from pydantic.env_settings import SettingsSourceCallable
 
 from kpops.component_handlers.helm_wrapper.model import HelmConfig, HelmDiffConfig
 from kpops.utils.yaml_loading import load_yaml_file
 
+if TYPE_CHECKING:
```
disrupted, line 11: why is this necessary?

kpops/cli/pipeline_config.py
```diff
@@ -104,17 +111,20 @@ def customise_sources(
             init_settings: SettingsSourceCallable,
             env_settings: SettingsSourceCallable,
             file_secret_settings: SettingsSourceCallable,
-        ):
+        ) -> tuple[SettingsSourceCallable | Callable[[PipelineConfig], dict], ...]:
```
disrupted, line 114:
```suggestion
        ) -> tuple[SettingsSourceCallable | Callable[[PipelineConfig], dict[str, Any]], ...]:
```

kpops/cli/pipeline_config.py
```diff
@@ -1,11 +1,18 @@
...
+from typing import TYPE_CHECKING, Any
 
 from pydantic import BaseConfig, BaseSettings, Field
-from pydantic.env_settings import SettingsSourceCallable
 
 from kpops.component_handlers.helm_wrapper.model import HelmConfig, HelmDiffConfig
 from kpops.utils.yaml_loading import load_yaml_file
 
+if TYPE_CHECKING:
+    from collections.abc import Callable
```
disrupted, line 12: shouldn't this be
```suggestion
    from typing import Callable
```
?

disrupted (changes requested):

kpops/cli/pipeline_config.py
sujuka99: importing it from `typing` is deprecated in favor of `collections.abc` I believe. [link](https://docs.python.org/3.10/library/typing.html#callable)

kpops/cli/pipeline_config.py
sujuka99: See: [TCH002](https://beta.ruff.rs/docs/rules/typing-only-third-party-import/), [TCH003](https://beta.ruff.rs/docs/rules/typing-only-standard-library-import/)

kpops/cli/pipeline_config.py
disrupted: TIL, good point

kpops/cli/pipeline_config.py
disrupted: are you using Ruff locally already? 😄

kpops/cli/pipeline_config.py
```diff
@@ -104,17 +111,22 @@ def customise_sources(
...
 
-def yaml_config_settings_source(settings: PipelineConfig) -> dict | list:
+def yaml_config_settings_source(settings: PipelineConfig) -> dict[str, Any]:
     path_to_config = settings.Config.config_path
     if path_to_config.exists():
-        return load_yaml_file(path_to_config)
+        if isinstance(source := load_yaml_file(path_to_config), dict):
+            return source
+        err_msg = f"{path_to_config} must be a mapping."
+        raise TypeError(err_msg)
```
disrupted, line 131: nit: would prefer if you skipped the variable assignment if it's not used anywhere else

disrupted (approved):

raminqaf (approved):

kpops/cli/pipeline_config.py
sujuka99: I have been for some time now :grin: 

I like it a lot so far

kpops/cli/pipeline_config.py
sujuka99: [reason](https://beta.ruff.rs/docs/rules/raw-string-in-exception/)

If you think it isn't justified I can skip the assignment

disrupted: TIL, good point!

kpops/cli/pipeline_config.py
disrupted: nice, looking forward to adopting it for the project

PR #362, opened 2023-09-22
Fix environment variables documentation generation

### LINKED ISSUE #359: Doc generation script for environment variables should visit nested fields
see https://github.com/bakdata/kpops/pull/354/files#r1321209743

sujuka99:
closes #359

### DIFF
--- a/hooks/gen_docs/__init__.py
+++ b/hooks/gen_docs/__init__.py
@@ -6,11 +6,7 @@ from typing import Any
 
 
 class StrEnum(str, Enum):
-    """Adds constructors that return all items in a ``Generator``.
-
-    Introduces constructors that return a ``Generator`` object
-    either containing all items, only their names or their values.
-    """
+    """Temporary replacement for StrEnum while we suppport python3.10"""
 
     @classmethod
     def items(cls) -> Generator[tuple[Any, str], None, None]:
--- a/hooks/gen_docs/__init__.py
+++ b/hooks/gen_docs/__init__.py
@@ -1,24 +1,23 @@
 """Documentation generation."""
 
-from collections.abc import Generator
+from collections.abc import Iterator
 from enum import Enum
-from typing import Any
 
 
-class StrEnum(str, Enum):
+class IterableStrEnum(str, Enum):
     """Temporary replacement for StrEnum while we suppport python3.10"""
 
     @classmethod
-    def items(cls) -> Generator[tuple[Any, str], None, None]:
+    def items(cls) -> Iterator[tuple[str, str]]:
         """Return all item names and values in tuples."""
         return ((e.name, e.value) for e in cls)
 
     @classmethod
-    def keys(cls) -> Generator[str, None, None]:
+    def keys(cls) -> Iterator[str]:
         """Return all item names."""
         return (e.name for e in cls)
 
     @classmethod
-    def values(cls) -> Generator[Any, None, None]:
+    def values(cls) -> Iterator[str]:
         """Return all item values."""
         return (e.value for e in cls)
--- a/hooks/gen_docs/__init__.py
+++ b/hooks/gen_docs/__init__.py
@@ -7,7 +7,7 @@ from enum import Enum
 class IterableStrEnum(str, Enum):
     """Polyfill that also introduces dict-like behavior
 
-    Introduces constructors that return a ``Generator`` object
+    Introduces constructors that return a ``Iterator`` object
     either containing all items, only their names or their values.
     """
 
--- a/hooks/gen_docs/__init__.py
+++ b/hooks/gen_docs/__init__.py
@@ -5,7 +5,11 @@ from enum import Enum
 
 
 class IterableStrEnum(str, Enum):
-    """Temporary replacement for StrEnum while we suppport python3.10"""
+    """Polyfill that also introduces dict-like behavior
+
+    Introduces constructors that return a ``Generator`` object
+    either containing all items, only their names or their values.
+    """
 
     @classmethod
     def items(cls) -> Iterator[tuple[str, str]]:
--- a/hooks/gen_docs/__init__.py
+++ b/hooks/gen_docs/__init__.py
@@ -15,12 +15,12 @@ class SuperEnum(Generic[_T], Enum):
     """
 
     @classmethod
-    def items(cls) -> Generator[tuple[_T, Any], None, None]:
+    def items(cls) -> Generator[tuple[Any, _T], None, None]:
         """Return all item names and values in tuples."""
         return ((e.name, e.value) for e in cls)
 
     @classmethod
-    def keys(cls) -> Generator[_T, None, None]:
+    def keys(cls) -> Generator[str, None, None]:
         """Return all item names."""
         return (e.name for e in cls)
 
--- a/hooks/gen_docs/__init__.py
+++ b/hooks/gen_docs/__init__.py
@@ -2,12 +2,10 @@
 
 from collections.abc import Generator
 from enum import Enum
-from typing import Any, Generic, TypeVar
-
-_T = TypeVar("_T")
+from typing import Any
 
 
-class SuperEnum(Generic[_T], Enum):
+class StrEnum(str, Enum):
     """Adds constructors that return all items in a ``Generator``.
 
     Introduces constructors that return a ``Generator`` object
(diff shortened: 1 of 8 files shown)


tests/utils/test_doc_gen.py
```diff
@@ -0,0 +1,15 @@
...
+    def test_collect_fields(self):
+        collected_field_defaults = [
+            "not_nested_field",
+            "attr",
+            Ellipsis,
+        ]
+        collected_defaults = [
+            field.field_info.default for field in collect_fields(ParentSettings)
+        ]
+        assert collected_defaults == collected_field_defaults
```
raminqaf, line 15:
```suggestion
        expected = [
            "not_nested_field",
            "attr",
            Ellipsis,
        ]
        actual = [
            field.field_info.default for field in collect_fields(ParentSettings)
        ]
        assert actual == expected
```

hooks/gen_docs/gen_docs_cli_usage.py
```diff
@@ -21,7 +21,7 @@
         "--output",
         str(PATH_CLI_COMMANDS_DOC),
     ]
-    subprocess.run(typer_args)
+    subprocess.run(typer_args, check=True, capture_output=True)
```
raminqaf, line 24: Why do we need to set these two?

sujuka99: https://docs.astral.sh/ruff/rules/subprocess-run-without-check/

I thought that it's probably better that we have an explicit exception thrown upon failure.

I set `capture_output` because of the following text in `subprocess.run`'s documentation:

If check is True and the exit code was non-zero, it raises a CalledProcessError. The CalledProcessError object will have the return code in the returncode attribute, and output & stderr attributes if those streams were captured.

hooks/gen_docs/__init__.py
```diff
@@ -2,25 +2,23 @@
...
+class StrEnum(str, Enum):
     """Adds constructors that return all items in a ``Generator``.
 
     Introduces constructors that return a ``Generator`` object
     either containing all items, only their names or their values.
     """
 
     @classmethod
-    def items(cls) -> Generator[tuple[_T, Any], None, None]:
+    def items(cls) -> Generator[tuple[Any, str], None, None]:
```
disrupted, line 16: isn't the type `[str, str]`?

hooks/gen_docs/__init__.py
```diff
@@ -2,25 +2,23 @@
...
         """Return all item names and values in tuples."""
         return ((e.name, e.value) for e in cls)
 
     @classmethod
-    def keys(cls) -> Generator[_T, None, None]:
+    def keys(cls) -> Generator[str, None, None]:
```
disrupted, line 21: consider annotating as `Iterator[str]`

disrupted (changes requested):

hooks/gen_docs/gen_docs_env_vars.py
```diff
@@ -248,28 +250,37 @@ def __fill_csv_pipeline_config(target: Path) -> None:
...
-            config_field.field_info.description
+    for field in collect_fields(PipelineConfig):
+        field_info = PipelineConfig.Config.get_field_info(field.name)
+        field_description: str = (
+            field.field_info.description
             or "No description available, please refer to the pipeline config documentation."
         )
-        config_field_default = None or config_field.field_info.default
-        if config_env_var := config_field_info.get(
+        field_default = None or field.field_info.default
```
disrupted, line 259: this will always be evaluated as `field.field_info.default`

hooks/gen_docs/__init__.py
```diff
@@ -2,25 +2,23 @@
...
 from collections.abc import Generator
 from enum import Enum
-from typing import Any, Generic, TypeVar
+from typing import Any
 
-_T = TypeVar("_T")
 
-
-class SuperEnum(Generic[_T], Enum):
+class StrEnum(str, Enum):
```
disrupted, line 8: how about `IterableStrEnum`?

tests/utils/test_doc_gen.py
raminqaf: @sujuka99 Still not resolved 😄

hooks/gen_docs/gen_docs_env_vars.py
sujuka99: Ah, I thought that if the second value is also false the first one would stick

disrupted: you could flip them to achieve the Falsy → `None` logic

tests/utils/test_doc_gen.py
sujuka99: Should be now :)

hooks/gen_docs/__init__.py
sujuka99: indeed..

hooks/gen_docs/__init__.py
sujuka99: Could I annotate the other 2 as `Iterator`, too?

disrupted: yep

hooks/gen_docs/__init__.py
```diff
@@ -1,30 +1,23 @@
...
 
-
-class SuperEnum(Generic[_T], Enum):
-    """Adds constructors that return all items in a ``Generator``.
-
-    Introduces constructors that return a ``Generator`` object
-    either containing all items, only their names or their values.
-    """
+class IterableStrEnum(str, Enum):
+    """Temporary replacement for StrEnum while we suppport python3.10"""
```
disrupted, line 8: it's not just that, no?
that would be a fitting description for a simple polyfill like `class StrEnum(str, Enum): ...`
but here the focus is more on the extra functionality that makes it similar to a dict, wdyt?

hooks/gen_docs/gen_docs_env_vars.py
```diff
@@ -248,29 +254,38 @@ def __fill_csv_pipeline_config(target: Path) -> None:
...
-                config_field_description,
-                config_field.name,
+                field_default,
+                field_description,
+                field.name,
             )
 
 
-def __fill_csv_cli(target: Path) -> None:
+def collect_fields(settings: type[BaseSettings]) -> Generator[ModelField, None, None]:
```
disrupted, line 276: Iterator

disrupted (changes requested):

tests/utils/test_doc_gen.py
```diff
@@ -0,0 +1,254 @@
...
+        ],
+    )
+    def test_csv_append_env_var(
+        self,
+        tmp_path,
+        var_name,
+        default_value,
+        description,
+        extra_args,
+        expected_outcome,
```
disrupted, line 74: types?

hooks/gen_docs/gen_docs_env_vars.py
disrupted, line 1: in `gen_vars` the `columns` parameter should be `list[str]`

disrupted (changes requested):

hooks/gen_docs/__init__.py
sujuka99: Hmm you are right. I had a (wrong) memory that `StrEnum` already had that functionality somehow built in.

hooks/gen_docs/gen_docs_env_vars.py
sujuka99: fixed

hooks/gen_docs/__init__.py
```diff
@@ -1,30 +1,27 @@
...
 
-_T = TypeVar("_T")
 
-
-class SuperEnum(Generic[_T], Enum):
-    """Adds constructors that return all items in a ``Generator``.
+class IterableStrEnum(str, Enum):
+    """Polyfill that also introduces dict-like behavior
 
     Introduces constructors that return a ``Generator`` object
```
disrupted, line 10: either lowercase generator or `Iterator` referring to the return type

tests/utils/test_doc_gen.py
```diff
@@ -0,0 +1,252 @@
...
+        ],
+    )
+    def test_csv_append_env_var(
+        self,
+        tmp_path: Path,
+        var_name: str,
+        default_value: Any,
+        description: str | list[str],
+        extra_args: tuple,
+        expected_outcome,
```
disrupted, line 72: type?

disrupted (commented):
looks good, just a nitpick

tests/utils/test_doc_gen.py
sujuka99: Ah, I missed it somehow, added now

disrupted (approved):

PR #363, opened 2023-09-26
Introduce ruff

sujuka99:
Replaces `isort`, `pyupgrade`, `flake8` with `ruff`


sujuka99:
I would also like to add at some point [ANN](https://docs.astral.sh/ruff/rules/#flake8-annotations-ann) to ensure that we annotate our code consistently, but currently I don't have it configured properly locally and it's just annoying. I would have to play around with it for a bit before I can (maybe) reach an acceptable state where it is useful, but not inconvenient.

tests/component_handlers/schema_handler/test_schema_handler.py
```diff
@@ -96,79 +96,91 @@ def test_should_lazy_load_schema_provider(find_class_mock: MagicMock):
...
+            id="components_module = None",
+        ),
+        pytest.param(
+            "",
+            id="components_module = ''",
+        ),
+    ],
+)
+def test_should_raise_value_error_when_schema_provider_is_called_and_components_module_is_empty(
+    components_module,
```
disrupted, line 149: is there a rule to mark untyped parameters as warning?

.github/workflows/ci.yaml
```diff
@@ -35,11 +35,8 @@ jobs:
...
-      - name: Lint (flake8)
-        run: poetry run pre-commit run flake8 --all-files --show-diff-on-failure
-
-      - name: Order of imports (isort)
-        run: poetry run pre-commit run isort --all-files --show-diff-on-failure
+      - name: Lint (ruff)
+        run: poetry run pre-commit run ruff --all-files --show-diff-on-failure
 
       - name: Formatting (black)
         run: poetry run pre-commit run black --all-files --show-diff-on-failure
```
disrupted, line 42: what's your opinion on creating a separate branch to test `ruff format` so that we can see the diff? if it doesn't change between alpha and beta I'd say we can switch to it.

kpops/cli/main.py
```diff
@@ -120,12 +122,17 @@ def setup_pipeline(
 
     handlers = setup_handlers(components_module, pipeline_config)
     return Pipeline.load_from_yaml(
-        pipeline_base_dir, pipeline_path, registry, pipeline_config, handlers
+        pipeline_base_dir,
+        pipeline_path,
+        registry,
+        pipeline_config,
+        handlers,
```
disrupted, line 129: is this a new formatting rule?

kpops/components/base_components/base_defaults_component.py
```diff
@@ -165,20 +170,23 @@ def defaults_from_yaml(path: Path, key: str) -> dict:
     """
     content = load_yaml_file(path, substitution=ENV)
     if not isinstance(content, dict):
-        raise TypeError(
+        msg = (
             "Default files should be structured as map ([app type] -> [default config]"
         )
+        raise TypeError(
+            msg,
```
disrupted, line 177: tbh not a fan of the forced trailing comma :( this could easily be a one liner

tests/component_handlers/schema_handler/test_schema_handler.py
sujuka99: `ANN`

.github/workflows/ci.yaml
sujuka99: Sounds good

kpops/components/base_components/base_defaults_component.py
sujuka99: This probably happened because it used to be indented like that before. If I make it a one-liner it will not be flagged

kpops/cli/main.py
sujuka99: Nop, maybe something enforced line length :thinking:

tests/component_handlers/schema_handler/test_schema_handler.py
disrupted: ah ok, sorry I just saw your comment about it

disrupted: otherwise I can be the one to enforce this rule manually 🤣

kpops/cli/main.py
disrupted: imo for function definitions with a lot of parameters this formatting is good, but function calls take up a lot of space this way

kpops/components/base_components/base_defaults_component.py
sujuka99: Tested it now, `black` did it, because of the forced comma

sujuka99: This will be flagged and formatted by both `black` and `ruff format`
```
    def apply_component(
        self, component_class: type[PipelineComponent], component_data: dict,
    ) -> None:
```

This won't:
```
    def apply_component(
        self, component_class: type[PipelineComponent], component_data: dict
    ) -> None:
```

disrupted: do we have to force the trailing comma though?

tests/component_handlers/schema_handler/test_schema_handler.py
disrupted: alternatively, we could configure Pyright to warn on symbols where it needs a type annotation, i.e. in all cases where it can't be inferred.

kpops/components/base_components/base_defaults_component.py
sujuka99: How can I revert just the [COM](https://github.com/bakdata/kpops/pull/363/commits/301284eb02dbb707fefa5ab50476b06ebee757ff) commit?

disrupted: `git revert 301284eb02dbb707fefa5ab50476b06ebee757ff`?

sujuka99: WDYT about getting rid of [COM812](https://docs.astral.sh/ruff/rules/missing-trailing-comma/), but keeping [COM818](https://docs.astral.sh/ruff/rules/trailing-comma-on-bare-tuple/) and [COM819](https://docs.astral.sh/ruff/rules/prohibited-trailing-comma/)

disrupted: good idea, that looks ideal

tests/component_handlers/schema_handler/test_schema_handler.py
sujuka99: :thinking: That's quite nice, it is probably the better approach imo

kpops/cli/main.py
sujuka99: Ah, the whole time I thought we had [this discussion](https://github.com/bakdata/kpops/pull/363#discussion_r1338252174) here.

disrupted:
would it be possible to have a _problem matcher_ for Ruff where it directly annotates the errors in the PR diff? similar to [actionlint](https://github.com/rhysd/actionlint/blob/main/docs/usage.md#problem-matchers)

kpops/cli/main.py
sujuka99: What a nightmare... Had to cherry-pick commits one-by-one after the `revert`

Not sure what I did wrong, really tried to only revert the comma related stuff, but to no avail

sujuka99: Should be good now tho

sujuka99:
> would it be possible to have a _problem matcher_ for Ruff where it directly annotates the errors in the PR diff? similar to [actionlint](https://github.com/rhysd/actionlint/blob/main/docs/usage.md#problem-matchers)

Should be, should I try to find or write one in this PR?

disrupted:
> > would it be possible to have a _problem matcher_ for Ruff where it directly annotates the errors in the PR diff? similar to [actionlint](https://github.com/rhysd/actionlint/blob/main/docs/usage.md?rgh-link-date=2023-09-27T10%3A31%3A17Z#problem-matchers)
> 
> Should be, should I try to find or write one in this PR?

that would be great. I know that for example Home Assistant has one if you wanna try that

kpops/component_handlers/helm_wrapper/helm.py
```diff
@@ -31,17 +29,16 @@ def __init__(self, helm_config: HelmConfig) -> None:
...
+            raise RuntimeError(
+                f"The supported Helm version is 3.x.x. The current Helm version is {self._version.major}.{self._version.minor}.{self._version.patch}"
+            )
 
     def add_repo(
         self,
         repository_name: str,
         repository_url: str,
-        repo_auth_flags: RepoAuthFlags | None = None,
+        repo_auth_flags: RepoAuthFlags = RepoAuthFlags(),
```
disrupted, line 40: please avoid changing any logic in this PR if possible

disrupted: ah, you're just doing that to test the problem matcher, right?

sujuka99:
[The problem matcher seems to work now](https://github.com/bakdata/kpops/pull/363/commits/ed8b99d49589f54f4bc2649f78eb59d570c7645f#diff-1ca7465daf9238dfe221304df636d96704ebf4935288b854a5c4c6b3ea9e3162)

pyproject.toml
```diff
@@ -67,6 +66,99 @@ mike = "^1.1.2"
...
+  "PLW2901", # `for` loop variable `{var}` overwritten by assignment target -- Inconvenient to enforce
+  "TRY002", # Create your own exception -- Inconvenient to enforce
+  "TRY003", # Avoid specifying long messages outside the exception class -- Inconvenient to enforce
+]
+select = [
+  "F", # Pyflakes
+  "E", # pycodestyle Errors
+  "W", # pycodestyle Warnings
+  "C90", # mccabe
+  "I", # isort
```
disrupted, line 115: should we set
```toml
[tool.ruff.isort]
split-on-trailing-comma = false
```
to preserve the current behavior?
https://docs.astral.sh/ruff/settings/#isort-split-on-trailing-comma

disrupted:
> [The problem matcher seems to work now](https://github.com/bakdata/kpops/pull/363/commits/ed8b99d49589f54f4bc2649f78eb59d570c7645f#diff-1ca7465daf9238dfe221304df636d96704ebf4935288b854a5c4c6b3ea9e3162)

is it possible to change the displayed name from _Test_ to _Ruff_?

README.md
```diff
@@ -3,6 +3,7 @@
 [![Build status](https://github.com/bakdata/kpops/actions/workflows/ci.yaml/badge.svg)](https://github.com/bakdata/kpops/actions/workflows/ci.yaml)
 [![pypi](https://img.shields.io/pypi/v/kpops.svg)](https://pypi.org/project/kpops)
 [![versions](https://img.shields.io/pypi/pyversions/kpops.svg)](https://github.com/bakdata/kpops)
+[![Ruff](https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/astral-sh/ruff/main/assets/badge/v2.json)](https://github.com/astral-sh/ruff)
```
disrupted, line 6: haha cool 😎

pyproject.toml
sujuka99: Sure!

sujuka99:
> > [The problem matcher seems to work now](https://github.com/bakdata/kpops/pull/363/commits/ed8b99d49589f54f4bc2649f78eb59d570c7645f#diff-1ca7465daf9238dfe221304df636d96704ebf4935288b854a5c4c6b3ea9e3162)
> 
> is it possible to change the displayed name from _Test_ to _Ruff_?

The only way I can see is to put Ruff into a separate job

raminqaf:
I agree with @disrupted that we should gradually introduce the changes (fixes) made by Ruff. Maybe it would be good to ignore most of the (critical) rules at first and do them step by step in follow-up PRs to avoid breaking the code.

disrupted:
> I agree with @disrupted that we should gradually introduce the changes (fixes) made by Ruff. Maybe it would be good to ignore most of the (critical) rules at first and do them step by step in follow-up PRs to avoid breaking the code.

I will look through all of the changes today and add comments if I run into particularly big changes. otherwise, of course it's best practice to bundle formatting changes in one big commit since it's easier to exclude from git blame.

disrupted:
> > is it possible to change the displayed name from _Test_ to _Ruff_?
> 
> The only way I can see is to put Ruff into a separate job

ok, then it's fine

.github/ruff-matcher.json
disrupted, line 1: do you think there would be an advantage of using separate matchers for warnings and errors similar to how [Home Assistant configures it](https://github.com/home-assistant/core/blob/217a895b5a0ffe67ae17f184ca577b4ed7cff955/.github/workflows/matchers/ruff.json)?

.github/workflows/ci.yaml
```diff
@@ -35,11 +35,17 @@ jobs:
...
 
-      - name: Lint (flake8)
-        run: poetry run pre-commit run flake8 --all-files --show-diff-on-failure
-
-      - name: Order of imports (isort)
-        run: poetry run pre-commit run isort --all-files --show-diff-on-failure
+      - name: Lint (ruff)
+        shell: bash
+        run: |
+          if [[ "$RUNNER_OS" == "Linux" && $(python -V) == *"3.10"* ]];
```
disrupted, line 41:
```suggestion
          if [[ "$RUNNER_OS" == "Linux" && "${{ matrix.python-version }}" == "3.10" ]];
```

.pre-commit-config.yaml
```diff
@@ -1,10 +1,23 @@
...
-        exclude: ^tests/.*snapshots/
+      - id: ruff
+        name: ruff
+        entry: ruff check .
+        args: [ --config, pyproject.toml, --fix, --show-fixes, --exit-non-zero-on-fix ]
+        language: system
+        types_or: [python]
+        require_serial: true # run once for all files
+        pass_filenames: false
+      - id: ruff-ci
```
disrupted, line 12: this one is only used for the one CI job with the problem matcher enabled, right? does it then even need to be in pre-commit?

kpops/component_handlers/__init__.py
```diff
@@ -2,11 +2,10 @@
 
 from typing import TYPE_CHECKING
 
-from kpops.component_handlers.kafka_connect.kafka_connect_handler import (
-    KafkaConnectHandler,
-)
-
 if TYPE_CHECKING:
+    from kpops.component_handlers.kafka_connect.kafka_connect_handler import (
+        KafkaConnectHandler,
```
disrupted, line 7: get rid of the trailing comma

kpops/component_handlers/helm_wrapper/helm.py
```diff
@@ -29,16 +31,17 @@ def __init__(self, helm_config: HelmConfig) -> None:
...
 
     def add_repo(
         self,
         repository_name: str,
         repository_url: str,
-        repo_auth_flags: RepoAuthFlags = RepoAuthFlags(),
+        repo_auth_flags: RepoAuthFlags | None = None,
     ) -> None:
+        if repo_auth_flags is None:
+            repo_auth_flags = RepoAuthFlags()
```
disrupted, line 44: good! 👏🏻

kpops/component_handlers/helm_wrapper/helm.py
```diff
@@ -206,8 +212,8 @@ def __execute(self, command: list[str]) -> str:
         log.debug(f"Executing {' '.join(command)}")
         process = subprocess.run(
             command,
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
+            check=True,
+            capture_output=True,
```
disrupted, line 216: did you verify that those are equal?

kpops/utils/dict_differ.py
```diff
@@ -53,9 +57,9 @@ def from_dicts(
         d1: dict, d2: dict, ignore: set[str] | None = None
     ) -> Iterator[Diff]:
         for diff_type, keys, changes in diff(d1, d2, ignore=ignore):
-            if not isinstance(changes, list):
-                changes = [("", changes)]
-            for key, change in changes:
+            if not isinstance(changes_tmp := changes, list):
+                changes_tmp = [("", changes)]
+            for key, change in changes_tmp:
```
disrupted, line 62: could you explain this change?

pyproject.toml
```diff
@@ -67,6 +66,102 @@ mike = "^1.1.2"
...
+"tests/*/__init__.py" = ["F401"]
+
+[tool.ruff.isort]
+split-on-trailing-comma = false
+
+[tool.ruff.flake8-bugbear]
+extend-immutable-calls = ["typer.Argument"]
+
+[tool.ruff.flake8-type-checking]
+runtime-evaluated-base-classes = ["pydantic.BaseModel", "kpops.components.base_components.kafka_app.KafkaApp"]
```
disrupted, line 163: why KafkaApp?

tests/cli/snapshots/snap_test_schema_generation.py
disrupted, line 1: don't format the snapshots, they're automatically generated

tests/cli/test_schema_generation.py
```diff
@@ -118,7 +120,6 @@ def test_gen_pipeline_schema_no_components(self):
                 ],
                 catch_exceptions=False,
             )
-            assert result.exit_code == 1
```
disrupted, line 121: why did you remove the assertion? probably it should just be moved outside the context manager, no?

tests/compiler/test_pipeline_name.py
```diff
@@ -8,49 +8,51 @@
...
-    assert "random" == ENV["pipeline_name_1"]
-    assert "path" == ENV["pipeline_name_2"]
-    assert "for" == ENV["pipeline_name_3"]
-    assert "testing" == ENV["pipeline_name_4"]
+    assert ENV["pipeline_name"] == "some-random-path-for-testing"
+    assert ENV["pipeline_name_0"] == "some"
+    assert ENV["pipeline_name_1"] == "random"
+    assert ENV["pipeline_name_2"] == "path"
+    assert ENV["pipeline_name_3"] == "for"
+    assert ENV["pipeline_name_4"] == "testing"
```
disrupted, line 22: thanks, that was bothering me too 😅

tests/component_handlers/schema_handler/test_schema_handler.py
```diff
@@ -116,45 +116,48 @@ def test_should_raise_value_error_if_schema_provider_class_not_found():
...
```
disrupted, line 144: add type annotation

tests/pipeline/snapshots/snap_test_example.py
disrupted, line 1: don't format

tests/pipeline/snapshots/snap_test_pipeline.py
disrupted, line 1: don't format

disrupted (dismissed):
looks great 🤩 
(almost) no complaints :)

kpops/utils/dict_differ.py
sujuka99: [PLW2901](https://docs.astral.sh/ruff/rules/redefined-loop-name/) complained. I was not sure whether to suppress it, but ended up following the advice.

.github/ruff-matcher.json
sujuka99: Is the matcher output supposed to be read mainly by reviewers? If I could run the same matcher locally, it would be quite nice, otherwise it may not be as useful. (to me, that is)

We would have to decide for each rule-group whether it is a warning or an error, so the regex would become rather long with our current rule set and I'm not sure how maintaining it would be like.

We could only specify the shorter list and include/exclude? Gotta remember to keep it up to date after each rule addition tho :x

.pre-commit-config.yaml
sujuka99: Good point, it doesn't, will fix

.github/ruff-matcher.json
disrupted: ah, that would be quite inconvenient. Then let's keep your more inclusive/robust regex.
yeah, mostly during review

kpops/component_handlers/helm_wrapper/helm.py
sujuka99: `check` is an addition, [adds some safety](https://docs.astral.sh/ruff/rules/subprocess-run-without-check/)

`capture_output` is equivalent to the removed lines

pyproject.toml
sujuka99: Not sure why that is the fix, but otherwise [TCH001](https://docs.astral.sh/ruff/rules/typing-only-first-party-import/) moves a couple imports into a `if TYPE_CHECKING: ...` block and breaks KPOps. `pydantic.BaseModel` should be enough to prevent that from happening I think, but maybe the `ABC` parent of `KafkaApp` throws `ruff` off.

tests/cli/test_schema_generation.py
sujuka99: :thinking: true, I don't remember why I removed, probably didn't intend to.

Will bring it back.

sujuka99: I read a bit and I think it could be that we don't need that assertion. We expect an exception, so either we get it and the assertion isn't reached or we don't get the exception and the test fails anyway.

sujuka99: [See here](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.raises):

> When using pytest.raises as a context manager, it’s worthwhile to note that normal context manager rules apply and that the exception raised must be the final line in the scope of the context manager. Lines of code after that, within the scope of the context manager will not be executed. For example:

```python
>>> value = 15
>>> with pytest.raises(ValueError) as exc_info:
...     if value > 10:
...         raise ValueError("value must be <= 10")
...     assert exc_info.type is ValueError  # this will not execute
Instead, the following approach must be taken (note the difference in scope):
```
```python
>>> with pytest.raises(ValueError) as exc_info:
...     if value > 10:
...         raise ValueError("value must be <= 10")
...
>>> assert exc_info.type is ValueError
```

disrupted: makes sense

.github/workflows/ci.yaml
sujuka99: https://github.com/bakdata/kpops/pull/365

tests/component_handlers/schema_handler/test_schema_handler.py
sujuka99: Outright enabling `strict` mode produced 999 errors, 
```toml
[tool.pyright]
strict = ["*"]
```
while enabling just the following 5 rules produced 849:
```toml
[tool.pyright]
reportUnknownArgumentType = "error"
reportUnknownLambdaType = "error"
reportUnknownMemberType = "error"
reportUnknownParameterType = "error"
reportUnknownVariableType = "error"
```

[Pyright configuration docs](https://microsoft.github.io/pyright/#/configuration)

pyproject.toml
disrupted: perhaps I am getting a slightly different error, it says to move `StreamsAppConfig` (in `StreamsApp`) and `ProducerValues` (in `StreamsApp`) (*weird naming discrepancy btw) in a type checking block. would that be fine?

sujuka99: It's the same error, it caused crashes when I tried, no idea why tbh

disrupted: what do you mean by crashes? kpops runtime errors?

sujuka99: yes

sujuka99: If you remove `KafkaApp` from the list and let `ruff` autofix the `TCH001` errors with `ruff check . --fix` `pytest` would fail

disrupted: I think we should open an issue for that, as you said flagging `pydantic.BaseModel` should suffice

sujuka99: Here is an example: `FAILED tests/components/test_streams_app.py::TestStreamsApp::test_set_topics - pydantic.errors.ConfigError: field "app" not yet prepared so type is still a ForwardRef, you might need to call StreamsApp.update_for
ward_refs().`

sujuka99: > I think we should open an issue for that, as you said flagging `pydantic.BaseModel` should suffice. I could imagine that the parent class resolving has a max depth or something similar which causes them not to be detected as children

Could be

sujuka99: The problem persists if you replace `KafkaApp` with `PipelineComponent` or `KubernetesApp`

sujuka99: Shouldn't be a depth issue in that case I think

sujuka99: Also persists with `StreamsApp` in place of `KafkaApp`

disrupted: yep, that's a good title

.github/workflows/ci.yaml
```diff
@@ -59,11 +64,6 @@ jobs:
       - name: Generate pipeline definitions
         run: poetry run pre-commit run gen-docs-components --all-files --show-diff-on-failure
 
-      # TODO: enable when PEP 604 incompatibilty is in typer is resolved https://github.com/tiangolo/typer/issues/348
```
disrupted, line 62: is this enabled now in Ruff? I believe Typer still doesn't support the new type union syntax

sujuka99: It is not, but I have excluded only that particular rule from pyupgrade

sujuka99: https://github.com/bakdata/kpops/pull/363/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R96

pyproject.toml
sujuka99: when I remove it the error no longer appears

disrupted: ok, then let's do it like this for now. still sounds like a bug that we should file an issue for, because `StreamsAppConfig` and `ProducerValues` weren't resolved as children of `pydantic.BaseModel`

sujuka99: Do you mean that I should get rid of the import?

tests/component_handlers/schema_handler/test_schema_handler.py
sujuka99: #367

disrupted (approved):

PR #368, opened 2023-10-10
Enable transparent lookups from internal environment on OS environment

### LINKED ISSUE #317: Environment variables are not updated after kpops import
#217 introduced a wrapper for `os.environ`. This, however, freezes the environment known to kpops to the state when `ENV` is first (transitively) imported. If kpops is called programmatically, changes to `os.environ` after importing kpops are ignored.

Minimal setup:
```
# defaults.yaml
kafka-app:
  app:
    streams:
      brokers: ${broker}
      schemaRegistryUrl: ${schema_registry_url}
```
```
# test/pipeline.yaml
- type: streams-app
  name: test
  to:
    topics:
      ${PIPELINE_PREFIX}topic:
        type: output
```
```
# main.py

import os
import sys
from pathlib import Path

os.environ.update(
    {
        "KPOPS_KAFKA_BROKERS": "http://brokers",
        "KPOPS_SCHEMA_REGISTRY_URL": "http://schema-registry",
        "KPOPS_CONNECT_HOST": "http://connect",
        "KPOPS_REST_PROXY_HOST": "http://rest-proxy",
        "KPOPS_ENVIRONMENT": "dev"
    }
)

# Case 1: NOT expanded
from kpops.cli import main

os.environ["PIPELINE_PREFIX"] = "my-prefix-"

# Case 2: expanded
# from kpops.cli import main

if __name__ == "__main__":
     # this might not be the intended way to call kpops programmatically, though :D 
    sys.argv[1:] = [
        "generate",
        "test/pipeline.yaml",
        "--defaults",
        Path(__file__).parent,
    ]
    main.app()
```

disrupted:
Fixes #317

### DIFF
--- a/kpops/utils/environment.py
+++ b/kpops/utils/environment.py
@@ -1,34 +1,16 @@
 import os
-import platform
 from collections import UserDict
-from collections.abc import Callable, Mapping
+from collections.abc import Mapping
 
 
 class Environment(UserDict[str, str]):
     def __init__(self, mapping: Mapping[str, str] | None = None, /, **kwargs) -> None:
-        _transformer: Callable[[str], str] = Environment.__get_transformation()
         env = os.environ
         if mapping is not None:
             env.update(mapping)
         if kwargs:
             env.update(**kwargs)
-        env = {_transformer(key): value for key, value in env.items()}
         super().__init__(env)
-
-    @staticmethod
-    def __key_camel_case_transform(key: str) -> str:
-        return key.lower()
-
-    @staticmethod
-    def __key_identity_transform(key: str) -> str:
-        return key
-
-    @staticmethod
-    def __get_transformation() -> Callable[[str], str]:
-        if platform.system() == "Windows":
-            return Environment.__key_camel_case_transform
-        else:
-            return Environment.__key_identity_transform
 
     def __getitem__(self, key: str) -> str:
         if key in self.data:
--- a/kpops/utils/environment.py
+++ b/kpops/utils/environment.py
@@ -1,21 +1,19 @@
 import os
 import platform
 from collections import UserDict
-from collections.abc import Callable
+from collections.abc import Callable, Mapping
 
 
-class Environment(UserDict):
-    def __init__(self, mapping=None, /, **kwargs) -> None:
-        transformation = Environment.__get_transformation()
+class Environment(UserDict[str, str]):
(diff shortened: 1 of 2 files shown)


tests/pipeline/resources/pipeline-with-env-defaults/defaults.yaml
```diff
@@ -34,6 +34,5 @@ kafka-connector:
     key.ignore: "false"
     linger.ms: "5000"
     max.buffered.records: "20000"
-    name: "sink-connector"
```
disrupted, line 37: this was actually a mistake that didn't error before this PR

kpops/utils/environment.py
```diff
@@ -1,36 +1,42 @@
...
-        if mapping is not None:
-            mapping = {transformation(key): value for key, value in mapping.items()}
-        else:
+class Environment(UserDict[str, str]):
+    """Internal environment wrapping OS environment."""
+
+    def __init__(
+        self, mapping: MutableMapping[str, str] | None = None, /, **kwargs
+    ) -> None:
+        self._global = os.environ
```
mkcode92, line 12: this still freezes the state of the environment, I think. Why don't you directly refer to `os.environ` ?

sujuka99 (approved):
LGTM

mkcode92:
maybe a stupid question, but why is this the `Environment` class still necessary at all? It seems to be an enriched os.environ, but this enrichment is never used. So it should behave like using os.environ everywhere you now use ENV. I think originally it was used to handle differently cased variables but this mechanism is now seemingly obsolete...

disrupted:
> maybe a stupid question, but why is this the `Environment` class still necessary at all? It seems to be an enriched os.environ, but this enrichment is never used. So it should behave like using os.environ everywhere you now use ENV. I think originally it was used to handle differently cased variables but this mechanism is now seemingly obsolete...

we have a set of _internal_ KPOps variables, such as the [`${pipeline_name}` variables](https://bakdata.github.io/kpops/2.0/user/core-concepts/variables/substitution/#pipeline-name-variables), for substitution. We want to store them in a local environment without polluting the OS environment. Likewise, we want environment variables to also work without replicating them to our internal environment – as we did previously. The `Environment` class is a thin abstraction combining both environments for convenience. It wraps the OS environment in a read-only way. Internal (lowercase) variables are only stored in memory without exporting.

mkcode92 (approved):

raminqaf (approved):

PR #370, opened 2023-10-12
Create HelmApp component

### LINKED ISSUE #364: Create separate HelmApp component
The builtin `KubernetesApp` component should be generic – not be specific to Helm. We should create a `HelmApp` component that inherits from `KubernetesApp`

disrupted:
Fixes #364


docs/docs/user/core-concepts/components/kubernetes-app.md
```diff
@@ -22,11 +22,11 @@ Can be used to deploy any app in Kubernetes using Helm, for example, a REST serv
 
 #### deploy
 
-Deploy using Helm.
+Do nothing.
 
 #### destroy
 
-Uninstall Helm release.
+Do nothing.
```
disrupted, line 29: question for team: 
should we leave the implementation of these empty or provide a default implementation that does a simple _apply_ / _delete_ using the Kubernetes API or kubectl?

sujuka99: Since the empty ones are already in `PipelineComponent` I'd say provide a default implementation.

kpops/components/base_components/helm_app.py
```diff
@@ -0,0 +1,166 @@
...
+    @override
+    def template(self) -> None:
+        stdout = self.helm.template(
+            self.helm_release_name,
+            self.helm_chart,
+            self.namespace,
+            self.to_helm_values(),
+            self.template_flags,
+        )
+        print(stdout)
```
sujuka99, line 106: Could we remove this from `PipelineComponent`? The docstring says it is specific to Helm

docs/docs/user/core-concepts/components/kubernetes-app.md
raminqaf: Unless we need to deploy anything with Kubernetes API, I would keep it like that. If the feature is needed I would add it in a separate PR. 
I propose using the Kubernetes API and avoid using the `kubectl` command (wrapper).

disrupted: agreed, could be separate PR once there's a need

sujuka99 (approved):
LGTM

kpops/components/base_components/helm_app.py
raminqaf, line 106: Didn't we want to change the print statement here with a better solution? 🤔 
Something like: https://github.com/bakdata/kpops/blob/main/kpops/pipeline_generator/pipeline.py#L249

PR #376, opened 2023-10-21
Fix early exit upon helm exit code 1

sujuka99:
> I am getting this error when I try to use dry-run: subprocess.CalledProcessError: Command '['helm', 'get', 'manifest', 'account-producer', '--namespace', 'my-namespace']' returned non-zero exit status 1. why? does he launch it first in dry-run ? when I do that in my commandline I get
To learn more, consult https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke
Error: release: not found

The [problem came from](https://github.com/bakdata/kpops/compare/2.0.9...2.0.10#diff-1ca7465daf9238dfe221304df636d96704ebf4935288b854a5c4c6b3ea9e3162R215) setting `subproces.run(..., check = True)` without handling the raised exception.


kpops/component_handlers/helm_wrapper/helm.py
```diff
@@ -210,13 +210,17 @@ def load_manifest(yaml_contents: str) -> Iterator[HelmTemplate]:
...
+        try:
+            process = subprocess.run(
+                command,
+                check=True,
+                capture_output=True,
+                text=True,
+            )
+        except subprocess.CalledProcessError as e:
+            Helm.parse_helm_command_stderr_output(e.stderr)
+            log.debug(e.stdout)
```
raminqaf, line 222: Do you think this is a debug log or an error one?

tests/pipeline/test_pipeline.py
```diff
@@ -16,6 +16,7 @@
 PIPELINE_BASE_DIR_PATH = RESOURCE_PATH.parent
 
 
+@pytest.mark.usefixtures("_mock_env")
```
raminqaf, line 19: Stay consistent with the annotation placement. Either put it on class or method.

examples/bakdata/atm-fraud-detection/config.yaml
```diff
@@ -4,9 +4,9 @@ topic_name_config:
   default_error_topic_name: "${pipeline_name}-${component_name}-dead-letter-topic"
   default_output_topic_name: "${pipeline_name}-${component_name}-topic"
 
-brokers: "http://k8kafka-cp-kafka-headless.kpops.svc.cluster.local:9092"
+brokers: "http://k8kafka-cp-kafka-headless.${NAMESPACE}.svc.cluster.local:9092"
```
disrupted, line 7: good idea!

kpops/component_handlers/helm_wrapper/helm.py
disrupted, line 222: Isn't the log duplicated now? And what's the advantage of the new behavior compared with `check=False` prior?

tests/conftest.py
```diff
@@ -0,0 +1,16 @@
+import os
+from collections.abc import Iterator
+from unittest import mock
+
+import pytest
+
+
+@pytest.fixture()
+def _mock_env() -> Iterator[None]:
```
disrupted, line 9: consider yielding the `os.environ`, so that the return value doesn't have to be `None`

tests/pipeline/test_example.py
```diff
@@ -8,6 +9,7 @@
 
 
 class TestExample:
+    @pytest.mark.usefixtures("_mock_env")
```
disrupted, line 12: nice!

disrupted (commented):
please keep consistent capitalisation in PR title, i.e. Helm

kpops/component_handlers/helm_wrapper/helm.py
sujuka99: There is no difference in the behavior. The code is a bit more explicit, but I don't see a clear advantage over `check=False`. Should I change it to that? It's gonna get rid of some code at least.

sujuka99: Debug. I kept the behavior identical to the one prior to introducing the bug. see https://github.com/bakdata/kpops/pull/376#discussion_r1368230652

sujuka99: Not duplicated because of the `return`

tests/pipeline/test_pipeline.py
sujuka99: I wasn't sure whether future tests in the `TestExample` class would benefit from the fixture, so I only enabled it for the 1 test inside.

Here there were too many tests, so I opted for decorating the whole class.

sujuka99: I'll move the one in `test_example`

raminqaf (approved):

PR #380, opened 2023-10-25
Refactor generate template for Python API usage

### LINKED ISSUE #103: `kpops generate --template` prints both `pipeline.yaml` and `helm template`
preceded by #32 

Problem: Currently it is possible to print both `pipeline.yaml` and `helm template` with one call to `generate` such as `kpops generate --template`. How can this be improved?

Takes:
- Print both with one call in `generate`, `--template` should only add an option to the `generate` functionality.
- Print only one with a call to `generate`, printing both is overwhelming and the user is unlikely to need them both printed with one call.
- Make `--template` into `kpops template`, no need to have it together with `generate`.

disrupted:
`template` renamed to ~~`render`~~ `manifest` as it's no longer strictly tied to Helm template, can be used to render final resources of any pipeline component, e.g. Kubernetes manifest
Extract `kpops generate --template` into separate ~~`kpops render`~~ `kpops manifest` command 
Return Kubernetes manifest as mapping for use with Python API

Closes #103


kpops/component_handlers/kubernetes/model.py
```diff
@@ -0,0 +1,28 @@
...
+    from typing_extensions import Self
+
+Json: TypeAlias = dict[str, "Json"] | list["Json"] | str | int | float | bool | None
+
+
+class KubernetesManifest(UserDict[str, Json]):
+    """Representation of a Kubernetes API object as YAML/JSON mapping."""
+
+    @classmethod
+    def from_yaml(cls, /, content: str) -> Iterator[Self]:
```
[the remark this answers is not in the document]
disrupted: it indicates that `content` is a positional-only argument.

docs/docs/user/references/cli-commands.md
```diff
@@ -18,9 +18,10 @@ $ kpops [OPTIONS] COMMAND [ARGS]...
 * `clean`: Clean pipeline steps
 * `deploy`: Deploy pipeline steps
 * `destroy`: Destroy pipeline steps
-* `generate`: Enriches pipelines steps with defaults.
+* `generate`: Enrich pipeline steps with defaults.
+* `render`: In addition to generate, render final...
```
disrupted, line 22: fix: do not truncate

tests/pipeline/test_manifest.py
```diff
@@ -0,0 +1,341 @@
...
+            RESOURCE_PATH / "custom-config/pipeline.yaml",
+            defaults=RESOURCE_PATH / "no-topics-defaults",
+            output=False,
+            environment="development",
+        )
+        assert len(steps) == 2
+        manifests = steps[0]
+        assert len(manifests) == 1
+        assert isinstance(manifests[0], KubernetesManifest)
+        assert manifests[0] == {
```
raminqaf, line 115: How about we read the manifest content from a YAML file? IMO, it makes the test file smaller and easier to read.

disrupted: good idea! Probably we could even use a snapshot to benefit from automatic re-generation when we make changes

docs/docs/user/references/cli-commands.md
disrupted: @sujuka99 do you know if it's supposed to be like this? would be better to see the full help for each of the commands

disrupted: ok, thanks for checking. If it's outside our control then so be it

sujuka99: See https://github.com/pallets/click/issues/486

Could you try putting one of the `help` strings in `short_help`?

disrupted: Great! I've added a separate `short_help` for the command now, which should suffice

sujuka99: Ah, I was thinking about putting the whole `help` in `short_help`

disrupted: yeah, I got that. But then I thought short help is named as such for a reason, so I gave it a shorter string. The full help is then visible for the command `kpops manifest --help`. I did the same for `kpops generate` now and I think it looks nice and clean. What do you think?

sujuka99: Ah got it now. I think I was mixing up the 2 descriptions in the docs. I agree, it looks quite nice.

sujuka99 (approved):

PR #347, opened 2023-08-30
Migrate to Pydantic v2

sujuka99:
closes #341

PR #387, opened 2023-11-06
Define custom components module & pipeline base dir globally

disrupted:
Depends on #347 
for added convenience, ability to set pipeline base dir and components module globally for the project. therefore it won't be necessary to specify these with each invocation
will be used for automatic schema generation (editor integration)


docs/docs/resources/variables/config_env_vars.env
disrupted, line 1: Depends on #384

disrupted:
should we remove those two options from the CLI now that they're global settings?

kpops/cli/main.py
```diff
@@ -28,6 +28,8 @@
 
     from kpops.components.base_components import PipelineComponent
 
+ENV_PREFIX = KpopsConfig.model_config.get("env_prefix")
```
sujuka99, line 31: This is already [defined in the config](https://github.com/bakdata/kpops/blob/v3/kpops/config.py#L17), you could import the constant from there

kpops/cli/main.py
```diff
@@ -377,7 +423,7 @@ def reset(
         component.reset(dry_run)
 
 
-@app.command(help="Clean pipeline steps")  # pyright: ignore[reportGeneralTypeIssues] https://github.com/rec/dtyper/issues/8
+@app.command(help="Clean pipeline steps")  # pyright: ignore[reportGeneralTypeIssues] https://github.com/rec/dtyper/issues/8eneralTypeIssues] https://github.com/rec/dtyper/issues/8
```
sujuka99, line 426: typo

sujuka99 (changes requested):

sujuka99:
> should we remove those two options from the CLI now that they're global settings?

Sounds good

raminqaf:
Since this is a breaking change (by removing the extra CLI parameters) please label it as `breaking-change`. One more thing: Please address the changes in the migration guide 😉

sujuka99 (approved):

docs/docs/user/migration-guide/v2-v3.md
```diff
@@ -81,3 +81,16 @@ The `--config` flag in the CLI now points to the directory that contains `config
     enabled: true
     url: "http://my-custom-sr.url:8081"
 ```
+
+## [Define custom components module & pipeline base dir globally](https://github.com/bakdata/kpops/pull/387)
+
+The options for a custom `components_module` and `pipeline_base_dir` are now global settings, defined in `config.yaml`. The previous CLI parameters have been removed.
```
raminqaf: very tiny:
let's make the statement `The previous CLI parameters have been removed.` bold with a warning
https://squidfunk.github.io/mkdocs-material/reference/admonitions/#+type:warning
WDYT?

disrupted: good idea! I'll check to see if I did it correctly

raminqaf (approved):

PR #391, opened 2023-12-01
Allow overriding config files

### LINKED ISSUE #153: Allow different config.yaml per environment
Currently, there is only a global config.yaml. Similar to defaults and pipeline files, it would be good to allow config overrides per environment by creating files such as config_development.yaml for development specific configurations

sujuka99:
closes #153


docs/docs/user/migration-guide/v2-v3.md
```diff
@@ -63,3 +63,17 @@ steps:
       command: deploy --execute
       # ...
 ```
+
+## [Allow overriding config files](https://github.com/bakdata/kpops/pull/391)
+
+`environment` is no longer specified in `config.yaml`. Instead, it can be either set via the cli flag `--environment` or with the environment variable `KPOPS_ENVIRONMENT`.
```
raminqaf, line 69: Please mention that the `--config` is now pointing at the directory and not the file anymore

kpops/pipeline_generator/pipeline.py
```diff
@@ -137,13 +138,21 @@ def load_from_yaml(
...
+        Pipeline.set_environment_name(environment)
 
         main_content = load_yaml_file(path, substitution=ENV)
         if not isinstance(main_content, list):
             msg = f"The pipeline definition {path} should contain a list of components"
             raise TypeError(msg)
         env_content = []
-        if (env_file := Pipeline.pipeline_filename_environment(path, config)).exists():
+        if (
+            ENV.get("environment") is not None
```
raminqaf, line 149: You can only do this:
```suggestion
            environment is not None
```
Or is there a reason you read the ENV dictionary again?

kpops/pipeline_generator/pipeline.py
```diff
@@ -304,14 +313,14 @@ def validate(self) -> None:
         self.components.validate_unique_names()
 
     @staticmethod
-    def pipeline_filename_environment(path: Path, config: KpopsConfig) -> Path:
+    def pipeline_filename_environment(path: Path, environment: str) -> Path:
```
raminqaf, line 316: `path` is maybe not the best name anymore. Let's update it. How about `pipeline_path`?

kpops/utils/pydantic.py
```diff
@@ -111,20 +113,48 @@ def json_schema_extra(schema: dict[str, Any], model: type[BaseModel]) -> None:
...
+        # Check if path to settings yaml definition is valid
+        if not self.path_to_settings.exists():
+            msg = f"Path to config directory {self.path_to_settings} does not exist."
+            raise ValueError(msg)
+        elif self.path_to_settings.is_file():
+            msg = f"Path to config directory {self.path_to_settings} must point to a directory."
+            raise ValueError(msg)
+        default_settings = self.__load_settings(
+            self.path_to_settings / f"{self.settings_file_base_name}.yaml"
+        )
```
raminqaf, line 133: Extract to method and call it `check_if_path_config_yaml_is_valid`

tests/conftest.py
```diff
@@ -14,3 +16,9 @@ def mock_env() -> Iterator[os._Environ[str]]:
     """
     with mock.patch.dict(os.environ, clear=True):
         yield os.environ
+
+
+@pytest.fixture()
+def _load_yaml_file_clear_cache() -> Iterator[None]:
```
raminqaf, line 22:
```suggestion
def clear_load_yaml_file_cache() -> Iterator[None]:
```

raminqaf (changes requested):

kpops/pipeline_generator/pipeline.py
sujuka99: Agreed

tests/conftest.py
sujuka99: It's a [convention](https://docs.astral.sh/ruff/rules/pytest-missing-fixture-name-underscore/)

docs/docs/user/migration-guide/v2-v3.md
sujuka99: Added some text, lmk what you think

kpops/utils/pydantic.py
sujuka99: Only extracted the 2 `if`s and chose a different name. Do you agree with the change?

kpops/cli/main.py
```diff
@@ -207,17 +217,22 @@ def log_action(action: str, pipeline_component: PipelineComponent):
...
-    config: Path, defaults: Optional[Path], verbose: bool, dotenv: Optional[list[Path]]
+    config: Path,
+    defaults: Optional[Path],
+    verbose: bool,
+    dotenv: Optional[list[Path]],
+    environment: Optional[str],
 ) -> KpopsConfig:
     setup_logging_level(verbose)
-    YamlConfigSettingsSource.path_to_config = config
+    YamlConfigSettingsSource.path_to_settings = config
```
raminqaf, line 227: I like this name more
```suggestion
    YamlConfigSettingsSource.path_to_config = config
```

kpops/utils/pydantic.py
```diff
@@ -111,20 +113,55 @@ def json_schema_extra(schema: dict[str, Any], model: type[BaseModel]) -> None:
 class YamlConfigSettingsSource(PydanticBaseSettingsSource):
     """Loads variables from a YAML file at the project's root."""
 
-    path_to_config = Path("config.yaml")
+    log = logging.getLogger()
+
+    path_to_settings = Path()
```
raminqaf, line 118: I think we should stick with `config` and not `setting`
```suggestion
    path_to_config = Path()
```

kpops/utils/pydantic.py
```diff
@@ -111,20 +113,55 @@ def json_schema_extra(schema: dict[str, Any], model: type[BaseModel]) -> None:
...
+    settings_file_base_name = "config"
```
raminqaf, line 119:
```suggestion
    config_file_base_name = "config"
```

tests/pipeline/test_pipeline.py
raminqaf, line 1: Do we have a test where the user doesn't pass in the `--environment` and doesn't have a `config.yaml` file?

sujuka99: Yes, `test_read_from_component` for example.

tests/conftest.py
sujuka99: Removed the leading underscore and removed the `PT` rules from `ruff`.

[reason](https://github.com/astral-sh/ruff/issues/8796)

kpops/utils/pydantic.py
```diff
@@ -111,20 +113,59 @@ def json_schema_extra(schema: dict[str, Any], model: type[BaseModel]) -> None:
...
+        :raises ValueError: Path to config directory does not exist.
+        :raises ValueError: Path to config directory must point to a directory.
+        """
+        if not self.path_to_config.exists():
+            msg = f"Path to config directory {self.path_to_config} does not exist."
+            raise ValueError(msg)
+        elif self.path_to_config.is_file():
+            msg = f"Path to config directory {self.path_to_config} must point to a directory."
+            raise ValueError(msg)
+
```
raminqaf, line 150: Do we have tests for this?

disrupted: why not simply
```suggestion
    def is_config_path_valid(self) -> None:
        """Check if the specified config directory exists.

        :raises ValueError: Config directory does not exist.
        """
        if not self.path_to_config.is_dir():
            msg = f"Config directory {self.path_to_config} does not exist."
            raise ValueError(msg)
```

kpops/utils/pydantic.py
```diff
@@ -111,20 +113,59 @@ def json_schema_extra(schema: dict[str, Any], model: type[BaseModel]) -> None:
 class YamlConfigSettingsSource(PydanticBaseSettingsSource):
     """Loads variables from a YAML file at the project's root."""
 
-    path_to_config = Path("config.yaml")
+    log = logging.getLogger()
+
+    path_to_config = Path()
```
disrupted, line 118:
```suggestion
    config_dir = Path()
```

kpops/utils/pydantic.py
sujuka99: > Do we have tests for this?

I added one now.

docs/docs/user/migration-guide/v2-v3.md
```diff
@@ -63,3 +63,21 @@ steps:
       command: deploy --execute
       # ...
 ```
+
+## [Allow overriding config files](https://github.com/bakdata/kpops/pull/391)
+
+Specifying the environment is no longer mandatory. If not defined, only the default files will be used.
```
disrupted, line 69:
```suggestion
Specifying the environment is no longer mandatory. If not defined, only the global files will be used.
```

perhaps like this to avoid confusion with `defaults.yaml`?

docs/docs/user/migration-guide/v2-v3.md
```diff
@@ -63,3 +63,21 @@ steps:
...
+
+`environment` is no longer specified in `config.yaml`. Instead, it can be either set via the cli flag `--environment` or with the environment variable `KPOPS_ENVIRONMENT`.
```
disrupted, line 71:
```suggestion
`environment` is no longer specified in `config.yaml`. Instead, it can be either set via the CLI flag `--environment` or with the environment variable `KPOPS_ENVIRONMENT`.
```

kpops/cli/main.py
```diff
@@ -63,13 +63,13 @@
 )
 
 CONFIG_PATH_OPTION: Path = typer.Option(
-    default=Path("config.yaml"),
+    default=Path(),
     exists=False,
```
disrupted, line 67: should we enable this check?

sujuka99: Sure

kpops/utils/pydantic.py
```diff
@@ -111,20 +113,55 @@ def json_schema_extra(schema: dict[str, Any], model: type[BaseModel]) -> None:
...
+            self.load_config(
+                self.config_dir
+                / f"{self.config_file_base_name}_{self.environment}.yaml"
+            )
+            if self.environment
+            else {}
+        )
+        self.config = update_nested_pair(env_config, default_config)
+
+    def is_config_dir_valid(self) -> None:
```
disrupted, line 138: methods named `is_<some-condition>` usually indicate a boolean return value. Since we don't return anything here we should consider naming it differently, e.g. `verify_config_dir_valid`

sujuka99: How about `check_config_dir_exists`?

kpops/cli/main.py
disrupted: no, I don't know. Which output do you get with it enabled if the directory doesn't exist?

disrupted: I am fine with leaving it disabled if that improves the user experience because of our custom error, but then we should add a comment here explaining why it's disabled.

sujuka99: enabled it and removed the custom check

tests/pipeline/test_pipeline.py
```diff
@@ -583,25 +583,6 @@ def test_env_specific_config_env_def_in_cli(
             == expected_url
         )
 
-    def test_config_dir_doesnt_exist(self):
```
disrupted: you could keep the test and assert a non-zero exit code

docs/docs/resources/variables/config_env_vars.md
```diff
@@ -1,17 +1,16 @@
...
+|------------------------|----------------------------------------------------------------------------------------------------------------------------------|--------|------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------|
+|defaults_path           |.                                                                                                                                 |False   |The path to the folder containing the defaults.yaml file and the environment defaults files. Paths can either be absolute or relative to `config.yaml`|defaults_path           |
+|kafka_brokers           |PydanticUndefined                                                                                                                 |False   |The comma separated Kafka brokers address.                                                                                                            |kafka_brokers           |
```
disrupted, line 6: this probably still needs the docs fix?

sujuka99: updated it just now

tests/conftest.py
```diff
@@ -14,3 +16,9 @@ def mock_env() -> Iterator[os._Environ[str]]:
     """
     with mock.patch.dict(os.environ, clear=True):
         yield os.environ
+
+
+@pytest.fixture()
+def load_yaml_file_clear_cache() -> Iterator[None]:
+    yield
+    load_yaml_file.cache.clear()
```
disrupted, line 24: nice one!

disrupted (approved):

PR #397, opened 2024-01-02
Refactor Helm `nameOverride`

### LINKED ISSUE #327: Remove `name_override` workaround for Helm apps
at the moment we have a builtin workaround to [set `name_override` for all Helm apps](https://github.com/bakdata/kpops/blob/f0b198f3491810b443ef1cecde0facccae68bed5/kpops/pipeline_generator/pipeline.py#L76) although there's also the [`helm_release_name` property](https://github.com/bakdata/kpops/blob/f0b198f3491810b443ef1cecde0facccae68bed5/kpops/components/base_components/kubernetes_app.py#L107). This could be an issue with streams-bootstrap Helm charts and might need to be fixed there first.

disrupted:
Fixes #327 

Remove previous workaround for setting Helm app `nameOverride` in favor of a more robust implementation in `HelmApp`

### DIFF
--- a/tests/pipeline/snapshots/snap_test_pipeline.py
+++ b/tests/pipeline/snapshots/snap_test_pipeline.py
@@ -10,7 +10,6 @@ snapshots = Snapshot()
 snapshots['TestPipeline.test_default_config test-pipeline'] = [
     {
         'app': {
-            'nameOverride': 'resources-custom-config-app1',
             'resources': {
                 'limits': {
                     'memory': '2G'
@@ -58,7 +57,6 @@ snapshots['TestPipeline.test_default_config test-pipeline'] = [
             'labels': {
                 'pipeline': 'resources-custom-config'
             },
-            'nameOverride': 'resources-custom-config-app2',
             'streams': {
                 'brokers': 'http://k8kafka-cp-kafka-headless.kpops.svc.cluster.local:9092',
                 'errorTopic': 'resources-custom-config-app2-error',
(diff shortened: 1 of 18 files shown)


tests/pipeline/snapshots/snap_test_pipeline.py
```diff
@@ -10,7 +10,6 @@
 snapshots['TestPipeline.test_default_config test-pipeline'] = [
     {
         'app': {
-            'nameOverride': 'resources-custom-config-app1',
```
raminqaf, line 13: Why are these removed? If we don't set anything streamsbootsrap will use the defined value in `value.yaml` see [here](https://github.com/bakdata/streams-bootstrap/blob/master/charts/streams-app/values.yaml#L1)

Shouldn't this be `full_name` if not set? according to your code above
```python
        if self.app.name_override is None:
            self.app.name_override = self.full_name
```

disrupted: it's only set in the [`to_helm_values`](https://github.com/bakdata/kpops/blob/1c4e3134b99de5ffb9964a4c9b16d7b70b620e83/kpops/components/base_components/helm_app.py#L161) method
it's invoked during `template`, `deploy`, etc.
Unless the user sets it in their pipeline definition, it's not part of `generate` output

raminqaf:
fixes #327 ?

raminqaf (approved):

PR #398, opened 2024-01-02
Refactor streams-bootstrap cleanup jobs as individual HelmApp

disrupted:
Depends on #397 

extracted the cleanup job needed for streams-bootstrap producer and streams apps to its own HelmApp nested inside the parent component

- [x] Create docs for `StreamsBootstrap` base
- [x] Update KPOps component inheritance graph
- [x] Update docs gen for multiple component inheritance (#406)

PR #400, opened 2024-01-04
Refactor Kafka Connector resetter as individual HelmApp

disrupted:
Depends on #398


sujuka99 (approved):

kpops/components/base_components/kafka_connector.py
```diff
@@ -90,47 +148,27 @@ def connector_config_should_have_component_name(
...
+            **self.model_dump(
+                exclude={"_resetter", "resetter_values", "resetter_namespace", "app"}
+            ),
+            app=KafkaConnectorResetterValues(
+                connector_type=self._connector_type.value,
+                config=KafkaConnectorResetterConfig(
+                    connector=self.full_name,
+                    brokers=self.config.kafka_brokers,
+                ),
+                **self.resetter_values.model_dump(),
```
raminqaf, line 170: Will the nameOverride value be in the dump here?
```
name_override=self.full_name + "-clean",
```

kpops/components/base_components/kafka_connector.py
```diff
@@ -25,50 +20,113 @@
...
 from kpops.components.base_components.pipeline_component import PipelineComponent
 from kpops.utils.colorify import magentaify
 from kpops.utils.docstring import describe_attr
 
 log = logging.getLogger("KafkaConnector")
 
 
-class KafkaConnector(PipelineComponent, ABC):
-    """Base class for all Kafka connectors.
+class KafkaConnectorResetter(HelmApp):
```
raminqaf, line 32: I prefer that the kafka_connector.py file starts with the KafkaConnector class, then the sink/source connectors, and, in the end, with the Kafka ConnectorResetter class. WDYT?

tests/components/test_kafka_connector.py
```diff
@@ -110,21 +119,10 @@ def test_connector_config_name_override(
...
-            handlers=handlers,
-            app=connector_config,
-            namespace="test-namespace",
-        )
+    def test_resetter_release_name(self, connector: KafkaConnector):
         assert connector.app.name == CONNECTOR_FULL_NAME
-        assert connector._resetter_release_name == CONNECTOR_CLEAN_RELEASE_NAME
+        # resetter = connector._resetter  # FIXME
+        # assert isinstance(resetter, KafkaConnectorResetter)
+        # assert connector._resetter.helm_release_name == CONNECTOR_CLEAN_RELEASE_NAME
```
raminqaf, line 128: Why can't this be fixed?

tests/components/test_kafka_sink_connector.py
```diff
@@ -52,7 +53,7 @@ def connector(
             config=config,
             handlers=handlers,
             app=connector_config,
-            namespace="test-namespace",
+            resetter_namespace="test-namespace",
```
raminqaf, line 56: Extract const

docs/docs/user/migration-guide/v2-v3.md
```diff
@@ -56,6 +56,16 @@ Previously the default `KafkaApp` component configured the [streams-bootstrap](h
     version: ...
 ```
 
+## [Refactor Kafka Connector resetter as individual HelmApp](https://github.com/bakdata/kpops/pull/400)
+
```
raminqaf, line 60: Just add a brief description here.

tests/pipeline/resources/defaults.yaml
```diff
@@ -104,8 +104,7 @@ should-inflate:
           retention.ms: "-1"
 
 kafka-connector:
-  name: "sink-connector"
-  namespace: "example-namespace"
+  name: sink-connector
```
raminqaf, line 107: Change the `kubernetes-app` in this defaults.yaml to `helm-app`

tests/components/test_kafka_connector.py
disrupted: I have no idea, as it works in the other tests, such as Kafka sink connector

kpops/components/base_components/kafka_connector.py
raminqaf: Is it possible now to define in your pipeline.yaml a `KafkaConnectorResetter`? Because it is a HelmApp, and if I call KPOps to deploy, will it deploy it?

disrupted: I agree, for some reason it slightly changes the pipeline schema when I reorder the classes

```diff
diff --git a/docs/docs/schema/pipeline.json b/docs/docs/schema/pipeline.json
index 186863f6..aebe7127 100644
--- a/docs/docs/schema/pipeline.json
+++ b/docs/docs/schema/pipeline.json
@@ -306,13 +306,13 @@
                 },
                 "type": {
                     "const": "kafka-sink-connector",
+                    "default": "kafka-sink-connector",
                     "title": "Type"
                 }
             },
             "required": [
                 "name",
-                "app",
-                "type"
+                "app"
             ],
             "title": "KafkaSinkConnector",
             "type": "object"
@@ -401,13 +401,13 @@
                 },
                 "type": {
                     "const": "kafka-source-connector",
+                    "default": "kafka-source-connector",
                     "title": "Type"
                 }
             },
             "required": [
                 "name",
-                "app",
-                "type"
+                "app"
             ],
             "title": "KafkaSourceConnector",
             "type": "object"
```

I don't know why it differs from the other components then, would prefer to have it consistent.

kpops/components/base_components/kafka_connector.py
disrupted: it's assigned the standard way
1. `full_name` is overridden to include the suffix https://github.com/bakdata/kpops/blob/f9aae5cde7f0c6e4c866a6be9f5cb59fe91a0a78/kpops/components/base_components/kafka_connector.py#L52-L55
3. `HelmApp` sets the `name_override` https://github.com/bakdata/kpops/blob/f9aae5cde7f0c6e4c866a6be9f5cb59fe91a0a78/kpops/components/base_components/helm_app.py#L176-L177

It's verified by the tests, e.g. https://github.com/bakdata/kpops/blob/f9aae5cde7f0c6e4c866a6be9f5cb59fe91a0a78/tests/components/test_kafka_sink_connector.py#L234

kpops/components/base_components/kafka_connector.py
disrupted: We _could_ technically make it possible to define `KafkaConnectorResetter` in the pipeline, but what would be the point as it's not meant to be used standalone. Or is there a use-case?
That's why it's excluded from the public components

tests/pipeline/resources/defaults.yaml
disrupted: why?

disrupted: I renamed kafka-connector → kafka-sink-connector though

tests/components/test_kafka_connector.py
disrupted: I get an `AttributeError` trying to access `connector._resetter`, it doesn't even show up in `dir(connector)`. Do you have an idea?

kpops/components/base_components/kafka_connector.py
raminqaf: No, there isn't any use-case I can think of so far. I just wanted to know if the scenario is possible or not.

tests/components/test_kafka_connector.py
raminqaf: I think the problem is with the `_connector_type`. Whenever it wants to compute the _resetter filed it creates this:
```python
        kafka_connector_resetter = KafkaConnectorResetter(
            config=self.config,
            handlers=self.handlers,
            **kwargs,
            **self.model_dump(
                exclude={"_resetter", "resetter_values", "resetter_namespace", "app"}
            ),
            app=KafkaConnectorResetterValues(
                connector_type=self._connector_type.value, # <------ Here it tells me that the self._connectot_type does not exist
                config=KafkaConnectorResetterConfig(
                    connector=self.full_name,
                    brokers=self.config.kafka_brokers,
                ),
                **self.resetter_values.model_dump(),
            ),
        )
```

tests/pipeline/resources/defaults.yaml
disrupted: but it doesn't contain anything Helm-specific in this case, only `name` and `namespace`. I moved `name` to `pipeline-component` defaults now, but `namespace` is fine for all `kubernetes-app` imo

tests/components/test_kafka_connector.py
disrupted: oh nice, thanks!! that makes total sense

tests/pipeline/resources/defaults.yaml
raminqaf: Okay makes sense, I thought that the kubernetes-app is a more abstract type that is better to replace with a concrete implementation (like helm-app).

raminqaf (approved):

PR #405, opened 2024-01-10
Add steps filter to `kpops generate`

### LINKED ISSUE #110: Add filtering to `generate` command for printing the `pipeline.yaml`

disrupted:
Closes #110


raminqaf (approved):
It LGTM!

PR #418, opened 2024-01-16
Fix wrong Helm release name character limit

disrupted:


sujuka99:
Is it breaking? Please note in the migration guide that the limit is 53 instead of the currently documented 52.

disrupted:
Good point, technically I would consider this more of a fix, but it could require deleting previous Helm releases. Added the breaking label per your suggestion. Migration guide should be fixed now.

sujuka99 (approved):
LGTM!

PR #430, opened 2024-01-19
Update poetry publish workflow version to latest

raminqaf:
image

There is a way to increase the timeout:
https://python-poetry.org/docs/faq/#my-requests-are-timing-out

### Edit
The above issue is still not solved. Even after setting the env variable, the timeout does not increase. Related issue for TestPyPI: https://github.com/pypi/support/issues/3396

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -3,14 +3,11 @@ on:
   push:
     tags:
       - '*'
-env:
-  POETRY_REQUESTS_TIMEOUT: 120
 
 jobs:
   publish:
-    uses: bakdata/ci-templates/.github/workflows/python-poetry-publish-pypi.yaml@1.40.2
+    uses: bakdata/ci-templates/.github/workflows/python-poetry-publish-pypi.yaml@main
     with:
       publish-to-test: false
-      poetry-version: "1.7.1"
     secrets:
       pypi-token: "${{ secrets.PYPI_TOKEN }}"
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yaml
```diff
@@ -6,9 +6,8 @@ on:
 
 jobs:
   publish:
-    uses: bakdata/ci-templates/.github/workflows/python-poetry-publish-pypi.yaml@1.25.3
+    uses: bakdata/ci-templates/.github/workflows/python-poetry-publish-pypi.yaml@1.40.4
     with:
       publish-to-test: false
-      poetry-version: "1.5.1"
```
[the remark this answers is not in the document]
raminqaf: done

.github/workflows/publish.yaml
```diff
@@ -9,5 +9,6 @@ jobs:
     uses: bakdata/ci-templates/.github/workflows/python-poetry-publish-pypi.yaml@1.40.4
     with:
       publish-to-test: false
+      poetry-version: "1.4.1"
```
disrupted, line 12: I just saw the timeout env var was only added in Poetry 1.5.0. https://github.com/python-poetry/poetry/pull/7081
Does it make sense to update to the latest version?

raminqaf: It make sense to have that above 1.5.0. But we still have the issue with [poetry and windows](https://github.com/bakdata/kpops/pull/430/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddL31). I can give it a try with the newest version (1.7.1)

raminqaf: It seems that the problem with the windows runners is not occurring. We can keep the 1.7.1

PR #434, opened 2024-01-22
Fix Helm diff output

disrupted:
Fix `!!python/object:kpops.component_handlers.kubernetes.model.KubernetesManifest` in Helm diff


sujuka99 (approved):

irux:
Do you have any text of what was before the fix?

disrupted:
> Do you have any text of what was before the fix?

before the diff looked incorrect like this

```diff
!!python/object:kpops.component_handlers.kubernetes.model.KubernetesManifest
+ data:
+   a: 1
```

with the fix it looks like this again

```diff
+ a: 1
```

irux (approved):

PR #435, opened 2024-01-24
Simplify loading of defaults

disrupted:


docs/docs/resources/pipeline-components/dependencies/kpops_structure.yaml
```diff
@@ -85,36 +85,43 @@ kpops_components_inheritance_ref:
     parents:
     - kubernetes-app
     - pipeline-component
+    - base-defaults-component
```
disrupted, line 88: it doesn't seem to cause any side effects to include the base component

sujuka99 (approved):
Nice!

PR #438, opened 2024-01-26
Distribute defaults across multiple files

### LINKED ISSUE #55: Defaults should be distributed across multiple files
We should leverage our nested folder structure and allow placing defaults.yaml in each folder. These should be automatically loaded given a pipeline path. No `--defaults` parameter should be necessary then. Defaults should be merged from top to bottom, i.e., the up-most level is used as a baseline and child defaults are merged afterwards. This means that conflicting keys defined in more specific locations override keys in more general locations.

raminqaf:
Closes #55


kpops/components/base_components/base_defaults_component.py
```diff
@@ -122,25 +121,29 @@ def extend_with_defaults(cls, config: KpopsConfig, **kwargs: Any) -> dict[str, A
...
+            + typer.style(cls.type, bold=True, underline=True)
         )
         return update_nested_pair(kwargs, defaults)
 
     @classmethod
     def load_defaults(cls, *defaults_file_paths: Path) -> dict[str, Any]:
         """Resolve component-specific defaults including environment defaults.
 
-        :param *defaults_file_paths: Path to `defaults.yaml`, ordered from lowest to highest priority, i.e. `defaults.yaml`, `defaults_{environment}`.yaml
+        :param *defaults_file_paths: Path to `defaults.yaml`, ordered from highest to lowest priority,
```
disrupted, line 145: breaking change?

also in the other PR you removed the star operator (`*`) from the docs. I am not an expert at pydocs, not sure what's correct, just wanted to point it out

raminqaf: Yes breaking change! 🙂 
We are totally removing the `--defaults` flag from the CLI, and the `default-path` from the config!
It was the pyCharm doc string checker giving me the warning fix and I took it!

disrupted (commented):
this calls for a new migration guide :)

kpops/pipeline.py
```diff
@@ -76,7 +78,7 @@ def add(self, component: PipelineComponent) -> None:
     def __bool__(self) -> bool:
         return bool(self.components)
 
-    def __iter__(self) -> Iterator[PipelineComponent]:
+    def __iter__(self) -> Iterator[PipelineComponent]:  # pyright: ignore [reportIncompatibleMethodOverride]
```
disrupted, line 81: do you know where this comes from? does Pydantic define this? I never had it in Pyright before

kpops/components/base_components/base_defaults_component.py
```diff
@@ -193,31 +193,44 @@ def defaults_from_yaml(path: Path, key: str) -> dict:
...
+            default_paths.append(environment_default_file_path)
+
+        defaults_yaml_path = path.parent / f"{config.defaults_filename_prefix}.yaml"
+        if defaults_yaml_path.is_file():
+            default_paths.append(defaults_yaml_path)
+
+        path = path.parent
 
-    return main_default_file_path, environment_default_file_path
+    return default_paths
```
sujuka99, line 233: Does it make sense to extract the logic that looks for the files into a function which we could later use for the config?

sujuka99 (approved):

docs/docs/user/migration-guide/v3-v4.md
```diff
@@ -0,0 +1,32 @@
+# Migrate from V2 to V3
+
+## [Distribute defaults across multiple files](https://github.com/bakdata/kpops/pull/438/)
+
+The default files are now read from where the pipeline path is defined. The `--defaults` flag is removed. Imagine the following folder structure:
```
disrupted, line 5: In the first sentence, I'd add a brief explanation that multiple defaults files are now possible, before going into detail

docs/docs/user/migration-guide/v3-v4.md
```diff
@@ -0,0 +1,32 @@
+# Migrate from V2 to V3
```
disrupted, line 1:
```suggestion
# Migrate from V3 to V4
```

docs/docs/user/migration-guide/v3-v4.md
```diff
@@ -0,0 +1,32 @@
...
+       └── pipeline-deep
+           ├── defaults.yaml
+           └── pipeline.yaml
+```
+
+The `pipeline-base-dir` is set to `pipelines`. Now if we generate this pipeline with the following command:
+
+```bash
+kpops generate \
+      --config . \
```
disrupted, line 21: necessary?

docs/docs/user/migration-guide/v3-v4.md
```diff
@@ -0,0 +1,32 @@
...
+      ./pipelines/distributed-defaults/pipeline-deep/pipeline.yaml
+```
+
+The defaults would be picked in the following order (high to low priority):
+
+- `./pipelines/distributed-defaults/pipeline-deep/defaults.yaml`
+- `./pipelines/distributed-defaults/defaults_dev.yaml`
+- `./pipelines/distributed-defaults/defaults.yaml`
+
+The deepest default.yaml file (i.e., the closest one to the `pipeline.yaml`) overwrites the higher-level defaults' values.
```
disrupted, line 32:
```suggestion
The deepest defaults.yaml file (i.e., the closest one to the `pipeline.yaml`) overwrites the higher-level defaults' values.
```

kpops/components/base_components/base_defaults_component.py
```diff
@@ -128,25 +127,26 @@ def extend_with_defaults(cls, config: KpopsConfig, **kwargs: Any) -> dict[str, A
...
 
-        log.debug(
-            typer.style(
-                "Enriching component of type ", fg=typer.colors.GREEN, bold=False
-            )
-            + typer.style(cls.type, fg=typer.colors.GREEN, bold=True, underline=True)
+        env_pipeline_path = ENV.get("pipeline_path")
+        if not env_pipeline_path:
+            env_pipeline_path = Path()
+        pipeline_path = Path(env_pipeline_path)
```
disrupted, line 133: possible to shorten this a bit?
```suggestion
        pipeline_path = Path(ENV.get("pipeline_path", ""))
```

kpops/components/base_components/base_defaults_component.py
```diff
@@ -193,31 +193,44 @@ def defaults_from_yaml(path: Path, key: str) -> dict:
...
-    environment_default_file_path = (
-        defaults_dir
-        / Path(f"{config.defaults_filename_prefix}_{environment}").with_suffix(".yaml")
-        if environment is not None
-        else main_default_file_path
-    )
+    default_paths = []
+
+    if not pipeline_path.is_file():
+        message = f"No pipeline.yaml found in directory: {pipeline_path}"
```
disrupted, line 212: pipeline_path is the filepath itself, not a directory

tests/components/test_base_defaults_component.py
```diff
@@ -202,3 +204,62 @@ def test_merge_defaults(self, config: KpopsConfig, handlers: ComponentHandlers):
...
+                    ),
+                    Path(f"{RESOURCES_PATH}/defaults.yaml"),
+                ],
+            ),
+        ],
+    )
+    def test_get_defaults_file_paths(
+        self,
+        pipeline_path: Path,
+        environment: str,
```
disrupted, line 254:
```suggestion
        environment: str | None,
```
please add a test for the `None` case

tests/components/test_helm_app.py
```diff
@@ -160,7 +163,7 @@ async def test_should_deploy_app_with_local_helm_chart(
         app_values: HelmAppValues,
     ):
         class AppWithLocalChart(HelmApp):
-            repo_config: None = None
+            repo_config: HelmRepoConfig | None = None
```
disrupted, line 166: judging by the class name, I believe it was set to `None` on purpose

tests/components/test_streams_app.py
```diff
@@ -418,8 +422,6 @@ async def test_reset_when_dry_run_is_false(
                     STREAMS_APP_CLEAN_RELEASE_NAME,
                     dry_run,
                 ),
-                ANY,  # __bool__
```
disrupted, line 421: nice! does it pass without those now?

tests/utils/test_dict_ops.py
```diff
@@ -3,11 +3,56 @@
...
+                {"k1": None},
+            ),
+            # do not overwrite existing values
+            (
+                [{"k1": 1}, {"k1": 2}, {"k1": 3}],
+                {"k1": 1},
+            ),
+        ],
+    )
+    def test_update_nested(
```
disrupted, line 49: instead of repeating the test for update pair, please simply assert that update pair is called with the right arguments when update_nested has more than 2 or more args

disrupted (changes requested):

docs/docs/user/migration-guide/v3-v4.md
disrupted: do we need to add this new file to the mkdocs index?

kpops/components/base_components/base_defaults_component.py
raminqaf: It makes sense, but let's keep everything simple and see if the core functionality of KPOps works as expected.

tests/components/test_helm_app.py
raminqaf: I was getting an error on PyRight here. `repo_config` type is `HelmRepoConfig | None`
https://github.com/bakdata/kpops/blob/2edbb7b5aae959df3d954bb79fc9f399c41007a3/kpops/components/base_components/helm_app.py#L64

tests/utils/test_dict_ops.py
raminqaf: The reason I wrote a test for both of these methods is that both are public. The public method should be tested separately.
I can convert the method `update_nested_pair` into a private method mock it, and verify if it is called.

disrupted: imo there's no need to make it private. I agree about tests, but since `update_nested` doesn't have much functionality on its own it doesn't make sense to apply the whole set of parametrized tests that are used for testing `update_nested_pair`

docs/docs/user/migration-guide/v3-v4.md
```diff
@@ -0,0 +1,41 @@
...
+
+<!-- dprint-ignore-start -->
+
+!!! warning inline end
+    **The `--defaults` flag is removed**
+
+<!-- dprint-ignore-end -->
+
+It is possible now to use multiple default values. The `defaults.yaml` (or `defaults_<env>.yaml`) files can be distributed across multiple files. These will be picked up by KPOps and get merged into a single `pipeline.yaml` file.
+KPOps starts from reading the default files from where the pipeline path is defined and picks up the every default file on its way to where the `pipeline_base_dir` is defined.
```
disrupted, line 13:
```suggestion
KPOps starts from reading the default files from where the pipeline path is defined and picks up every defaults file on its way to where the `pipeline_base_dir` is defined.
```

disrupted: great explanation!

kpops/components/base_components/base_defaults_component.py
```diff
@@ -193,31 +190,44 @@ def defaults_from_yaml(path: Path, key: str) -> dict:
...
-    environment_default_file_path = (
-        defaults_dir
-        / Path(f"{config.defaults_filename_prefix}_{environment}").with_suffix(".yaml")
-        if environment is not None
-        else main_default_file_path
-    )
+    default_paths = []
+
+    if not pipeline_path.is_file():
+        message = f"No pipeline.yaml found in directory: {pipeline_path.parent}"
```
disrupted, line 209: I would prefer
```suggestion
        message = f"{pipeline_path} is not a valid pipeline file."
```
the reason is that KPOps also supports `pipeline.json` which we use in one external project that's based on KPOps

tests/components/test_helm_app.py
disrupted: could you ignore the error for now, please? otherwise the override doesn't make sense

disrupted (changes requested):

tests/components/resources/pipelines/test-distributed-defaults/defaults_development.yaml
```diff
@@ -0,0 +1,3 @@
+helm-app:
```
disrupted, line 1: is it on purpose that this has the exact same contents as `defaults.yaml`?

raminqaf: The content is not important for the tests. I checked the order of the files picked up. The snapshot tests are already covering the merge functionality. Yet, I have changed it, so it should now be less confusing.

disrupted (changes requested):
please update the GitHub action and docs

tests/utils/test_dict_ops.py
raminqaf: done

tests/utils/test_dict_ops.py
```diff
@@ -2,12 +2,41 @@
...
+    generate_substitution,
+    update_nested,
+    update_nested_pair,
+)
 from kpops.utils.types import JsonType
 
 
 class TestDictOps:
+    def test_update_nested(self, mocker: MockerFixture):
+        list_dic = [{"k1": {"foo": 1}}, {"k1": {"bar": ""}}, {"k1": {"baz": "2"}}]
```
disrupted, line 17: looks great!
I would just rename the input to something like `dicts`, it doesn't really matter if it's a tuple or list

tests/utils/test_dict_ops.py
```diff
@@ -2,12 +2,41 @@
...
+
+        actual = update_nested(*list_dic)
+
+        update_nested_pair_mock.assert_has_calls(
+            [
+                mocker.call({"k1": {"foo": 1}}, {"k1": {"bar": ""}}),
+                mocker.call(
+                    {"k1": {"bar": "", "baz": "2", "foo": 1}}, {"k1": {"baz": "2"}}
+                ),
+            ]  # type: ignore[reportGeneralTypeIssues]
```
disrupted, line 33: annoying that this doesn't work ootb, but you can avoid the ignore like this
```python
from unittest import mock

...assert_has_calls([mock.call(...), ...])
```

raminqaf:
> the main docs also need to be updated, it currently says that a singular defaults.yaml file is supported https://bakdata.github.io/kpops/3.2/user/core-concepts/defaults/

Updated 
image

tests/utils/test_dict_ops.py
```diff
@@ -54,7 +84,8 @@ def test_update_nested_pair(
         d2: dict[str, JsonType],
         expected: dict[str, JsonType],
     ):
-        assert update_nested_pair(d1, d2) == expected
+        pair = update_nested_pair(d1, d2)
+        assert pair == expected
```
disrupted, line 88: revert? the output is not a pair

tests/components/test_helm_app.py
```diff
@@ -14,18 +13,22 @@
...
+    PIPELINE_BASE_DIR,
+    RESOURCES_PATH,
+)
 
 
 @pytest.mark.usefixtures("mock_env")
 class TestHelmApp:
     @pytest.fixture()
     def config(self) -> KpopsConfig:
+        ENV[PIPELINE_PATH] = str(RESOURCES_PATH / "pipeline.yaml")
```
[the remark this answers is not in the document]
raminqaf: Unfortunately, it is necessary to set it. If I don't, the env variable `pipeline_path` defaults to `.` and the raises the exception here:
https://github.com/bakdata/kpops/blob/37d71312014423a88e793178433d5f23319bcc2c/kpops/components/base_components/base_defaults_component.py#L264

tests/utils/test_dict_ops.py
raminqaf: Sry left over from changing the tests 😄 reverted

tests/components/test_helm_app.py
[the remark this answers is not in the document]
disrupted: hm, we might have to reevaluate that, especially if we want to support using KPOps entirely through the Python Pipeline API, where we don't have a `pipeline.yaml` file. It doesn't have to be solved right away, but can you think of any implications if we skip the requirement for a pipeline file in the function you linked?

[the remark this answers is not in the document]
disrupted: I believe we should simply skip the enrichment if `pipeline_path` is not defined.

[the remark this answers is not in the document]
raminqaf: done

kpops/pipeline.py
```diff
@@ -20,6 +20,8 @@
 from kpops.utils.environment import ENV
 from kpops.utils.yaml import load_yaml_file
 
+PIPELINE_PATH = "pipeline_path"
```
disrupted, line 23: how about `pipeline.path`? similar to the existing `${pipeline.name}` etc.

kpops/components/base_components/base_defaults_component.py
```diff
@@ -82,7 +81,8 @@ class BaseDefaultsComponent(DescConfigModel, ABC):
     )
 
     def __init__(self, **values: Any) -> None:
-        if values.get("enrich", True):
+        pipeline_path = ENV.get("pipeline_path", None)
+        if values.get("enrich", True) and pipeline_path:
```
disrupted, line 85: imo the check should be in `extend_with_defaults`. that way, we can still run substitution

kpops/components/base_components/base_defaults_component.py
```diff
@@ -176,22 +175,24 @@ def extend_with_defaults(cls, config: KpopsConfig, **kwargs: Any) -> dict[str, A
...
-        defaults_file_paths_ = get_defaults_file_paths(
-            pipeline_path, config, ENV.get("environment")
-        )
-        defaults = cls.load_defaults(*defaults_file_paths_)
-        log.debug(
-            typer.style("Enriching component of type ", bold=False)
-            + typer.style(cls.type, bold=True, underline=True)
-        )
-        return update_nested_pair(kwargs, defaults)
+        if (pipeline_path_str := ENV.get(PIPELINE_PATH)) is not None:
```
disrupted, line 176: use an early return instead

disrupted (dismissed):
LGTM, but perhaps wait for the CI fix @sujuka99 is working on before actually merging

kpops/components/base_components/base_defaults_component.py
```diff
@@ -238,8 +238,9 @@ def defaults_from_yaml(path: Path, key: str) -> dict:
     value = content.get(key)
     if value is None:
         return {}
+    default_path = path.relative_to(Path.cwd())
     log.debug(
-        f"\tFound defaults for component type {typer.style(key, bold=True, fg=typer.colors.MAGENTA)} in file:  {path}"
+        f"Found defaults for component type {typer.style(key, bold=True, fg=typer.colors.MAGENTA)} in {default_path}"
```
disrupted, line 243: why did you decide against the indent?

raminqaf: I just seemed weird to have the indent. We have no indents anywhere that is why I have it removed

disrupted (approved):

PR #443, opened 2024-01-30
Improve Sphinx docs highlighting using RST markup

disrupted:
we can use reStructuredText markup to better highlight code blocks in pydocs


raminqaf (approved):
🆒

kpops/cli/exception.py
```diff
@@ -1,2 +1,2 @@
 class ClassNotFoundError(Exception):
-    """Similar to builtin ModuleNotFoundError; class doesn't exist inside module."""
```
sujuka99, line 2: How did you choose where to use single or double back tick?

disrupted: that's one of the things I am quite unsure about, I usually just reused what was there already. but noticed that there's both ways throughout the codebase. do you know if there's a difference?

sujuka99: [There is](https://docutils.sourceforge.io/docs/user/rst/quickref.html#inline-markup), but I am not sure if it matter at all in our case

disrupted: I checked the highlighting in nvim but for me there's no difference between single and double backtick

sujuka99 (approved):

PR #446, opened 2024-02-01
Simplify execution graph logic

disrupted:


tests/pipeline/test_generate.py
```diff
@@ -786,8 +786,8 @@ async def name_runner(component: PipelineComponent):
 
         assert called_component.mock_calls == [
             mock.call("transaction-avro-producer-1"),
-            mock.call("s3-connector-1"),
             mock.call("transaction-joiner"),
+            mock.call("s3-connector-1"),
         ]
```
sujuka99, line 791: Was this a bug?

sujuka99 (approved):

tests/pipeline/test_generate.py
disrupted: that's what I asked on the last sync 😄 Alejandro looked into it and the order doesn't matter in this case

PR #447, opened 2024-02-01
Refactor Kafka topics

disrupted:
aka "topics as first-class citizen"


sujuka99 (approved):

sujuka99:
Please wait for @raminqaf for at least 1 more approval

kpops/components/streams_bootstrap/producer/model.py
```diff
@@ -8,18 +8,7 @@
...
-    :param output_topic: Output topic, defaults to None
-    """
-
-    extra_output_topics: dict[str, str] = Field(
-        default={}, description=describe_attr("extra_output_topics", __doc__)
-    )
-    output_topic: str | None = Field(
-        default=None, description=describe_attr("output_topic", __doc__)
-    )
+    """Kafka Streams settings specific to Producer."""
```
raminqaf, line 11: Is this class needed? Can't we just on the field `KafkaStreamsConfig`?
```diff
- streams: ProducerStreamsConfig = Field(
        default=..., description=describe_attr("streams", __doc__)
    )
```
```diff
+ streams: KafkaStreamsConfig = Field(
        default=..., description=describe_attr("streams", __doc__)
    )
```

raminqaf (changes requested):

kpops/components/base_components/models/topic.py
```diff
@@ -0,0 +1,97 @@
...
+        """Ensure that `cls.role` is used correctly, assign type if needed."""
+        if self.type and self.role:
+            msg = "Define `role` only if `type` is undefined"
+            raise ValueError(msg)
+        return self
+
+
+class KafkaTopic(BaseModel):
+    name: str
+    config: TopicConfig = TopicConfig()
```
raminqaf, line 82: Do we want the config to have a default value? Shouldn't we make this a mandatory field? What happens if the user doesn't define it and the default instance is used?

kpops/components/streams_bootstrap/producer/model.py
raminqaf: Sounds good! Just wanted to be sure this was the attended way!

kpops/component_handlers/topic/handler.py
```diff
@@ -15,69 +17,64 @@
...
+    def __init__(self, proxy_wrapper: ProxyWrapper) -> None:
         self.proxy_wrapper = proxy_wrapper
 
-    async def create_topics(self, to_section: ToSection, dry_run: bool) -> None:
-        for topic_name, topic_config in to_section.topics.items():
-            topic_spec = self.__prepare_body(topic_name, topic_config)
-            if dry_run:
-                await self.__dry_run_topic_creation(
-                    topic_name, topic_spec, topic_config
+    async def create_topic(self, topic: KafkaTopic, dry_run: bool) -> None:
```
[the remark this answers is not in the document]
disrupted: I didn't create this 😄 but you're right, I'll add some

kpops/component_handlers/kafka_connect/model.py
```diff
@@ -44,19 +49,47 @@ def json_schema_extra(schema: dict[str, Any], model: type[BaseModel]) -> None:
...
             msg = f"Invalid connector class {connector_class}"
             raise ValueError(msg)
         return connector_class
 
+    @pydantic.field_validator("topics", mode="before")
+    @classmethod
+    def validate_topics(cls, topics: Any) -> list[KafkaTopic] | None | Any:
+        if isinstance(topics, str):
+            return [KafkaTopic(name=topic_name) for topic_name in topics.split(",")]
+        return topics
```
raminqaf, line 68: I am wondering if this is the correct behaviour here. Shouldn't the "validate" method actually validate the input? For instance, if the user passes a malformed input. The method should check it and raise a ValidationError.

kpops/components/base_components/models/topic.py
raminqaf: No need for it then. Maybe just add a small comment with the great explanation you just wrote! 😄

kpops/component_handlers/topic/handler.py
```diff
@@ -127,14 +97,46 @@ async def __dry_run_topic_creation(
...
+                        json_body.append(
+                            {"name": difference.key, "value": config_value}
+                        )
+                await self.proxy_wrapper.batch_alter_topic_config(topic.name, json_body)
+
+            else:
+                log.info(
+                    f"Topic Creation: config of topic {topic.name} didn't change. Skipping update."
+                )
+        except TopicNotFoundException:
```
raminqaf, line 137: Not part of this PR but maybe we should look at the proxy wrapper and improve this logic. Back then we handled these with exception but maybe this would be cleaner (also more performant) with an if-else check.

disrupted: I called it `validate_topics` based on the usage of a Pydantic "`BeforeValidator`" and the examples in their docs. Essentially it is a custom deserializer (that's what the "before" indicates). 
from their docs:
> Before validators run before Pydantic's internal parsing and validation (e.g. coercion of a str to an int). These are more flexible than After validators since they can modify the raw input, but they also have to deal with the raw input, which in theory could be any arbitrary object.

The actual validation is done during by Pydantic during the init. If you want I can name it `deserialize_topics` instead

kpops/component_handlers/kafka_connect/model.py
disrupted: I called it `validate_topics` based on the usage of a Pydantic "`BeforeValidator`" and the examples in their docs. Essentially it is a custom deserializer (that's what the "before" indicates). 
from their docs:
> Before validators run before Pydantic's internal parsing and validation (e.g. coercion of a str to an int). These are more flexible than After validators since they can modify the raw input, but they also have to deal with the raw input, which in theory could be any arbitrary object.

The actual validation is done during by Pydantic during the init. If you want I can name it `deserialize_topics` instead

kpops/component_handlers/topic/handler.py
disrupted: what do you have in mind to improve the performance here?
I agree that this (and most of our other handlers and API wrappers) needs refactoring, but I believe the exception is a good pattern in this case.

kpops/component_handlers/kafka_connect/model.py
raminqaf: Yeah I like that name more. I was using this functionality in another project and at first it seemed very confusing to me!

kpops/component_handlers/topic/handler.py
raminqaf: I got mixed up with an article I read recently for Java. In Java this is considered as an anti-pattern. But apparently it's okay in Python. Here a short explanation:

https://stackoverflow.com/a/1835844

raminqaf (dismissed):

tests/pipeline/snapshots/snap_test_example.py
```diff
@@ -25,8 +25,6 @@
                 'schedule': '0 12 * * *',
                 'streams': {
                     'brokers': 'http://k8kafka-cp-kafka-headless.kpops.svc.cluster.local:9092',
-                    'extraOutputTopics': {
```
raminqaf, line 28: Is this the correct behavior? The `generate` output does not show the `extraOutputTopics`, `from`, and `to` sections anymore.

disrupted: yep, extra output topics are hidden if empty now and cleaners shouldn't have `from` and `to` section

raminqaf (approved):

PR #450, opened 2024-02-07
Fix order of pipeline steps for clean/reset

disrupted:
#405 broke this

### DIFF
--- a/tests/pipeline/test_clean.py
+++ b/tests/pipeline/test_clean.py
@@ -58,15 +58,13 @@ class TestClean:
         streams_app_mock_clean = mocker.patch(
             "kpops.components.streams_bootstrap.streams.streams_app.StreamsApp.clean",
         )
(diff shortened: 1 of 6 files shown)


tests/pipeline/test_clean.py
```diff
@@ -0,0 +1,92 @@
...
+    #         ProducerApp(
+    #             name="producer",
+    #             namespace="test-namespace",
+    #             app=ProducerAppValues(
+    #                 streams=ProducerStreamsConfig(brokers=config.kafka_brokers)
+    #             ),
+    #             **kwargs,
+    #         )
+    #     )
+    #     return pipeline
```
sujuka99, line 52: Is this intentionally left like this?

disrupted: yep, we don't have this part of the Pipeline API (deploy, destroy, clean, ...) yet but I am planning to add it soon

sujuka99 (approved):

PR #449, opened 2024-02-06
Fix substitution

sujuka99:
closes #448

PR #451, opened 2024-02-07
Fix nested substitution

sujuka99:
Preceeds #449

### DIFF
--- a/kpops/utils/yaml.py
+++ b/kpops/utils/yaml.py
@@ -1,3 +1,4 @@
+import json
 from collections.abc import Mapping
 from pathlib import Path
 from typing import Any
@@ -48,7 +49,6 @@ def substitute_nested(input: str, **kwargs) -> str:
     """Allow for multiple substitutions to be passed.
 
     Will make as many passes as needed to substitute all possible placeholders.
-    A ceiling is set to avoid infinite loops.
 
     HINT: If :param input: is a ``Mapping`` that you converted into ``str``,
     You can pass it as a string, and as a ``Mapping`` to enable self-reference.
@@ -63,7 +63,7 @@ def substitute_nested(input: str, **kwargs) -> str:
         }
     >>> input = "${a}, ${b}, ${c}, ${d}"
     >>> print("Substituted string: " + substitute_nested(input, **substitution))
-    0, 0, 0, 0
+    "0, 0, 0, 0"
 
     :param input: The raw input containing $-placeholders
     :param **kwargs: Substitutions
@@ -72,6 +72,7 @@ def substitute_nested(input: str, **kwargs) -> str:
     """
     if not kwargs:
         return input
+    kwargs = substitute_in_self(kwargs)
     old_str, new_str = "", substitute(input, kwargs)
     steps = set()
     while new_str not in steps:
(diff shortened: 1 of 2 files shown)


tests/utils/test_yaml.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+@pytest.mark.parametrize(
+    ("input", "substitution", "expected"),
+    [
+        pytest.param(
+            '"a": "b", "${a}": "c", "a_${b}": "d", "e": "${a_${b}}"',
+            {"a": "b", "${a}": "c", "a_${b}": "d", "e": "${a_${b}}"},
+            '"a": "b", "b": "c", "a_c": "d", "e": "d"',
+            id="requires-multiple-passes",
```
disrupted, line 13: nice 🤩  I didn't know about these. does it print the id if a certain test case fails?

sujuka99: It does :)

sujuka99:
I tried hard, but couldn't come up with a way to combine the 2 functions that made sense.

kpops/utils/yaml.py
```diff
@@ -44,14 +45,22 @@ def prepare_substitution(substitution: Mapping[str, Any]) -> dict[str, Any]:
     return ImprovedTemplate(input).safe_substitute(**prepare_substitution(substitution))
 
 
+def _diff_substituted_str(s1: str, s2: str):
+    """Compare 2 strings, raise exception if equal.
```
disrupted, line 49: description doesn't match the code

kpops/utils/yaml.py
```diff
@@ -63,26 +72,43 @@ def substitute_nested(input: str, **kwargs) -> str:
...
+def substitute_in_self(input: dict[str, Any]) -> dict[str, Any]:
+    """Substitute all self-references in mapping.
+
+    Will make as many passes as needed to substitute all possible placeholders.
+
+    :param input: Mapping containing $-placeholders
+    :raises ValueError: An infinite loop condition detected. Check substitution variables.
+    :return: Substituted input mapping as dict
+    """
+    old_str, new_str = "", substitute(json.dumps(input), input)
```
disrupted, line 103: could you give a little insight what's the general idea here? this looks very similar to the existing substitution function, only that we're converting to JSON first. Wouldn't it be easier to just do `model_dump_json` in `BaseDefaultsComponent`?

tests/components/test_base_defaults_component.py
```diff
@@ -189,11 +189,11 @@ def test_multiple_generations(
...
         self, config: KpopsConfig, handlers: ComponentHandlers
     ):
-        ENV["pipeline_name"] = str(DEFAULTS_PATH)
+        ENV["pipeline_name"] = DEFAULTS_PATH.as_posix()
         component = EnvVarTest(config=config, handlers=handlers)
 
-        assert component.name == str(
-            DEFAULTS_PATH
+        assert (
+            Path(component.name or "") == DEFAULTS_PATH
```
disrupted, line 196: we should remove the `name` override in `EnvVarTest`, it should never be `None`

disrupted (changes requested):

kpops/utils/yaml.py
sujuka99: It resolves the self-references in a dict. Before this PR, "chained" placeholders in the substitution mapping would get only partially substituted in the input string.

I chose to do that in a separate function, because I would have to add complexity to `substitute_nested` to differentiate between the cases of 'self-substitution' and 'normal substitution'. It seemed needlessly tangled to me. If I wanted to avoid the checks, I would have to add the code from `substitute_in_self` to `substitute_nested`. That would only lead again to worse readability and would make the function more complex without any benefits.

`substitute_in_self` is mostly useful for preparing the `kwargs` in `substitute_nested`. When substituting in a component, they are a flat dict that looks like this:

```
{
    ...
    'component.name': 'name',
    'component.type': 'type',
    'component.app.section.leaf': 'value',
    ...
}
```

tests/components/test_base_defaults_component.py
sujuka99: `name` is defined in `PipelineComponent`, so it is not inherited from `BaseDefaultsComponent` by `EnvVarTest`. I could have a Pyright ignore comment instead of the `or`?

disrupted: Ah, true. Then please put a prior `assert component.name`

disrupted (approved):
please make a patch release

PR #456, opened 2024-02-15
Trim and hash Helm name override to 63 characters

### LINKED ISSUE #455: Clean up fails for names > 63 chars
In #390, hash-based trimming for Helm releases was already implemented. I now run into similar problems but with k8s resources. Having names bigger than 63 chars, KPOPs creates resources (in my case clean up jobs) with the same name and fails eventually because of it.

Also, it might fail because of auto-generated labels that are too long:

```
RuntimeError: Error: Job.batch "my-long-name" is invalid: [metadata.labels: Invalid value: "my-long-name": must be no more than 63 characters, spec.template.labels: Invalid value: "my-long-name: must be no more than 63 characters]
```

disrupted:
Fixes #455

### DIFF
--- a/kpops/component_handlers/kubernetes/utils.py
+++ b/kpops/component_handlers/kubernetes/utils.py
@@ -1,0 +1,28 @@
+import hashlib
+import logging
+
+log = logging.getLogger("K8sUtils")
+
+
+def trim(max_len: int, name: str, suffix: str) -> str:
+    """Shortens long K8s identifiers.
+
+    Kubernetes has a character limit for certain identifiers.
+    If the name exceeds the character limit:
+    1. trim the string and fetch the first ``max_len - len(suffix)`` characters.
+    2. replace the last 6 characters with the SHA-1 encoded string (with "-") to avoid collision
+    3. append the suffix if given
+
+    :param name: The name to be shortened.
+    :param suffix: The release suffix to preserve
+    :return: Trimmed + hashed version of the name if it exceeds the character length otherwise the actual name
+    """
+    if len(name) > max_len:
+        exact_name = name[: max_len - len(suffix)]
+        hash_name = hashlib.sha1(name.encode()).hexdigest()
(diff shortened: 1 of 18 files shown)


kpops/component_handlers/kubernetes/utils.py
```diff
@@ -0,0 +1,28 @@
...
+
+    :param name: The name to be shortened.
+    :param suffix: The release suffix to preserve
+    :return: Trimmed + hashed version of the name if it exceeds the character length otherwise the actual name
+    """
+    if len(name) > max_len:
+        exact_name = name[: max_len - len(suffix)]
+        hash_name = hashlib.sha1(name.encode()).hexdigest()
+        new_name = exact_name[:-6] + "-" + hash_name[:5] + suffix
+        log.critical(
```
disrupted, line 24: should we change the log level? 😄 this is so common that I wouldn't call it _critical_

raminqaf: Yeah I think info would be good

kpops/component_handlers/kubernetes/utils.py
```diff
@@ -0,0 +1,28 @@
+import hashlib
+import logging
+
+log = logging.getLogger("K8sUtils")
+
+
+def trim(max_len: int, name: str, suffix: str) -> str:
```
raminqaf, line 7: If this is under `kubernetes` folder would it make sense to move it under `KubernetesApp` class?

disrupted: I don't have a preference, I just kept it the same way as the previous Helm trimming function which was part of Helm wrapper utils.

raminqaf: Okay, we can perspectively refactor it there. I think it is cleaner.

raminqaf (approved):

PR #465, opened 2024-02-28
Document precedence between env vars and config.yaml

JakobEdding:


docs/docs/resources/variables/config_env_vars.md
```diff
@@ -1,4 +1,4 @@
-These variables are a lower priority alternative to the settings in `config.yaml`. Variables marked as required can instead be set in the global config.
+These variables take precedence over the settings in `config.yaml`. Variables marked as required can instead be set in the global config.
```
disrupted, line 1: How about "these variables _always_ take precedence"? (over CLI flags, config.yaml, dotenv)

JakobEdding: It seems to me that this section specifically talks about setting values that can otherwise be specified in config.yaml. I'll add a similar wording to the CLI section below. Does that work for you?

raminqaf (commented):
Since you are at it, please add a pyDoc to the function in the [KPOps config class](https://github.com/bakdata/kpops/blob/main/kpops/config.py#L134)

PR #470, opened 2024-03-04
Validate autoscaling mandatory fields when enabled

raminqaf:


kpops/components/streams_bootstrap/streams/model.py
```diff
@@ -183,6 +188,16 @@ class StreamsAppAutoScaling(CamelCaseConfigModel, DescConfigModel):
     )
     model_config = ConfigDict(extra="allow")
 
+    @model_validator(mode="after")
+    def validate_model_config(self) -> Self:
+        if self.enabled and (not self.consumer_group or not self.lag_threshold):
```
disrupted, line 193: this will fail if `lag_threshold=0` (perhaps include in your test cases?)

disrupted (changes requested):
great!

kpops/components/streams_bootstrap/streams/model.py
```diff
@@ -183,6 +188,16 @@ class StreamsAppAutoScaling(CamelCaseConfigModel, DescConfigModel):
...
+            msg = (
+                "If app.autoscaling.enabled is set to true, "
+                "the app.autoscaling.fields consumer_group and app.autoscaling.lag_threshold should be set."
```
disrupted, line 196:
```suggestion
                "the firlds app.autoscaling.consumer_group and app.autoscaling.lag_threshold should be set."
```

kpops/components/streams_bootstrap/streams/model.py
```diff
@@ -183,6 +188,16 @@ class StreamsAppAutoScaling(CamelCaseConfigModel, DescConfigModel):
     )
     model_config = ConfigDict(extra="allow")
 
+    @model_validator(mode="after")
+    def validate_model_config(self) -> Self:
```
disrupted, line 192: this will cause an error with Pyright on Python 3.10, for now I believe it's best to 
```suggestion
    def validate_model_config(self) -> StreamsAppAutoScaling:
```

and

```python
from __future__ import annotations
```

perhaps add a comment similar to here
https://github.com/bakdata/kpops/blob/108b25cb569a434107ceecc27591575430cef41a/kpops/component_handlers/kubernetes/model.py#L20

kpops/components/streams_bootstrap/streams/model.py
raminqaf: True, I can change this to:
```suggestion
        if self.enabled and (self.consumer_group is None or not self.lag_threshold is None):
```

kpops/components/streams_bootstrap/streams/model.py
raminqaf: -.-
```
 error: Argument of type "(self: Self@StreamsAppAutoScaling) -> StreamsAppAutoScaling" cannot be assigned to parameter of type "_AnyModelAfterValidator[_ModelType@model_validator]"
```

disrupted: hm, in that case we have to remove the return type annotation unfortunately

raminqaf: @disrupted Fixed!

kpops/components/streams_bootstrap/streams/model.py
disrupted: yup, also remove the `not`, otherwise looks good

disrupted (approved):

PR #481, opened 2024-04-03
Add pyright matcher

sujuka99:


.github/workflows/ci.yaml
```diff
@@ -52,7 +52,15 @@ jobs:
         run: poetry run pre-commit run ruff-format --all-files --show-diff-on-failure
 
       - name: Typing (pyright)
-        run: poetry run pre-commit run pyright --all-files
+        shell: bash
```
disrupted, line 55: we could set it as default shell https://docs.github.com/en/actions/using-jobs/setting-default-values-for-jobs#example-set-the-default-shell-and-working-directory

disrupted (approved):

PR #485, opened 2024-04-29
Allow custom timeout for external services

### LINKED ISSUE #457: httpx timeout when running with `--parallel` flag
My clean up job fails after ~1m30s even with a timeout set to 10min. The connection to Kafka Connection seems to timeout because it is busy with all the concurrent requests.

disrupted:
Closes #457


tests/cli/test_kpops_config.py
```diff
@@ -25,7 +25,6 @@ def test_kpops_config_with_default_values():
     assert default_config.schema_registry.url == AnyUrl("http://localhost:8081")
     assert default_config.kafka_rest.url == AnyUrl("http://localhost:8082")
     assert default_config.kafka_connect.url == AnyUrl("http://localhost:8083")
-    assert default_config.timeout == 300
```
raminqaf, line 28: should we test the default value for each operation? 
```python
assert default_config.kafka_rest.timeout == 30
assert default_config.kafka_connect.timeout == 30
assert default_config.schema_registry.timeout == 30
```

docs/docs/user/migration-guide/v4-v5.md
```diff
@@ -0,0 +1,22 @@
...
+  kafka_rest:
+    url: "http://my-custom-rest.url:8082"
++   timeout: 300
+  kafka_connect:
+    url: "http://my-custom-connect.url:8083"
++   timeout: 300
+  schema_registry:
+    enabled: true
+    url: "http://my-custom-sr.url:8081"
++   timeout: 300
```
raminqaf, line 21: I would mention that the default for each operation is 30 seconds, and I would also apply it to the diff. We moved away from the 300 😄

raminqaf (commented):
Nit picks but LGTM

docs/docs/user/migration-guide/v4-v5.md
```diff
@@ -0,0 +1,22 @@
+# Migrate from V4 to V5
+
+## [Allow custom timeout for external services](https://github.com/bakdata/kpops/pull/485)
+
+The global `timeout` setting has been removed. Instead, an individual timeout can be set for each external service. The default is 30 seconds.
+
+#### config.yaml
+
+```diff
+- timeout: 300
```
JakobEdding, line 10: Removing this general timeout is the breaking change of this PR, right? What do you think about removing it in a separate PR? Then we would be able to ship the specific timeouts already in a minor KPOps release and wouldn't have to wait for this feature until v5.

disrupted: Yes, exactly. What do you mean by "wait for this feature until v5"? We will release the new major and you can update to it right away.

Afaict the global timeout was only ever used for Kafka Connect tbf 🤷🏻‍♂️

JakobEdding: Ah ok! I thought maybe you were planning to defer the major release to a point in the future.

raminqaf (approved):

PR #488, opened 2024-05-14
Fix missing await on Kubernetes API

raminqaf:
The method needs to be awaited. See example:
https://github.com/tomplus/kubernetes_asyncio#example

### DIFF
--- a/kpops/component_handlers/kubernetes/pvc_handler.py
+++ b/kpops/component_handlers/kubernetes/pvc_handler.py
@@ -23,7 +23,7 @@ class PVCHandler:
         async with ApiClient() as api:
             core_v1_api = client.CoreV1Api(api)
             pvc_list = await core_v1_api.list_namespaced_persistent_volume_claim(
-                self.namespace, label_selector=f"app={self.app_name}"
+                namespace=self.namespace, label_selector=f"app={self.app_name}"
             )
 
             pvc_names = [pvc.metadata.name for pvc in pvc_list.items]
--- a/kpops/component_handlers/kubernetes/pvc_handler.py
+++ b/kpops/component_handlers/kubernetes/pvc_handler.py
@@ -44,6 +44,6 @@ class PVCHandler:
                 f"Deleting in namespace '{self.namespace}' StatefulSet '{self.app_name}' PVCs '{pvc_names}'"
             )
             for pvc_name in pvc_names:
-                core_v1_api.delete_namespaced_persistent_volume_claim(
+                await core_v1_api.delete_namespaced_persistent_volume_claim(
                     pvc_name, self.namespace
-                )
+                )  # type: ignore [reportGeneralTypeIssues]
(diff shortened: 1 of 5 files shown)


disrupted (approved):
weird that this wasn't caught by CI

kpops/components/base_components/cleaner.py
```diff
@@ -20,6 +20,10 @@ class Cleaner(HelmApp, ABC):
     def full_name(self) -> str:
         return super().full_name + self.suffix
 
+    @property
+    def app_full_name(self) -> str:
+        return super().full_name
+
```
raminqaf, line 26: A hack for now to have the full_name of the app without the `-clean` suffix

disrupted: could you elaborate?

raminqaf: `super().full_name` will evoke from the `Cleaner` class. So, we will still have the suffix 
```python
    @property
    @override
    def full_name(self) -> str:
        return super().full_name + self.suffix
```

disrupted: that sounds like the behavior of `self.full_name`. I believe `super().full_name` does what you want (same as your new property without defining a function for it)

disrupted: how about `super(HelmApp).full_name`?

raminqaf:
```
AttributeError: 'super' object has no attribute 'full_name'
```

disrupted: my bad, needs to be `super(HelmApp, self).full_name`

raminqaf: Cool! Done

kpops/component_handlers/kubernetes/pvc_handler.py
```diff
@@ -44,6 +44,6 @@ async def delete_pvcs(self) -> None:
                 f"Deleting in namespace '{self.namespace}' StatefulSet '{self.app_name}' PVCs '{pvc_names}'"
             )
             for pvc_name in pvc_names:
-                core_v1_api.delete_namespaced_persistent_volume_claim(
+                await core_v1_api.delete_namespaced_persistent_volume_claim(
```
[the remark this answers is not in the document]
raminqaf: It is very confusing. It didn't work without the await. I will have another test before merging it. I would make the change to Lighkube in another PR.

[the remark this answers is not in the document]
raminqaf: This is the example code in his repository:
```python
    # use the context manager to close http sessions automatically
    async with ApiClient() as api:

        v1 = client.CoreV1Api(api)
        print("Listing pods with their IPs:")
        ret = await v1.list_pod_for_all_namespaces()
```
I checked `list_pod_for_all_namespaces()` and does not have `async def`...

disrupted (approved):

PR #491, opened 2024-05-22
Add YAML separator (---) to stdout

raminqaf:
The output is not a valid Kubernetes manifest whenever calling the' manifest' command. 
```diff
+ ---
apiVersion: batch/v1
kind: Job
metadata:
  labels:
    app: word-count-data-producer
#...
+ ---
apiVersion: apps/v1
kind: Deployment
#...
```

### DIFF
--- a/kpops/utils/yaml.py
+++ b/kpops/utils/yaml.py
@@ -121,6 +121,7 @@ def print_yaml(data: Mapping | str, *, substitution: dict | None = None) -> None
     """
     if not isinstance(data, str):
         data = yaml.safe_dump(dict(data))
+        data = f"---\n{data}"
     syntax = Syntax(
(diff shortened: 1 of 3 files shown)


kpops/utils/yaml.py
```diff
@@ -121,6 +121,7 @@ def print_yaml(data: Mapping | str, *, substitution: dict | None = None) -> None
     """
     if not isinstance(data, str):
         data = yaml.safe_dump(dict(data))
+        data = f"---\n{data}"
```
disrupted, line 124: technically there's also `yaml.safe_dump_all` for a sequence of YAML documentes but I am fine with merging it like this

disrupted (approved):

PR #495, opened 2024-05-28
Remove default filename prefix from KPOps config

raminqaf:


disrupted (changes requested):
please also remove it from the docs

PR #498, opened 2024-06-25
clean/reset streams-bootstrap components with cluster values

raminqaf:


kpops/api/__init__.py
```diff
@@ -217,7 +217,6 @@ def clean(
     )
 
     async def clean_runner(component: PipelineComponent):
-        await component.destroy(dry_run)
```
raminqaf, line 220: TODO: This is problematic. We should fetch the Helm values first and then destroy the app.

kpops/components/base_components/helm_app.py
```diff
@@ -141,6 +141,9 @@ def template_flags(self) -> HelmTemplateFlags:
             api_version=self.config.helm_config.api_version,
         )
 
+    def helm_values(self) -> dict[str, Any] | None:
```
disrupted, line 144: imo the wrapper method is unnecessary, why not just call `helm.get_values` directly?

raminqaf: Yeah, I know I will have it removed 😄

kpops/component_handlers/helm_wrapper/helm.py
```diff
@@ -129,6 +129,29 @@ async def uninstall(
...
+            "get",
+            "values",
+            release_name,
+            "--namespace",
+            namespace,
+            "--output",
+            "yaml",
+        ]
+        try:
+            return yaml.safe_load(self.__execute(command))
```
raminqaf, line 149: TODO: make async

kpops/components/base_components/kafka_app.py
```diff
@@ -110,13 +110,28 @@ async def clean(self, dry_run: bool) -> None:
...
+        """Fetch the image tag of the streams-bootstrap app using the 'helm get values' command.
+
+        If the release doesn't exist, it will fall back to the specified imageTag in the default.yaml/pipeline.yaml
+
+        :return: Image tag of the streams-bootstrap app
+        """
+        cluster_values = await self.helm.get_values(
+            self.namespace, self.helm_release_name
+        )
+        return cluster_values["imageTag"] if cluster_values else self.app.image_tag
```
disrupted, line 133: how about deserializing it to `StreamsBootstrapValues`?

```suggestion
if cluster_values:
    return StreamsBootstrapValues.model_validate(cluster_values).image_tag
```

I think I would also prefer returning an optional here and let the caller handle what happens in the `None` case

raminqaf: Addressed

disrupted: @raminqaf did you forget to push? 😄

kpops/components/streams_bootstrap/producer/producer_app.py
```diff
@@ -28,6 +28,19 @@ def helm_chart(self) -> str:
...
+        self.app.streams.delete_output = True
+        await super().clean(dry_run)
+
+    async def update_cleaner_with_cluster_values(self):
+        """Update cleaner with cluster values if the release exists."""
+        cluster_values = await self.fetch_values_from_cluster()
+        if cluster_values:
+            self.app = ProducerAppValues.model_validate(cluster_values)
+            self.app.name_override = self.helm_name_override
+
```
raminqaf, line 43: @disrupted I am using this method on the `producer_app.py` & `streams_app.py`. The difference is only `ProducerAppValues` & `StreamsAppValues`. Is there a nice Pydantic way to abstract this into the KafkaAppCleaner?

kpops/components/base_components/kafka_app.py
raminqaf: I have changed the logic a bit. So this comment becomes obsolete :)

kpops/components/streams_bootstrap/producer/producer_app.py
disrupted: you could try `self.app.__class__.model_validate`

kpops/components/base_components/kafka_app.py
```diff
@@ -117,6 +117,16 @@ async def clean(self, dry_run: bool) -> None:
...
             await self.destroy(dry_run)
 
+    async def update_cleaner_with_cluster_values(self) -> None:
+        """Update cleaner with cluster values if the release exists."""
+        cluster_values = await self.helm.get_values(
+            self.namespace, self.helm_release_name
+        )
+        if cluster_values:
+            self.app = self.app.__class__.model_validate(cluster_values)
+            self.app.name_override = self.helm_name_override
```
disrupted, line 127: perhaps we could consider setting it to `None` here? Then the behavior is aligned with our other components, where it gets assigned during `to_helm_values` https://github.com/bakdata/kpops/blob/e692e8245524300a163bd769fe0b10fb58918f49/kpops/components/base_components/helm_app.py#L189

kpops/components/streams_bootstrap/streams/streams_app.py
```diff
@@ -123,10 +134,8 @@ def helm_chart(self) -> str:
 
     @override
     async def reset(self, dry_run: bool) -> None:
-        self._cleaner.app.streams.delete_output = False
-        await self._cleaner.clean(dry_run)
+        await self._cleaner.reset(dry_run)
```
disrupted, line 137: nice!

tests/components/test_producer_app.py
```diff
@@ -352,3 +361,72 @@ def test_get_output_topics(
             KafkaTopic(name="producer-app-output-topic"),
             KafkaTopic(name="extra-topic-1"),
         ]
+
+    @pytest.mark.asyncio()
+    async def test_should_deploy_clean_up_job_with_values_in_cluster_when_clean(
```
disrupted, line 366: should we add a test for reset as well?

kpops/components/streams_bootstrap/producer/producer_app.py
```diff
@@ -28,6 +31,10 @@ def helm_chart(self) -> str:
             f"{self.repo_config.repository_name}/{AppType.CLEANUP_PRODUCER_APP.value}"
         )
 
+    @override
+    async def clean(self, dry_run: bool) -> None:
+        await super().clean(dry_run)
```
disrupted, line 36: if this only calls super we can remove it

tests/components/test_producer_app.py
raminqaf: The producer app doesn't have a reset, but I added a test to check it 😄

kpops/components/base_components/kafka_app.py
disrupted: discussed it further and agreed against it

kpops/components/streams_bootstrap/streams/streams_app.py
```diff
@@ -124,6 +125,19 @@ def add_extra_output_topic(self, topic: KafkaTopic, role: str) -> None:
...
+        cluster_values = await self.helm.get_values(
+            self.namespace, self.helm_release_name
+        )
+        if cluster_values:
+            log.debug("Fetched Helm chart values from cluster")
+            name_override = self._cleaner.helm_name_override
+            self._cleaner.app = self.app.model_validate(cluster_values)
+            self._cleaner.app.name_override = name_override
+
+        await super().destroy(dry_run)
```
disrupted, line 139: don't we first have to destroy before running the cleaner?

raminqaf: This is destroying the component 🤔 It calls `super().destroy()`, which destroys the component.

disrupted: yes, I am asking about the sequence. Here we're deploying the cleaner first, then destroying the parent app

disrupted: nvm, my bad

Issue #379, opened 2023-10-25
Expand Kafka Connect API support for resetting offsets


disrupted:
Kafka 3.6 has added support for resetting the connector offset using the Kafka Connect REST API
when can use the new official API instead of the custom streams-bootstrap connector resetter
https://cwiki.apache.org/confluence/display/KAFKA/KIP-875%3A+First-class+offsets+support+in+Kafka+Connect

PR #499, opened 2024-06-26
Add image tag field to streams-bootstrap app values

raminqaf:


kpops/components/base_components/kafka_app.py
```diff
@@ -131,6 +131,14 @@ class KafkaApp(PipelineComponent, ABC):
         description=describe_attr("app", __doc__),
     )
 
+    @pydantic.model_validator(mode="after")
+    def warning_for_latest_image_tag(self) -> Self:
+        if self.app.image_tag == "latest" and "$" not in self.name:
```
raminqaf, line 136: The validation was called before `self.name` got substituted, which is why this condition exists.

kpops/components/base_components/kubernetes_app.py
```diff
@@ -19,7 +19,14 @@
 
 
 class KubernetesAppValues(CamelCaseConfigModel, DescConfigModel):
-    """Settings specific to Kubernetes apps."""
+    """Settings specific to Kubernetes apps.
+
+    :param image_tag: The image tag of the Kubernetes app.
+    """
+
+    image_tag: str = Field(
```
disrupted, line 27: is this standard described somewhere by Kubernetes docs? I was under the impression that it's more of a best practice thing originating in our Helm charts

raminqaf: I pushed a new definition

disrupted: it doesn't answer my question though. is this even standardized by Kubernetes?

raminqaf: Regarding the helm charts, we can use the command `show values` to fetch the values from the `values.yaml` file as the default.
https://helm.sh/docs/helm/helm_show_values/#helm-show-values

disrupted: I think we're talking at cross purposes 😄 what I am curious about is if the `imageTag` attribute is standardized by Kubernetes?

raminqaf: it is actually set as the `image:tag` in the Deployment, Pod, StatefulSet, etc.
https://kubernetes.io/docs/concepts/containers/images/#:~:text=If%20you%20don%27t%20specify%20a%20tag%2C%20Kubernetes%20assumes%20you%20mean%20the%20tag%20latest.

raminqaf: I would rather have it in `KafkaAppValues`.

disrupted: but it's not related to `KafkaApp` 🤷🏻‍♂️

kpops/components/streams_bootstrap/producer/model.py
```diff
@@ -4,14 +4,15 @@
...
+from kpops.components.streams_bootstrap import StreamsBootstrapValues
 from kpops.utils.docstring import describe_attr
 
 
 class ProducerStreamsConfig(KafkaStreamsConfig):
     """Kafka Streams settings specific to Producer."""
 
 
-class ProducerAppValues(KafkaAppValues):
+class ProducerAppValues(StreamsBootstrapValues, KafkaAppValues):
```
disrupted, line 15: why does the generic producer have to inherit from streams bootstrap? 🤔

kpops/components/streams_bootstrap/__init__.py
```diff
@@ -29,3 +55,14 @@ class StreamsBootstrap(HelmApp, ABC):
         default=STREAMS_BOOTSTRAP_VERSION,
         description=describe_attr("version", __doc__),
     )
+
+    @pydantic.model_validator(mode="after")
+    def warning_for_latest_image_tag(self) -> Self:
+        if self.app.image_tag == "latest" and self.is_substituted():
```
disrupted, line 61: how about moving it as `field_validator` to `StreamsBootstrapValues`?

kpops/components/streams_bootstrap/__init__.py
```diff
@@ -29,3 +55,14 @@ class StreamsBootstrap(HelmApp, ABC):
...
+            log.warning(
```
disrupted, line 62: the warning is also raised when the image tag is explicitly set to 'latest' by the user, no?

raminqaf: Yeah, that is true... 🤔 Do you know how you can differentiate between the default and user-set values in Pydantic?

kpops/components/streams_bootstrap/producer/model.py
raminqaf: I couldn't find another way to pass the `imageTag` to the `ProducerAppValues`.
The class `ProducerApp` inherits `KafkaApp` and `StreamsBootstrap`. I think the same logic makes sense for the `ProducerAppValues`.

kpops/components/streams_bootstrap/__init__.py
raminqaf: If I move it under `StreamsBootstrapValues `, the component name cannot be fetched, and the log messages become very unclear.

kpops/components/streams_bootstrap/producer/model.py
disrupted: Ah you're right, the producer is actually not generic but specific to streams bootstrap

kpops/components/streams_bootstrap/__init__.py
disrupted: no, none that I am aware of. I do think the warning could make sense either way though. Would just need to he worded slightly differently

kpops/components/streams_bootstrap/__init__.py
disrupted: ok, that's sensible.
I would want to look into why it's not substituted though. Do you have any clues on that?

raminqaf: I haven't quite found the why, but looking at the call stack, it shows a very twisted back-and-forth between the two classes `PipelineComponent` and `BaseDefaultsComponent`.

When we are applying a component here:
https://github.com/bakdata/kpops/blob/main/kpops/pipeline.py#L308
create an object of `PipelineComponent` that calls the init of its base class `BaseDefaultsComponent`. 

Then, in the init of the base class, we call
https://github.com/bakdata/kpops/blob/3d79e8edf151b1390b00c3162b40aa526a95a198/kpops/components/base_components/base_defaults_component.py#L88

Then, the init of the `PipelineCompnent` class is called again. This calls the super class's init method again when the `enrich` field is set to False. 

I assume this will initiate the child classes (along with the validators) before anything is enriched.

kpops/components/streams_bootstrap/__init__.py
```diff
@@ -1,26 +1,60 @@
...
 )
 STREAMS_BOOTSTRAP_VERSION = "2.9.0"
 
+log = logging.getLogger("StreamsBootstrap")
+
+
+class StreamsBootstrapValues(HelmAppValues):
+    """Base value class for all streams bootstrap related components.
+
+    :param image_tag: Docker image tag of the Kafka Streams app.
```
disrupted, line 31:
```suggestion
    :param image_tag: Docker image tag of the streams-bootstrap app.
```

disrupted (changes requested):

kpops/component_handlers/kubernetes/utils.py
```diff
@@ -26,3 +29,27 @@ def trim(max_len: int, name: str, suffix: str) -> str:
         )
         return new_name
     return name
+
+
+def validate_image_tag(image_tag: str) -> str:
```
disrupted, line 34: imo the returned str is a bit useless here since it is never changed from the original input. I think raising the exception if it's invalid is enough and aligns better with the name of the function

kpops/component_handlers/kubernetes/utils.py
```diff
@@ -26,3 +29,27 @@ def trim(max_len: int, name: str, suffix: str) -> str:
...
+    msg = (
+        "Image tag is not valid. "
+        "Image tags consist of lowercase and uppercase letters, digits, underscores (_), periods (.), and dashes (-). "
+        "It can be up to 128 characters long."
+    )
+    raise ValidationError(msg)
+
+
+def is_valid_image_tag(image_tag: str) -> bool:
+    pattern = r"^[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}$"
```
disrupted, line 54: better perf if you move this to a const outside the function

kpops/api/exception.py
```diff
@@ -11,3 +11,8 @@ class ParsingException(Exception):
 
 class ClassNotFoundError(Exception):
     """Similar to builtin `ModuleNotFoundError`; class doesn't exist inside module."""
+
+
+class InvalidImageTagError(ValidationError):
+    def __init__(self, *args):
```
disrupted, line 17: needed?

kpops/component_handlers/kafka_connect/model.py
```diff
@@ -123,3 +125,10 @@ class KafkaConnectorResetterConfig(CamelCaseConfigModel):
 class KafkaConnectorResetterValues(HelmAppValues):
     connector_type: Literal["source", "sink"]
     config: KafkaConnectorResetterConfig
+    image_tag: str = Field(default="latest")
```
disrupted, line 128: inherit from `StreamsBootstrapValues` instead. Then you can drop the field and validator here

disrupted (changes requested):

kpops/component_handlers/kafka_connect/model.py
raminqaf: I didn't do this intentionally, because I don't think that `KafkaConnectorResetterValues` is or belongs to `StreamsBootstrapValues` and should be its own value.

disrupted: ah, I wasn't aware. It's just a little cumbersome to have the code duplication for it. do you think we could extract it to a shared base easily?

disrupted: but perhaps also we won't need the resetter much longer with the recent API changes we already talked about #379

raminqaf: Yeah I think so too. We can have it removed and keep it for streams app only

kpops/component_handlers/kafka_connect/model.py
```diff
@@ -123,3 +125,10 @@ class KafkaConnectorResetterConfig(CamelCaseConfigModel):
...
+
+    @pydantic.field_validator("image_tag", mode="before")
```
disrupted, line 130: wouldn't after make more sense? at least then we already know it's a str after Pydantic is done with its checks. Alternatively, I believe we could even set the regex pattern directly on the Pydantic field.

tests/component_handlers/kubernetes/test_utils.py
```diff
@@ -0,0 +1,37 @@
...
+
+from kpops.api.exception import ValidationError
+from kpops.component_handlers.kubernetes.utils import (
+    validate_image_tag,
+)
+
+MAX_LENGTH = 128
+
+
+def test_validate_image_tag_valid():
```
disrupted, line 11: please rewrite as parametrized test

kpops/component_handlers/kafka_connect/model.py
disrupted: e.g.

```python
ImageTag = Annotated[str, StringConstraints(pattern=IMAGE_TAG_PATTERN)]
```

raminqaf: Wouldn't that be enough?
```
    image_tag: str = Field(
        default="latest",
        pattern=IMAGE_TAG_PATTERN,
        description=describe_attr("image_tag", __doc__),
    )
```

disrupted: yes, that also works, but I believe `Annotated` is the newer/recommended way. I don't mind either way

disrupted: perhaps Annotated is transported to the generated schema, it might be worth to check

raminqaf: Changing it to Annotated, I get an error on `gen-schema` pre-commit hook 🤔 
```
pydantic.errors.PydanticUserError: `PipelineSchema` is not fully defined; you should define `pydantic`, then call `PipelineSchema.model_rebuild()`.

For further information visit https://errors.pydantic.dev/2.6/u/class-not-fully-defined

```

disrupted: idk, for me both work and produce the exact same schema

kpops/components/streams_bootstrap/__init__.py
```diff
@@ -1,26 +1,61 @@
...
+        default="latest",
+        pattern=IMAGE_TAG_PATTERN,
+        description=describe_attr("image_tag", __doc__),
+    )
+
 
 class StreamsBootstrap(HelmApp, ABC):
     """Base for components with a streams-bootstrap Helm chart.
 
+    :param app: Streams bootstrap app values
```
disrupted, line 48: nit but let's try to keep the naming consistent
```suggestion
    :param app: streams-bootstrap app values
```

disrupted (commented):
nit, otherwise LGTM

kpops/component_handlers/kafka_connect/model.py
raminqaf: After struggling with the Annotated way I went for the classical `pattern=` input in the `Field` method. Updated the tests accordingly

disrupted (approved):

PR #500, opened 2024-06-26
Automatic loading of namespaced custom components

disrupted:


docs/docs/user/migration-guide/v6-v7.md
raminqaf, line 1: Do you think it is worth mentioning the wildcard import here, or would you rather keep it in the main documentation as a tip?

kpops/api/registry.py
```diff
@@ -45,32 +52,52 @@ def __getitem__(self, component_type: str) -> type[PipelineComponent]:
...
-        return next(_find_classes(module_name, baseclass))
+        return next(_find_classes(modules, base=base))
     except StopIteration as e:
         raise ClassNotFoundError from e
 
 
-def _find_classes(module_name: str, baseclass: type[T]) -> Iterator[type[T]]:
+def import_module(module_name: str) -> ModuleType:
     module = importlib.import_module(module_name)
+    # TODO: remove? unnecessary now
```
raminqaf, line 72: What about this TODO?

kpops/api/registry.py
```diff
@@ -45,32 +52,52 @@ def __getitem__(self, component_type: str) -> type[PipelineComponent]:
...
             log.debug(f"Picked up: {file_path}")
-    for _, _class in inspect.getmembers(module, inspect.isclass):
-        if not __filter_internal_kpops_classes(
-            _class.__module__, module_name
-        ) and issubclass(_class, baseclass):
-            yield _class
+    return module
+
+
+def _find_classes(modules: Iterable[ModuleType], base: type[T]) -> Iterator[type[T]]:
```
raminqaf, line 83: Should we make a type aliase from it?
```suggestion
ClassIterator = Iterator[Type[T]]

#...
def _find_classes(modules: Iterable[ModuleType], base: type[T]) -> ClassIterator:
```

docs/docs/user/migration-guide/v6-v7.md
disrupted: good point, let's add this to the main docs together with an overall guide of how to build custom components

kpops/api/registry.py
disrupted: doesn't that make it a little harder to understand the function signature because it obfuscates the inner `T`? at least in my mind it's pretty straightforward with `T` in the parameters and the return

disrupted: ah, btw it would still need to be `ClassIterator[T]`

disrupted: perhaps we can find a better name for `T` though 😄

raminqaf: Yeah I think a better name for `T` would help here

kpops/api/registry.py
disrupted: I've changed the logging a bit to print out the module and filepath

kpops/api/registry.py
disrupted: done

docs/docs/user/migration-guide/v6-v7.md
disrupted: I've created a followup ticket for this #505

raminqaf (approved):
LGTM! Nice effort!

PR #501, opened 2024-06-27
Call destroy from inside of reset or clean

raminqaf:


tests/components/test_kafka_sink_connector.py
```diff
@@ -366,6 +372,7 @@ async def test_clean_when_dry_run_is_false(
 
         assert connector.to
         assert mock.mock_calls == [
+            mocker.call.destroy_connector(dry_run),
```
disrupted, line 375: do we really need to destroy the connector on reset? I thought resetting the offset was sufficient

raminqaf: That is a good point! I think we were doing that before. Since the rest API was calling destroy.

raminqaf: Getting back to your question. Yes deleting the connector is not necessary during reset

raminqaf: The comment is made on the test for cleaning the connectors. However, I have addressed your reviews and added a simple document for each method.

kpops/components/base_components/kafka_connector.py
```diff
@@ -229,10 +233,12 @@ def apply_from_inputs(self, name: str, topic: FromTopic) -> NoReturn:
 
     @override
     async def reset(self, dry_run: bool) -> None:
+        """Reset the state of a Kafka Connect source connector. The source connector is kept."""
         await self._resetter.reset(dry_run)
 
     @override
     async def clean(self, dry_run: bool) -> None:
+        """Delete the source connector. Reset the state of a Kafka Connect source connector."""
```
disrupted, line 241: how about a little more concise: "delete and reset state"? I think it's clear that it's about the connector since that's the component

kpops/components/base_components/pipeline_component.py
```diff
@@ -250,9 +250,11 @@ async def reset(self, dry_run: bool) -> None:
 
         :param dry_run: Whether to do a dry run of the command
         """
+        await self.destroy(dry_run)
```
disrupted, line 253: same question as for the connector, should we really destroy components on reset (by default)?

raminqaf: Good point. I moved the logic to the `StreamsBootstrap` class.

kpops/components/streams_bootstrap/producer/producer_app.py
```diff
@@ -95,4 +95,6 @@ def helm_chart(self) -> str:
 
     @override
     async def clean(self, dry_run: bool) -> None:
+        """Remove the producer app from the cluster. Deploy the cleanup job."""
```
disrupted, line 98:
```suggestion
        """Destroy and deploy cleanup job."""
```

disrupted (commented):
please have a look if you can make the other descriptions also a little more concise

docs/docs/user/migration-guide/v6-v7.md
```diff
@@ -60,3 +60,30 @@ As a result of the restructure, some imports need to be adjusted:
...
+
+Before v7, the KPOps CLI executed `destroy` before running `reset/clean` to ensure the component was destroyed.
+
+This logic has changed. From now on, the `destroy` method should be called within the `reset/clean` method if needed by the user.
+
+During the migration to v7, you should check your custom components and see if they are overriding the `reset/clean` methods. If so, you would need to add the `destroy` method at the beginning of your `reset/clean` override.
+
+#### components.py
+
+In the following example, we must call the `destroy` method inside the `clean` method. ** Otherwise, the logic of destroy will not be executed!**
```
disrupted, line 74: one extra space 🧐 
```suggestion
In the following example, we must call the `destroy` method inside the `clean` method. **Otherwise, the logic of destroy will not be executed!**
```

docs/docs/user/migration-guide/v6-v7.md
```diff
@@ -60,3 +60,30 @@ As a result of the restructure, some imports need to be adjusted:
...
+class MyCustomComponent(PipelineComponent):
+    
+    @override
+    async def destroy(self, dry_run: bool) -> None:
+        # Some custom destroy logic
+        # ...
+
+    @override
+    async def clean(self, dry_run: bool) -> None:
++       destroy(dry_run)
```
disrupted, line 86:
```suggestion
+       await destroy(dry_run)
```

docs/docs/user/migration-guide/v6-v7.md
disrupted, line 74: great writeup!

kpops/components/base_components/kafka_connector.py
```diff
@@ -181,12 +182,15 @@ async def deploy(self, dry_run: bool) -> None:
...
     async def destroy(self, dry_run: bool) -> None:
+        """Delete Kafka Connector (Source/Sink) from the Kafka connect cluster."""
         await self.handlers.connector_handler.destroy_connector(
             self.full_name, dry_run=dry_run
         )
 
     @override
     async def clean(self, dry_run: bool) -> None:
+        """Delete Kafka Connector. If schema handler is enabled, then remove schemas. Delete all the output topics."""
+        await self.destroy(dry_run)
```
disrupted, line 193: shouldn't it suffice to call `super().clean()` here?

kpops/components/common/streams_bootstrap.py
```diff
@@ -72,3 +73,11 @@ def warning_for_latest_image_tag(self) -> Self:
...
             )
         return self
+
+    @override
+    async def reset(self, dry_run: bool) -> None:
+        await self.destroy(dry_run)
+
+    @override
+    async def clean(self, dry_run: bool) -> None:
+        await self.destroy(dry_run)
```
disrupted, line 83: should we provide this as default implementation for `PipelineComponent` or are there any drawbacks?

kpops/components/base_components/kafka_connector.py
raminqaf: No, because we are not calling destroy in the superclass.

kpops/components/common/streams_bootstrap.py
raminqaf: I was considering doing so, and I think it is the correct way of handling `reset/clean.` Suppose you initiate a `clean` on your pipeline and have a `HelmApp` component. If we don't call it `destroy` in the `PipelineComponent` class, the component will not be uninstalled.

kpops/components/base_components/kafka_connector.py
raminqaf: Now I have changed the logic and called destroy inside of `PipelineComponent` there, so the `super().clean()` is sufficient.

docs/docs/user/migration-guide/v6-v7.md
```diff
@@ -60,3 +60,42 @@ As a result of the restructure, some imports need to be adjusted:
...
+class MyStreamsApp(StreamsApp):
+    
+    @override
+    async def clean(self, dry_run: bool) -> None:
++       await super().clean(dry_run)
+        # Some custom clean logic
+        # ...
+```
+
+If you implement the `PipelineComponent` you **must** call the `destroy` method inside the `clean` method.
```
disrupted, line 86: now we can change this paragraph, right?

kpops/components/streams_bootstrap/producer/producer_app.py
```diff
@@ -93,6 +93,15 @@ def add_extra_output_topic(self, topic: KafkaTopic, role: str) -> None:
     def helm_chart(self) -> str:
         return f"{self.repo_config.repository_name}/{AppType.PRODUCER_APP.value}"
 
+    async def reset(self, dry_run: bool) -> None:
+        """Not supported. Producer App cannot be reset.
```
disrupted, line 97: I think we need to reword this a little, e.g. producer app will only be destroyed. Reset is not necessary since it doesn't have consumer group offsets.

kpops/components/streams_bootstrap/streams/streams_app.py
```diff
@@ -28,8 +28,14 @@ class StreamsAppCleaner(KafkaAppCleaner):
     def helm_chart(self) -> str:
         return f"{self.repo_config.repository_name}/{AppType.CLEANUP_STREAMS_APP.value}"
 
+    @override
+    async def reset(self, dry_run: bool) -> None:
+        self.app.streams.delete_output = False
+        await super().clean(dry_run)
```
disrupted, line 34: shouldn't it be 
```suggestion
        await super().reset(dry_run)
```

raminqaf: This is a bit confusing, please have a look at [this code on main](https://github.com/bakdata/kpops/blob/main/kpops/components/streams_bootstrap/streams/streams_app.py#L132). Since the cleanup job is done by streams bootstrap we just need to set the delete output to flase and deploy the clean job.

docs/docs/user/migration-guide/v6-v7.md
raminqaf: I have changed it :)

kpops/components/streams_bootstrap/streams/streams_app.py
disrupted: I guess you're changing this anyways in #498 with the `reset` separate method so we can discuss it afterwards there 😄

docs/docs/user/migration-guide/v6-v7.md
```diff
@@ -60,3 +60,40 @@ As a result of the restructure, some imports need to be adjusted:
...
+
+Before v7, the KPOps CLI executed `destroy` before running `reset/clean` to ensure the component was destroyed.
+
+This logic has changed. The `destroy` method is now called within the `PipelineComponent`'s `reset/clean`.
+
+During migrating to v7, you should check your custom components and see if they override the `reset/clean` methods. If so, you need to call the supermethod `reset/clean` to trigger the `destroy` inside the parent class. Alternatively, if you are implementing the `PipelineComponent` class, you need to call the `destroy` method at the beginning of the method.
+
+#### components.py
+
+For example, when creating a custom `StreamsApp` or `ProducerApp` (or any other custom component), you **must** call the supermethod `rest/clean` to execute the `destroy` in the parent class. **Otherwise, the logic of destroy will not be executed!**
```
disrupted, line 74:
```suggestion
For example, when creating a custom `StreamsApp` or `ProducerApp` (or any other custom component), you **must** call the supermethod `reset`/`clean` to execute the `destroy` in the parent class. **Otherwise, the logic of destroy will not be executed!**
```

disrupted (changes requested):
typo

kpops/components/streams_bootstrap/streams/streams_app.py
raminqaf: yes, but I wanted to keep everything as is for this PR 😄

disrupted (approved):

PR #503, opened 2024-06-28
Generate developer docs for Python API

### LINKED ISSUE #492: Generate API documentation using `mkdocstrings`
Use https://github.com/mkdocstrings/mkdocstrings to create docs for the new's KPOps Python API

sujuka99:
Closes #492


docs/mkdocs.yml
```diff
@@ -37,6 +37,23 @@ plugins:
...
+          paths: [..]
+          options:
+            docstring_style: sphinx
+            show_if_no_docstring: true
+            line_length: 60
+            separate_signature: true
+            show_signature_annotations: true
+            docstring_section_style: spacy
+            show_root_heading: true
+            # The options below are not yet publicly available,
```
raminqaf, line 52: is there a link to this?

docs/docs/developer/api.md
```diff
@@ -0,0 +1,13 @@
...
+
+<!-- dprint-ignore-start -->
+
+::: kpops.api
+
+<!-- dprint-ignore-end -->
+
+<!-- dprint-ignore-start -->
+
+::: kpops.pipeline.Pipeline
```
raminqaf, line 11: I don't really the `Pipeline` class should be in the public API docs.

raminqaf:
Is there a way to omit `setup_handler` and `create_pipeline` from the API docs?

Besides, it would be nice to have some PyDocs under each method so it is reflected in the docs

image

docs/mkdocs.yml
sujuka99: Added links

docs/docs/developer/api.md
sujuka99: Removed

disrupted: yes, `Pipeline` can be made public imo

disrupted:
> Is there a way to omit `setup_handler` and `create_pipeline` from the API docs?
> 
> Besides, it would be nice to have some PyDocs under each method so it is reflected in the docs
> 
> image

rather than manually hiding those, I would prefer prefixing them with `_` to make them private and then only exporting the public ones.

raminqaf (approved):

PR #506, opened 2024-07-08
Rename app field

### LINKED ISSUE #502: Rename generic `app` config to something more specific
Currently, most components use `app` to specify its config. We should consider renaming it to something more specific so that the purpose becomes more clear.
Suggestions:
- `values` for Helm (and Kubernetes? I don't even understand what the config for Kubernetes does https://github.com/bakdata/kpops/blob/main/kpops/components/base_components/kubernetes_app.py#L21)
- `config` for Connectors (in line with https://docs.confluent.io/platform/current/connect/references/restapi.html#get--connectors-(string-name)-config)
- ...

disrupted:
Closes #502

connector: `config`
kubernetes-app (and children): `values`

global `KpopsConfig` and `ComponentHandlers` (naive singleton implementation), no longer needs to be passed during component initialization

examples test depends on https://github.com/bakdata/kpops-examples/pull/24


kpops/components/base_components/base_defaults_component.py
```diff
@@ -64,39 +65,35 @@ class BaseDefaultsComponent(DescConfigModel, ABC):
...
-        default=...,
-        description=describe_attr("config", __doc__),
-        exclude=True,
-    )
-    handlers: SkipJsonSchema[ComponentHandlers] = Field(
-        default=...,
-        description=describe_attr("handlers", __doc__),
-        exclude=True,
-    )
+    _config: KpopsConfig = PrivateAttr()
```
disrupted, line 68: necessary because otherwise it clashes with the newly renamed KafkaConnector config field

kpops/components/base_components/base_defaults_component.py
```diff
@@ -64,39 +65,35 @@ class BaseDefaultsComponent(DescConfigModel, ABC):
...
-                config=tmp_self.config,
-                handlers=tmp_self.handlers,
+                _config=_config,
+                _handlers=_handlers,
                 **values,
             )
         else:
             super().__init__(**values)
+            self._config = _config
+            self._handlers = _handlers
```
disrupted, line 96: this feels a little hacky but seems to do the job 🫠 
(via https://github.com/microsoft/pyright/issues/7209#issuecomment-1928292955)

disrupted: nvm, does not work for Kafka Connector `_resetter` computed field, because it needs access to those attributes

docs/docs/user/migration-guide/v6-v7.md
```diff
@@ -9,54 +9,94 @@ As a result of the restructure, some imports need to be adjusted:
...
++     KafkaSinkConnector,
++     KafkaSourceConnector,
++     KubernetesApp,
++     PipelineComponent,
++ )
++ from kpops.components.streams_bootstrap import (
++     StreamsBootstrap,
++     StreamsApp,
++     ProducerApp,
++ )
```
raminqaf, line 46: Why did they get formated? 🤔

kpops/component_handlers/__init__.py
```diff
@@ -11,6 +11,20 @@
...
+        cls,
+        schema_handler,
+        connector_handler,
+        topic_handler,
+        *args,
+        **kwargs,
+    ) -> ComponentHandlers:
+        if not cls._instance:
+            cls._instance = super().__new__(cls, *args, **kwargs)
+        return cls._instance
```
raminqaf, line 26: Singletone? 🥹

kpops/component_handlers/__init__.py
```diff
@@ -20,3 +34,10 @@ def __init__(
         self.schema_handler = schema_handler
         self.connector_handler = connector_handler
         self.topic_handler = topic_handler
+
+
+def get_handlers() -> ComponentHandlers:
```
raminqaf, line 39: Cannot be a propety?

kpops/config/__init__.py
```diff
@@ -162,3 +166,14 @@ def settings_customise_sources(
             dotenv_settings,
             file_secret_settings,
         )
+
+
+def get_config() -> KpopsConfig:
```
raminqaf, line 171: Same here with the property

kpops/config/__init__.py
```diff
@@ -162,3 +166,14 @@ def settings_customise_sources(
...
+    if not KpopsConfig._instance:
+        msg = f"{KpopsConfig.__name__} has not been initialized, call {KpopsConfig.__name__}.{KpopsConfig.create.__name__}"
+        raise RuntimeError(msg)
```
raminqaf, line 174: Instead of raising an error, wouldn't it be more convenient to initialize it?

tests/components/test_base_defaults_component.py
```diff
@@ -46,10 +46,11 @@ class EnvVarTest(BaseDefaultsComponent):
...
 
 
-@pytest.fixture()
-def config() -> KpopsConfig:
+@pytest.fixture(autouse=True)
+def config() -> None:
     ENV[PIPELINE_PATH] = str(RESOURCES_PATH / "pipeline.yaml")
-    return KpopsConfig(pipeline_base_dir=PIPELINE_BASE_DIR)
+    config = KpopsConfig(pipeline_base_dir=PIPELINE_BASE_DIR)
+    set_config(config)
```
raminqaf, line 53: If you initialize it on the get method you can omit these steps

raminqaf (commented):
Great effort! I left some comments! I prefer to have the singleton and the config/value renaming in two different PRs to see the diff a bit easier.

docs/docs/user/migration-guide/v6-v7.md
disrupted: I did that manually because I remembered that there should be one extra indentation space. I can move it to `v7` if it's confusing here, wdyt?

kpops/component_handlers/__init__.py
disrupted: yep, still feels a little like a naive singleton, but I consider this a valid solution

kpops/component_handlers/__init__.py
disrupted: property of ComponentHandlers? 🤔

kpops/config/__init__.py
disrupted: then we would need to pass the arguments for `KpopsConfig.create()`

kpops/component_handlers/__init__.py
raminqaf: Yes like this
```suggestion
@property
def component_handlers() -> ComponentHandlers:
```

disrupted: but it would need to be a classmethod property to be globally accessible which is deprecated in Python 3.11+. We have only have a workaround `cached_classproperty` which doesn't type check properly and I'd rather not use it more until there is a better approach.

kpops/config/__init__.py
raminqaf: Yeah that is true. I like it more like this.

raminqaf (approved):

raminqaf (commented):
The conflicts needs to be resolved 😃

raminqaf (approved):

raminqaf:
Let's go v7! 🚀 7️⃣

PR #514, opened 2024-07-18
KPOps 7.0

### LINKED ISSUE #502: Rename generic `app` config to something more specific

disrupted:
Depends on https://github.com/bakdata/kpops-examples/pull/24
Closes #502, #506, #498, #501, #500


raminqaf:
Closes #506, #498, #501, #500

raminqaf (approved):

raminqaf (dismissed):
I think we need to squash some of the commits for the changelog

PR #519, opened 2024-08-01
Add support for streams-bootstrap v3

raminqaf:
depends on https://github.com/bakdata/streams-bootstrap/pull/250/


examples
raminqaf, line 1: Strange that these changed 🤔

tests/api/test_registry.py
```diff
@@ -99,7 +102,9 @@ def test_registry():
         "kubernetes-app": KubernetesApp,
         "pipeline-component": PipelineComponent,
         "producer-app": ProducerApp,
+        "producer-app-v3": ProducerAppV3,
         "streams-app": StreamsApp,
+        "streams-app-v3": StreamsAppV3,
```
raminqaf, line 107: I had to suffix the streamApp and producerApp classes with v3 otherwise the registry would overwrite the streams bootstrap v2 classes

kpops/components/streams_bootstrap_v3/app_type.py
raminqaf, line 1: This is redundant, but I would still have it duplicated in this package to keep everything separate.

kpops/components/streams_bootstrap_v3/streams/model.py
```diff
@@ -0,0 +1,267 @@
...
+
+    @pydantic.field_validator("labeled_input_topics", mode="before")
+    @classmethod
+    def deserialize_labeled_input_topics(
+        cls, labeled_input_topics: dict[str, str] | Any
+    ) -> dict[str, list[KafkaTopic]] | Any:
+        if isinstance(labeled_input_topics, dict):
+            return {
+                label: [KafkaTopic(name=topic_name) for topic_name in topics]
+                for label, topics in labeled_input_topics.items()
```
disrupted, line 79: `topics: str` seems like a mistake

disrupted: most likely I made the mistake in https://github.com/bakdata/kpops/pull/515

tests/pipeline/resources/streams-bootstrap-v3/pipeline.yaml
```diff
@@ -0,0 +1,52 @@
...
+        memory: 2G
+  from:
+    topics:
+      my-input-topic:
+        type: input
+      my-labeled-input-topic:
+        role: my-input-topic-label
+      my-input-pattern:
+        type: pattern
+    # TODO: not working, for some reason the extra pattern is not serialized as Topic object but as str
```
disrupted, line 38: where do you get the error? do you have steps to reproduce?

tests/api/test_registry.py
disrupted: if streams-bootstrap v3 is going to become the new default, I would prefer keeping `streams-app` for v3 and adding the -v2 suffix to the old one (breaking change)

kpops/components/streams_bootstrap_v3/kafka_app.py
```diff
@@ -0,0 +1,150 @@
...
+
+class KafkaAppV3(PipelineComponent, ABC):
+    """Base component for Kafka-based components.
+
+    Producer or streaming apps should inherit from this class.
+
+    :param values: Application-specific settings
+    """
+
+    values: KafkaAppValues = Field(
```
philipp94831, line 136: What if we remove this from KafkaApp and then we don't need introduce KafkaAppV3? I don't think this is used anywhere directly, only via streams-bootstrap

kpops/components/streams_bootstrap_v3/producer/producer_app.py
```diff
@@ -0,0 +1,121 @@
...
+    OutputTopicTypes,
+    TopicConfig,
+)
+from kpops.components.streams_bootstrap_v3.app_type import AppType
+from kpops.components.streams_bootstrap_v3.kafka_app import KafkaAppV3
+from kpops.components.streams_bootstrap_v3.producer.model import ProducerAppValues
+from kpops.utils.docstring import describe_attr
+
+log = logging.getLogger("ProducerApp")
+STREAMS_BOOTSTRAP_V3 = "3.0.0"
```
philipp94831, line 22: 3.0.0 sadly won't happen

tests/pipeline/resources/streams-bootstrap-v3/pipeline.yaml
raminqaf: just uncomment it and run the newly added tests in manifest/generate.

kpops/components/streams_bootstrap_v3/producer/producer_app.py
raminqaf: Oh no! Why? what would be the version then?

philipp94831: We will need to discuss the version internally

kpops/components/streams_bootstrap_v3/producer/producer_app.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @property
+    @override
+    def helm_chart(self) -> str:
+        return (
+            f"{self.repo_config.repository_name}/{AppType.CLEANUP_PRODUCER_APP.value}"
+        )
+
+
+class ProducerAppV3(KafkaAppV3, StreamsBootstrap):
```
philipp94831, line 36:
```suggestion
class ProducerAppV3(KafkaAppV3, StreamsBootstrapV3):
```

kpops/components/streams_bootstrap_v3/streams/streams_app.py
```diff
@@ -0,0 +1,153 @@
...
+        pvc_handler = await PVCHandler.create(app_full_name, self.namespace)
+        if dry_run:
+            pvc_names = await pvc_handler.list_pvcs()
+            log.info(f"Deleting the PVCs {pvc_names} for StatefulSet '{app_full_name}'")
+        else:
+            log.info(f"Deleting the PVCs for StatefulSet '{app_full_name}'")
+            await pvc_handler.delete_pvcs()
+
+
+class StreamsAppV3(KafkaAppV3, StreamsBootstrap):
```
philipp94831, line 57:
```suggestion
class StreamsAppV3(KafkaAppV3, StreamsBootstrapV3):
```

philipp94831 (changes requested):

kpops/components/streams_bootstrap_v3/producer/producer_app.py
philipp94831: Also there should be some kind of test to ensure it is fixed I guess

raminqaf: I have merged `KafkaApp` with `StreamsBootstrap`, added the new configs and created a `StreamsBootstrapV3`

raminqaf: Single inheritance and clean abstraction.

philipp94831: But KafkaApp handles the deployment of topics and schemas so it should inherit from it

raminqaf: Wouldn't this be enough? 🤔 https://github.com/bakdata/kpops/pull/519/files#diff-1d74f42bef3cb5f575f8e945129fcefa298b3dba3c20036470f1819210866058R85

kpops/components/streams_bootstrap_v3/base.py
```diff
@@ -80,6 +81,17 @@ class StreamsBootstrapV3(HelmApp, ABC):
         description=describe_attr("version", __doc__),
     )
 
+    @override
+    async def deploy(self, dry_run: bool) -> None:
```
philipp94831, line 85: Why don't you reuse Kafka app?

kpops/components/streams_bootstrap_v3/producer/producer_app.py
philipp94831: I don't like it. Now you have the same logic in two places. KafkaApp was just fine. There was just no need to have values in it

kpops/components/streams_bootstrap_v3/base.py
```diff
@@ -0,0 +1,145 @@
...
+)
+STREAMS_BOOTSTRAP_VERSION = "3.0.0"
+
+# Source of the pattern: https://kubernetes.io/docs/concepts/containers/images/#image-names
+IMAGE_TAG_PATTERN = r"^[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}$"
+
+log = logging.getLogger("StreamsBootstrapV3")
+
+
+class StreamsBootstrapV3Values(HelmAppValues):
```
philipp94831, line 41: this should also have a field kafka with type KafkaConfig to centrally configure bootstrapservers and sr

philipp94831, line 41: Don't know if we want to do it in a follow up but we shoudl also add stuff like commandLine

kpops/components/streams_bootstrap_v3/streams/model.py
```diff
@@ -0,0 +1,266 @@
...
+        default=1,
+        title="Max replica count",
+        description=describe_attr("max_replicas", __doc__),
+    )
+    idle_replicas: int | None = Field(
+        default=None,
+        title="Idle replica count",
+        description=describe_attr("idle_replicas", __doc__),
+    )
+    topics: list[str] = Field(
```
philipp94831, line 188: What about internal topics?

philipp94831 (changes requested):

kpops/components/streams_bootstrap_v3/base.py
raminqaf: I agree with you. I would rather have it in a follow up PR.

raminqaf: Created an issue: https://github.com/bakdata/kpops/issues/520

kpops/components/streams_bootstrap_v3/streams/model.py
raminqaf: This should be renamed to `internal_topics`.

philipp94831: No, there is both

raminqaf: > No, there is both

Edited the message :D

tests/api/test_registry.py
raminqaf: I agree with you, but I would instead do it in a follow-up PR. We can create a new branch, v8, and do the changes step-by-step to keep the PRs small.

kpops/components/streams_bootstrap_v3/base.py
```diff
@@ -0,0 +1,171 @@
...
+        description=describe_attr("repo_config", __doc__),
+    )
+
+    version: str = Field(
+        default=STREAMS_BOOTSTRAP_VERSION,
+        pattern=STREAMS_BOOTSTRAP_VERSION_PATTERN,
+        description=describe_attr("version", __doc__),
+    )
+
+    @pydantic.model_validator(mode="after")
```
disrupted, line 88: shouldn't it be a field validator?
otherwise, it could also be builtin to the pattern where we only allow 3 as the major version

kpops/components/common/topic.py
```diff
@@ -60,7 +60,10 @@ class TopicConfig(DescConfigModel):
     configs: dict[str, str | int] = Field(
         default={}, description=describe_attr("configs", __doc__)
     )
+    # TODO: We can rename this but this would be a breaking change
```
philipp94831, line 63: Please rename. Breaking change is fine, we just need to be able to deploy streams-bootstrap v2 and v3 within the same project

raminqaf: I will do this in a separate PR that merges into v8. If I change this all the SNAPSHOTs are going to change and this PR becomes even bigger

kpops/components/streams_bootstrap_v3/base.py
raminqaf: I couldn't make it work with the field validator. The pattern didn't work either. :/

kpops/components/streams_bootstrap_v3/streams/model.py
```diff
@@ -0,0 +1,284 @@
...
+        description="Whether to use a Statefulset instead of a Deployment to deploy the streams app.",
+    )
+    persistence: PersistenceConfig = Field(
+        default=PersistenceConfig(),
+        description=describe_attr("persistence", __doc__),
+    )
+    model_config = ConfigDict(extra="allow")
+
+    @model_validator(mode="after")
+    def validate_mandatory_fields_are_set(
```
philipp94831, line 268: I appreciate the effort but shouldn't we delegate this validation to the helm chart?

raminqaf: As far as I remember, this check wasn't included in the Helm charts for [v2](https://github.com/bakdata/streams-bootstrap/blob/1a0f2a490e6d613fa4356cd42541009273bd511d/charts/streams-app/templates/scaled-object.yaml#L30).

Neither for [v3](https://github.com/bakdata/streams-bootstrap/blob/935b61d79f5022420a5add51975a9ffecb222292/charts/streams-app/templates/scaled-object.yaml#L30)

philipp94831: You are right but I think it is better to submit a PR there :)

raminqaf: https://github.com/bakdata/streams-bootstrap/pull/254

kpops/components/streams_bootstrap_v3/base.py
raminqaf: update: made it work 😉

kpops/components/streams_bootstrap_v3/streams/model.py
raminqaf: @philipp94831 Although you can argue that this makes it fail earlier in the generate step rather than the dry-run deployment step

kpops/components/streams_bootstrap_v3/streams/model.py
raminqaf: It was a typo by me. I forgot to update the field name in the `field_validator` decorator

tests/api/test_registry.py
```diff
@@ -101,6 +101,7 @@ def test_registry():
         "kafka-source-connector": KafkaSourceConnector,
         "kubernetes-app": KubernetesApp,
         "pipeline-component": PipelineComponent,
+        # TODO: change the old sterams bootstrap to -v2  and remove -v3
```
philipp94831, line 104: Will you open a follow up PR?

raminqaf: Yes I will :) I just don't want to make a branch from this PR and end up in rebase hell. I would rather do the reviews here. Merge it with v3 and then make a branch from v8 and continue with the TODOs.

kpops/components/streams_bootstrap_v3/__init__.py
```diff
@@ -0,0 +1,6 @@
+from kpops.components.common.streams_bootstrap import StreamsBootstrap
```
disrupted, line 1: shouldn't it export `StreamBootstrapV3`?

kpops/components/streams_bootstrap_v3/base.py
```diff
@@ -0,0 +1,84 @@
...
+    except ImportError:
+        from typing_extensions import Self
+
+
+STREAMS_BOOTSTRAP_HELM_REPO = HelmRepoConfig(
+    repository_name="bakdata-streams-bootstrap",
+    url="https://bakdata.github.io/streams-bootstrap/",
+)
+STREAMS_BOOTSTRAP_VERSION = "3.0.0"
+STREAMS_BOOTSTRAP_VERSION_PATTERN = r"^(\d+)\.(\d+)\.(\d+)(-[a-zA-Z]+(\.[a-zA-Z]+)?)?$"
```
disrupted, line 29: how about enforcing major version 3 like this?
```suggestion
STREAMS_BOOTSTRAP_VERSION_PATTERN = r"^(3)\.(\d+)\.(\d+)(-[a-zA-Z]+(\.[a-zA-Z]+)?)?$"
```

tests/components/streams_bootstrap_v3/test_producer_app_v3.py
disrupted, line 1: please check consistent naming. there is `test_producer_app_v3.py` but no `test_streams_app_v3.py`. I don't mind with or without the suffix since it's already in the folder name

disrupted (changes requested):

kpops/components/streams_bootstrap_v3/base.py
raminqaf: It works, but then we will lose the possibility of delivering a clear, easy-to-understand error message to the user. The message would look like this and IMO is harder to figure out what the problem is:

```
1 validation error for StreamsBootstrapV3\nversion\n  String should match pattern '^(3).(\\d+)\\.(\\d+)(-[a-zA-Z]+(\\.[a-zA-Z]+)?)?$' [type=string_pattern_mismatch, input_value='2.1.0', input_type=str]\n    For further information visit https://errors.pydantic.dev/2.6/v/string_pattern_mismatch
```

disrupted: gotcha, then I like your solution with the custom validator better 👍🏻

tests/components/streams_bootstrap_v3/test_producer_app_v3.py
raminqaf: It was typo. Corrected

kpops/components/streams_bootstrap_v3/__init__.py
raminqaf: Yes, corrected

kpops/components/streams_bootstrap_v3/base.py
```diff
@@ -0,0 +1,85 @@
...
+
+        if not pattern_match:
+            msg = f"Invalid version format: {version}"
+            raise ValueError(msg)
+
+        major, minor, patch, suffix, _ = pattern_match.groups()
+        major = int(major)
+
+        if major < 3:
+            msg = f"When using the streams bootstrap v3 component your version ('{version}') must be at least 3.0.0."
```
disrupted, line 74:
```suggestion
            msg = f"When using the streams-bootstrap v3 component your version ('{version}') must be at least 3.0.0."
```

kpops/components/streams_bootstrap_v3/base.py
```diff
@@ -0,0 +1,85 @@
...
+        pattern_match = COMPILED_VERSION_PATTERN.match(version)
+
+        if not pattern_match:
+            msg = f"Invalid version format: {version}"
+            raise ValueError(msg)
+
+        major, minor, patch, suffix, _ = pattern_match.groups()
+        major = int(major)
+
+        if major < 3:
```
disrupted, line 73: I'd say it's even safer to validate it is exactly 3. A newer major version could break the component

kpops/components/streams_bootstrap_v3/producer/producer_app.py
```diff
@@ -0,0 +1,120 @@
...
+    OutputTopicTypes,
+    TopicConfig,
+)
+from kpops.components.streams_bootstrap_v3.base import (
+    StreamsBootstrapV3,
+)
+from kpops.components.streams_bootstrap_v3.producer.model import ProducerAppValues
+from kpops.utils.docstring import describe_attr
+
+log = logging.getLogger("ProducerApp")
```
disrupted, line 20:
```suggestion
log = logging.getLogger("ProducerAppV3")
```

kpops/components/streams_bootstrap_v3/streams/model.py
```diff
@@ -0,0 +1,252 @@
...
+        cls, input_topics: list[str] | Any
+    ) -> list[KafkaTopic] | Any:
+        if isinstance(input_topics, list):
+            return [KafkaTopic(name=topic_name) for topic_name in input_topics]
+        return input_topics
+
+    @pydantic.field_validator("labeled_input_topics", mode="before")
+    @classmethod
+    def deserialize_labeled_input_topics(
+        cls, labeled_input_topics: Any
```
disrupted, line 72:
```suggestion
        cls, labeled_input_topics: dict[str, list[str]] | Any
```

kpops/components/streams_bootstrap_v3/streams/streams_app.py
```diff
@@ -0,0 +1,155 @@
...
+from kpops.components.common.topic import KafkaTopic
+from kpops.components.streams_bootstrap_v3.base import (
+    StreamsBootstrapV3,
+)
+from kpops.components.streams_bootstrap_v3.streams.model import (
+    StreamsAppValues,
+)
+from kpops.utils.docstring import describe_attr
+
+log = logging.getLogger("StreamsApp")
```
disrupted, line 20:
```suggestion
log = logging.getLogger("StreamsAppV3")
```

kpops/components/streams_bootstrap_v3/streams/streams_app.py
```diff
@@ -0,0 +1,155 @@
...
+
+STREAMS_BOOTSTRAP_V3 = "3.0.0"
```
disrupted, line 22: can't we import it from streams_bootstrap_v3/base?

disrupted (changes requested):

kpops/components/streams_bootstrap_v3/streams/streams_app.py
raminqaf: Improted

kpops/components/streams_bootstrap_v3/streams/model.py
raminqaf: good catch, corrected

kpops/components/streams_bootstrap_v3/streams/streams_app.py
```diff
@@ -67,7 +66,7 @@ class StreamsAppV3(StreamsBootstrapV3):
         description=describe_attr("values", __doc__),
     )
 
-    version: str | None = STREAMS_BOOTSTRAP_V3
+    version: str | None = STREAMS_BOOTSTRAP_VERSION
```
raminqaf, line 69: IMO we can also remove these fields from streams_app and producer_app. WDYT?

disrupted: yes 👍🏻

disrupted (approved):

PR #521, opened 2024-08-08
Fallback to user defined model when the validation of cluster model fails

raminqaf:


tests/components/test_producer_app.py
```diff
@@ -472,3 +472,74 @@ async def test_should_deploy_clean_up_job_with_values_in_cluster_when_clean(
             },
             HelmUpgradeInstallFlags(version="2.9.0", wait=True, wait_for_jobs=True),
         )
+
+    @pytest.mark.asyncio()
+    async def test_should_fall_back_to_local_values_when_validation_of_cluster_values_fails(
```
disrupted, line 477:
```suggestion
    async def test_clean_should_fall_back_to_local_values_when_validation_of_cluster_values_fails(
```

tests/components/test_streams_app.py
```diff
@@ -972,3 +966,87 @@ async def test_stateful_clean_with_dry_run_true(
             f"Deleting the PVCs {pvc_names} for StatefulSet '{STREAMS_APP_FULL_NAME}'"
             in caplog.text
         )
+
+    @pytest.mark.asyncio()
+    async def test_should_fall_back_to_local_values_when_validation_of_cluster_values_fails(
```
disrupted, line 971:
```suggestion
    async def test_clean_should_fall_back_to_local_values_when_validation_of_cluster_values_fails(
```

disrupted (approved):

PR #523, opened 2024-08-12
Fix incorrect parameter type annotation

disrupted:


kpops/components/streams_bootstrap/streams/model.py
```diff
@@ -65,7 +65,7 @@ def deserialize_input_topics(
     @pydantic.field_validator("extra_input_topics", mode="before")
     @classmethod
     def deserialize_extra_input_topics(
-        cls, extra_input_topics: dict[str, str] | Any
+        cls, extra_input_topics: dict[str, list[str]] | Any
```
raminqaf, line 68: Could we please validate this with a snapshot test? So in case we change it again the test would break

raminqaf: Or I think any type of test would be also sufficient

disrupted: there is no way to test it, the bug is only in the type checking

raminqaf (approved):

PR #525, opened 2024-08-12
Rename role to label

raminqaf:
depends on: https://github.com/bakdata/kpops-examples/pull/25

### DIFF
--- a/kpops/components/base_components/pipeline_component.py
+++ b/kpops/components/base_components/pipeline_component.py
@@ -110,10 +110,10 @@ class PipelineComponent(BaseDefaultsComponent, ABC):
         """
 
     def add_extra_input_topics(self, label: str, topics: list[KafkaTopic]) -> None:
-        """Add given extra topics that share a role to the list of extra input topics.
+        """Add given extra topics that share a label to the list of extra input topics.
 
         :param topics: Extra input topics
-        :param label: Topic role
+        :param label: Topic label
         """
 
     def set_input_pattern(self, name: str) -> None:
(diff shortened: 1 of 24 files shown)


kpops/components/base_components/pipeline_component.py
```diff
@@ -141,11 +141,11 @@ def set_error_topic(self, topic: KafkaTopic) -> None:
         :param topic: Error topic
         """
 
-    def add_extra_output_topic(self, topic: KafkaTopic, role: str) -> None:
+    def add_extra_output_topic(self, topic: KafkaTopic, label: str) -> None:
         """Add an output topic of type extra.
 
         :param topic: Output topic
-        :param role: Role that is unique to the extra output topic
+        :param label: Role that is unique to the extra output topic
```
disrupted, line 148: it still says role in the description 🤔

disrupted (changes requested):

kpops/components/base_components/pipeline_component.py
raminqaf: fixed! There should be no more `role` around. Appart from the ones that are the values of the test. I don't know if we want to change those

kpops/components/common/topic.py
```diff
@@ -61,7 +61,7 @@ class TopicConfig(DescConfigModel):
         default={}, description=describe_attr("configs", __doc__)
     )
     # TODO: We can rename this but this would be a breaking change
```
disrupted, line 63: todos resolved

disrupted (changes requested):

tests/components/streams_bootstrap_v3/test_streams_app.py
```diff
@@ -143,13 +143,13 @@ def test_set_topics(self):
                         "example-input": {"type": "input"},
                         "b": {"type": "input"},
                         "a": {"type": "input"},
-                        "topic-extra2": {"role": "role2"},
-                        "topic-extra3": {"role": "role2"},
-                        "topic-extra": {"role": "role1"},
+                        "topic-extra2": {"label": "role2"},
+                        "topic-extra3": {"label": "role2"},
+                        "topic-extra": {"label": "role1"},
```
disrupted, line 148: should we rename the values as well?

disrupted (approved):
one question, otherwise LGTM

tests/components/streams_bootstrap_v3/test_streams_app.py
raminqaf: We can, but these are just some values that don't have any meaning. Could be `foo` or `bar` or `baz`

Issue #538, opened 2024-10-24
Add ArgoCD `argocd.argoproj.io/sync-wave` annotation to generated Kubernetes resources


raminqaf:
Whenever the mode of KPOps is configured to `Argo` the manifestation process should add the `argocd.argoproj.io/sync-wave` annotation to determine the deployment order of the resources.

```
metadata:
  annotations:
    argocd.argoproj.io/sync-wave: "1"
```
All applications except for the KafkaTopics should have a sync-wave of `1`

raminqaf:
Closed with #541

PR #541, opened 2024-11-28
Introduce KPOps operation and manifest resources for deployment

### LINKED ISSUE #539: Add `operation_mode` to KPOps config
The user should be able to determine KPOps's "mode".
KPOps should run in two modes:
1. Argo: The commands (deploy, destroy, reset, and clean) will only manifest Kubernetes resources. See #536 
2. Manifest: only manifest the Helm charts
3. Default: How so far KPOps is handling the pipeline components

This can be passed the commands as an option `--operation-mode` or be configured through the `config.yaml` file

raminqaf:
Closes #539, #538


tests/conftest.py
```diff
@@ -56,6 +55,8 @@ def custom_components() -> Iterator[None]:
 
 @pytest.fixture(scope="module")
 def clear_kpops_config() -> Iterator[None]:
+    from kpops.config import KpopsConfig
```
raminqaf, line 58: The reason I changed this is circular dependency! 🙄

kpops/manifestors/streams_app_manifestor.py
```diff
@@ -0,0 +1,41 @@
...
+        if annotations:
+            values["annotations"] = annotations
+
+        return cleaner.helm.template(
+            cleaner.helm_release_name,
+            cleaner.helm_chart,
+            cleaner.namespace,
+            values,
+            cleaner.template_flags,
+        )
```
disrupted, line 41: this should be implemented in `cleaner.manifest()` and then called here

kpops/manifestors/helm_app_manifestor.py
```diff
@@ -0,0 +1,37 @@
...
+        if annotations:
+            values["annotations"] = annotations
+
+        return helm_app.helm.template(
+            helm_app.helm_release_name,
+            helm_app.helm_chart,
+            helm_app.namespace,
+            values,
+            helm_app.template_flags,
+        )
```
disrupted, line 37:
```suggestion
        return helm_app.manifest()
```

kpops/components/streams_bootstrap/streams/streams_app.py
```diff
@@ -152,3 +156,19 @@ async def clean(self, dry_run: bool) -> None:
         """Destroy and clean."""
         await super().clean(dry_run)
         await self._cleaner.clean(dry_run)
+
+    @override
+    def manifest_deploy(self) -> Resource:
+        manifest = super().manifest_deploy()
+        operation_mode = get_config().operation_mode
+
+        if operation_mode is OperationMode.ARGO:
```
disrupted, line 165: we don't need to check the operation mode here and again inside the manifestor, one check should be sufficient. But now that I see it I'll take my proposal back, I think the additional abstraction doesn't make that much sense as I thought. we can also discuss it more next week :)

kpops/api/operation.py
```diff
@@ -0,0 +1,9 @@
+from __future__ import annotations
+
+from enum import Enum
+
+
+class OperationMode(str, Enum):
+    ARGO = "argo"
+    MANIFEST = "manifest"
+    STANDARD = "standard"
```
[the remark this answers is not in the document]
raminqaf: That sounds great to me. The only thing that we should consider is the cleaner image tags. They need to be fetched from the Kubernetes cluster (if present) or default to the defined one in the `pipeline.yaml` or `defaults.yaml`.

kpops/components/streams_bootstrap/producer/producer_app.py
```diff
@@ -57,9 +74,11 @@ class ProducerApp(StreamsBootstrap):
     @computed_field
     @cached_property
     def _cleaner(self) -> ProducerAppCleaner:
-        return ProducerAppCleaner(
+        cleaner = ProducerAppCleaner(
             **self.model_dump(by_alias=True, exclude={"_cleaner", "from_", "to"})
         )
+        cleaner.values.name_override = None
```
raminqaf, line 80: Please refer to the function `to_helm_values()`.

kpops/cli/main.py
```diff
@@ -121,7 +127,7 @@ def init(
     path: Path = PROJECT_PATH,
     config_include_opt: bool = CONFIG_INCLUDE_OPTIONAL,
 ):
-    kpops.init(path, config_include_opt=config_include_opt)
+    kpops.init(path, config_include_optional=config_include_opt)
```
disrupted, line 130: please also rename the parameter of the parent function then

kpops/manifests/argo.py
```diff
@@ -0,0 +1,33 @@
...
+    POST_DELETE = "PostDelete"
+
+    @property
+    @override
+    def key(self) -> str:
+        return "argocd.argoproj.io/hook"
+
+
+class ArgoSyncWave(ArgoEnricher, str, enum.Enum):
+    SYNC_WAVE = "1"
```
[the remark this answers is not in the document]
raminqaf: Fair enough. I have changed it to a class that needs to be initialized with a sync_wave number

kpops/manifests/argo.py
```diff
@@ -0,0 +1,50 @@
+import enum
+from typing import Any
+
+from typing_extensions import override
+
+
+class ArgoEnricher:
+    @property
+    def key(self) -> str:
+        return NotImplemented
```
disrupted, line 10: I believe it should be `raise` instead of `return`

kpops/manifests/argo.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @property
+    @override
+    def key(self) -> str:
+        return "argocd.argoproj.io/hook"
+
+    @property
+    @override
+    def value(self) -> str:
+        return self.POST_DELETE._value_
```
disrupted, line 33:
```suggestion
        return self.value
```

kpops/manifests/argo.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class ArgoSyncWave(ArgoEnricher):
+    sync_wave: int
```
disrupted, line 37: the class var is unnecessary if you're assigning it as instance in `__init__` I believe

if possible I would prefer a proper dataclass/Pydantic model

disrupted (changes requested):

kpops/manifests/argo.py
raminqaf: The reason I did this is because `self.value` points to the value field and causes 
```
E   RecursionError: maximum recursion depth exceeded
!!! Recursion detected (same locals & position)
```

disrupted: oh ok, please avoid overriding the builtin `value` attribute of Enum

kpops/manifests/argo.py
```diff
@@ -0,0 +1,42 @@
+from __future__ import annotations
+
+import enum
+
+from pydantic import BaseModel, Field
+
+try:
+    from typing import Any, Self  # pyright: ignore[reportAttributeAccessIssue]
```
disrupted, line 8: looking forward to drop Python 3.10 support 😝

kpops/manifests/argo.py
```diff
@@ -0,0 +1,42 @@
...
+    def key(self) -> str:
+        return "argocd.argoproj.io/sync-wave"
+
+    @property
+    def value(self) -> str:
+        return str(self.sync_wave)
+
+    @classmethod
+    def create(cls, sync_wave: int) -> Self:
+        return cls(**{"SyncWave": sync_wave})
```
disrupted, line 42: not needed anymore?

kpops/manifests/argo.py
```diff
@@ -0,0 +1,42 @@
...
+class ArgoHook(str, enum.Enum):
+    POST_DELETE = "PostDelete"
+
+    @property
+    def key(self) -> str:
+        return "argocd.argoproj.io/hook"
+
+
+class ArgoSyncWave(BaseModel):
+    sync_wave: int = Field(default=0, serialization_alias="SyncWave")
```
disrupted, line 30: alias needed?
```suggestion
    sync_wave: int = 0
```

disrupted: is it possible to have negative sync waves? if not, I would be happy to further constrain this type, i.e. `pydantic.NonNegativeInt`

raminqaf: They can be negative :)

disrupted (approved):

PR #543, opened 2024-11-29
Fix: Use enum values when dumping models

raminqaf:


tests/pipeline/resources/streams-bootstrap/defaults.yaml
```diff
@@ -1,6 +1,7 @@
 streams-bootstrap:
   version: "3.0.3"
   values:
+    imagePullPolicy: IfNotPresent
```
disrupted, line 4: are you sure about this change? this deviates from the default defined in streams-bootstrap.

disrupted (commented):
could you briefly explain what the bug was?

tests/pipeline/resources/streams-bootstrap/defaults.yaml
raminqaf: I added the imagePullPolicy here only for testing. The problem was that when we called `kpops manifest ...` the values of the enums were not dumped as String. Here is the error:
```
yaml.representer.RepresenterError: ('cannot represent an object', <ImagePullPolicy.IF_NOT_PRESENT: 'IfNotPresent'>)
```

disrupted: gotcha! if you revert the change I'll approve :)

disrupted (approved):

PR #544, opened 2024-12-03
Convert all values of Kafka connector and topic config to string

disrupted:
Kafka Connect API interprets all config values as strings. Currently, it creates an unnecessary diff during `kpops deploy --dry-run` if the user doesn't specify all connector config values as strings. This is something we can do automatically.

image

### DIFF
--- a/kpops/component_handlers/kafka_connect/model.py
+++ b/kpops/component_handlers/kafka_connect/model.py
@@ -1,3 +1,4 @@
+import json
 from enum import Enum
 from typing import Any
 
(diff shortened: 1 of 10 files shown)


kpops/component_handlers/kafka_connect/model.py
```diff
@@ -74,15 +83,24 @@ def serialize_topics(self, topics: list[KafkaTopic]) -> str | None:
             return None
         return ",".join(topic.name for topic in topics)
 
+    @staticmethod
+    def serialize_to_str(value: Any) -> str:
+        if isinstance(value, bool):
```
philipp94831, line 88: Is it maybe cleaner to use json.dumps?

disrupted: yep, good idea!

kpops/utils/pydantic.py
```diff
@@ -42,6 +43,12 @@ def by_alias(model: BaseModel, field_name: str) -> str:
     return model.model_fields.get(field_name, Field()).alias or field_name
 
 
+def to_str(value: Any) -> str:
```
raminqaf, line 46: nitpicking: maybe add a tiny test for this method

raminqaf (approved):

PR #545, opened 2024-12-04
Manifest toSection with Strimzi Kafka topic

raminqaf:


kpops/manifests/strimzi/kafka_topic.py
```diff
@@ -0,0 +1,84 @@
...
+    spec: TopicSpec
+    status: dict[str, Any] | None = None
+
+    @classmethod
+    def create_strimzi_topic(cls, topic: KafkaTopic, bootstrap_servers: str) -> Self:
+        metadata = ObjectMeta.model_validate(
+            {
+                "name": topic.name,
+                "labels": {
+                    "strimzi.io/cluster": bootstrap_servers,
```
disrupted, line 70: are you sure about this? shouldn't it be the name of the Kafka cluster?

kpops/manifests/strimzi/kafka_topic.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+
+    api_version: str = "kafka.strimzi.io/v1beta2"
+    kind: str = "KafkaTopic"
+    metadata: ObjectMeta
+    spec: TopicSpec
+    status: dict[str, Any] | None = None
+
+    @classmethod
+    def create_strimzi_topic(cls, topic: KafkaTopic, bootstrap_servers: str) -> Self:
```
disrupted, line 65: how about naming it `from_topic`?

kpops/components/streams_bootstrap/base.py
```diff
@@ -81,3 +85,21 @@ def warning_for_latest_image_tag(self) -> Self:
                 f"The image tag for component '{self.name}' is set or defaulted to 'latest'. Please, consider providing a stable image tag."
             )
         return self
+
+    @override
+    def manifest_deploy(self) -> tuple[KubernetesManifest, ...]:
+        resource = super().manifest_deploy()
+        if self.to:
+            resource = resource + self.manifest_strimzi_topics(self.to.kafka_topics)
```
disrupted, line 93: no need to pass the argument to an instance method
```suggestion
            resource = resource + self.manifest_strimzi_topics()
```

disrupted (changes requested):

kpops/manifests/strimzi/kafka_topic.py
raminqaf: That is correct. This is the documentation of the Topic Operator. The `strimzi.io/cluster` must match the env var `STRIMZI_RESOURCE_LABELS`.

https://strimzi.io/docs/operators/latest/deploying#deploying-the-topic-operator-standalone-str

kpops/components/streams_bootstrap/base.py
raminqaf: The reason I did this was because of pyright. If I do the None check outside and use `self.to.kafka_topic` in the method it will throw me the error: 
```
error: "kafka_topics" is not a known member of "None" (reportOptionalMemberAccess)
```

kpops/manifests/strimzi/kafka_topic.py
disrupted: I still don't get it. according to the docs `STRIMZI_RESOURCE_LABELS` and `STRIMZI_KAFKA_BOOTSTRAP_SERVERS` are two independent settings

disrupted: ah, I didn't see that you changed it. looks good now

kpops/config/__init__.py
```diff
@@ -125,6 +125,10 @@ class KpopsConfig(BaseSettings):
         default=OperationMode.MANAGED,
         description="The operation mode of KPOps (managed, manifest, argo).",
     )
+    kafka_topic_resource_label: str | None = Field(
```
disrupted, line 128: if this is specific to Strimzi Kafka only, would it make sense to have a new config section and making it required there?

raminqaf: Done!

kpops/config/__init__.py
```diff
@@ -20,6 +22,28 @@
...
+    def deserialize_topics(cls, resource_label: Any) -> tuple[str, str]:
+        match resource_label:
+            case str(value) if "=" in value:
+                key, value = value.split("=")
+                return key, value
+            case dict(value) if len(value) == 1:
+                key, value = next(iter(value.items()))
+                return key, value
+            case _:
+                msg = "'resource_label' should be defined either like 'foo=bar' or as a valid dictionary."
```
disrupted, line 43: why not only allow dict then? Or define the required labels in the Pydantic model directly?

raminqaf: The configured `STRIMZI_RESOURCE_LABELS` on the topic operator can have any arbitrary key-value:
```
- bakdata.com/cluster=dev-cluster
- kpops.io/kafka-cluster=test-cluster
- strimizi-cluster=k8s-cluster
- foo=bar
```
Each topic operator must set a unique `<key>=<value>` pair. The `KafkaTopic` CRD must have the same label so that the topic operator can find the CRD and handle (create) the topic.

kpops/config/__init__.py
```diff
@@ -19,6 +20,32 @@
...
+
+    @property
+    def cluster_labels(self) -> tuple[str, str]:
+        """Return the defined strimzi_topic.label as a tuple."""
+        return next(iter(self.label_.items()))
+
+    @pydantic.field_validator("label_", mode="after")
+    @classmethod
+    def label_validator(cls, label: dict[str, str]) -> dict[str, str]:
+        if len(label) > 1:
```
disrupted, line 42: what if it's empty?

kpops/config/__init__.py
```diff
@@ -19,6 +20,32 @@
...
+
+    label_: dict[str, str] = Field(
+        alias="label",
+        description="The label to identify the KafkaTopic resources managed by the Topic Operator. This does not have to be the name of the Kafka cluster. It can be the label assigned to the KafkaTopic resource. If you deploy more than one Topic Operator, the labels must be unique for each. That is, the operators cannot manage the same resources.",
+    )
+
+    @property
+    def cluster_labels(self) -> tuple[str, str]:
+        """Return the defined strimzi_topic.label as a tuple."""
+        return next(iter(self.label_.items()))
```
disrupted, line 37: sweet! 👌🏻

disrupted (approved):

PR #546, opened 2024-12-04
Define Pydantic model to representing Kubernetes manifest

### LINKED ISSUE #548: Represent Kubernetes Resources as valid Pydantic models
Currently the manifest command is returning a type alias called [`Resource`](https://github.com/bakdata/kpops/blob/4a22ed809d60a3b562d8ee932f7054f87e61fd03/kpops/components/base_components/models/resource.py#L5). Resource is a sequence of dictionaries. There is no type validation. This should be changed. The new model will allow us to represent other Kubernetes resources like the StrimziKafkaTopic.

raminqaf:
Closes #548


kpops/components/common/kubernetes_model.py
```diff
@@ -1,12 +1,87 @@
...
+    @classmethod
+    def from_yaml(
+        cls, /, content: str
+    ) -> Iterator["KubernetesManifest"]:  # TODO: typing.Self for Python 3.11+
+        manifests: Iterator[dict[str, Any]] = yaml.load_all(content, yaml.Loader)
+        for manifest in manifests:
+            yield cls(**manifest)
+
+    @override
+    def model_dump(self, **_: Any) -> dict[str, Any]:
```
disrupted, line 77: why do we need this override?

kpops/components/common/kubernetes_model.py
```diff
@@ -1,12 +1,87 @@
...
+        None,
+        description="Deprecated field, not populated by Kubernetes in modern versions",
+    )
+    uid: str | None = None
+    model_config = ConfigDict(extra="allow")
+
+
+class KubernetesManifest(DescConfigModel):
+    # TODO: this is not working without the 'alias'
+    api_version: str = Field(alias="apiVersion")
```
disrupted, line 62: did you try inheriting from `CamelCaseConfigModel`?

kpops/component_handlers/helm_wrapper/helm.py
```diff
@@ -192,8 +190,10 @@ def template(
             ]
             command.extend(flags.to_command())
             output = self.__execute(command)
+            if output == "":
+                return ()
```
disrupted, line 194: what does `KubernetesManifest.from_yaml(output)` produce without handling this special case?

disrupted: also I am wondering in which case a Helm template would render nothing at all?!

kpops/component_handlers/helm_wrapper/helm_diff.py
```diff
@@ -16,7 +16,7 @@ def __init__(self, config: HelmDiffConfig) -> None:
     def calculate_changes(
         current_release: Iterable[HelmTemplate],
         new_release: Iterable[HelmTemplate],
-    ) -> Iterator[Change[KubernetesManifest, KubernetesManifest]]:
+    ) -> Iterator[Change[dict[str, Any], dict[str, Any]]]:
```
disrupted, line 19: do we really have to dump the dict here already? Can't we return the Pydantic model and let the caller decide what to do with it? Or is that impractical?

kpops/components/common/kubernetes_model.py
raminqaf: Yes, just corrected it. It works :)

kpops/components/common/kubernetes_model.py
raminqaf: I found it better to override this and enforce the exclusion settings in the base class instead of doing it separately everytime we call the method.

kpops/component_handlers/helm_wrapper/helm.py
raminqaf: It was my bad. We don't need this special case at all :)

kpops/component_handlers/helm_wrapper/helm_diff.py
raminqaf: The problem with returning the model is that we have no value for an empty state (`{}` or `KubernetesManifest()`) because the fields `api_version`, `kind`, `metaddata` are mandator.

kpops/component_handlers/helm_wrapper/helm_diff.py
```diff
@@ -32,13 +32,16 @@ def calculate_changes(
             # get corresponding dry-run release
             new_resource = new_release_index.pop(current_resource.filepath, None)
             yield Change(
-                current_resource.manifest,
-                new_resource.manifest if new_resource else KubernetesManifest(),
+                current_resource.manifest.model_dump(),
+                new_resource.manifest.model_dump() if new_resource else {},
```
raminqaf, line 36: @disrupted Here is where I return an empty dictionary. Before it was `KubernetesManifest()`. Now Pydantic will throw a `ValidatinError`

tests/component_handlers/helm_wrapper/test_helm_diff.py
```diff
@@ -18,11 +18,34 @@ def helm_diff(self) -> HelmDiff:
...
-        templates = [HelmTemplate(Path("a.yaml"), KubernetesManifest())]
+        templates = [
+            HelmTemplate(
+                Path("a.yaml"),
+                KubernetesManifest(
+                    **{
+                        "apiVersion": "v1",
+                        "kind": "Deployment",
+                        "metadata": {},
+                    }
```
raminqaf, line 29: @disrupted 
This is pretty okay here, tbh. We use this initializing type in other places in the tests (look streams_app or producer_app tests). Besides, if we want to initialize the metadata like `MetaObject(a=1)` pyright throws an error.

kpops/component_handlers/helm_wrapper/helm.py
```diff
@@ -20,13 +20,11 @@
     RepoAuthFlags,
     Version,
 )
-from kpops.component_handlers.kubernetes.model import KubernetesManifest
+from kpops.components.common.kubernetes_model import KubernetesManifest
```
disrupted, line 23: what's the reason for this move/rename?

kpops/components/common/kubernetes_model.py
```diff
@@ -1,12 +1,85 @@
...
-from kpops.utils.pydantic import DescConfigModel
+from kpops.utils.pydantic import CamelCaseConfigModel, DescConfigModel
 
 # Matches plain integer or numbers with valid suffixes: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory
 MEMORY_PATTERN = r"^\d+([EPTGMk]|Ei|Pi|Ti|Gi|Mi|Ki)?$"
+K8S_LABEL_MAX_LEN = 63
+
+
+class ManagedFieldsEntry(CamelCaseConfigModel):
+    # Define this class based on its actual structure
```
disrupted, line 18: is that a todo?

kpops/components/common/kubernetes_model.py
```diff
@@ -1,12 +1,85 @@
...
+    managed_fields: list[ManagedFieldsEntry] | None = None
+    name: str | None = None
+    namespace: str | None = None
+    owner_references: list[OwnerReference] | None = None
+    resource_version: str | None = None
+    self_link: str | None = Field(
+        default=None,
+        description="Deprecated field, not populated by Kubernetes in modern versions",
+    )
+    uid: str | None = None
```
disrupted, line 56: do we need all of those? I feel like many have no meaning to KPOps

kpops/component_handlers/helm_wrapper/helm.py
raminqaf: I wonder if the `KubernetesManifest` should be under `component_handlers`

kpops/components/common/kubernetes_model.py
raminqaf: They are a representation of the lightkube ObjectMeta. They are documented here:

https://gtsystem.github.io/lightkube-models/1.19/models/meta_v1/#objectmeta

We can have them removed and allow extra fields.

kpops/components/common/kubernetes_model.py
raminqaf: GPT comment 😄 will have it removed.

kpops/components/common/kubernetes_model.py
```diff
@@ -1,12 +1,87 @@
...
+        manifests: Iterator[dict[str, Any]] = yaml.load_all(content, yaml.Loader)
+        for manifest in manifests:
+            yield cls(**manifest)
+
+    @override
+    def model_dump(self, **_: Any) -> dict[str, Any]:
+        return super().model_dump(
+            mode="json",
+            by_alias=True,
+            exclude_unset=True,
```
raminqaf, line 83: @disrupted 
I was thinking about the `exclude_unset` here. This can cause problems when we define a model lime `StrimziKafkaTopic` from the `KubernetesManifest` class.

Imagine the model is defined as follows:

```python
class StrimziKafkaTopic(KubernetesManifest):
	api_version: str = "v1/beta2"
	kind = "KafkaTopic"
	
	# Rest of the fields...
	spec = TopicSpec
	metadata: ObjectMeta
```

If I create a model and dump it. The `exclude_unset` would not dump the two fields `api_version` and `kind`. 

If we remove that one, all the noise from the ObjectMeta will be dumped with it.

disrupted: good point, I think we would need a custom `@model_serializer` then. Afaict there is no builtin way to specify which default values to always include when `exclude_unset=True`

kpops/components/common/kubernetes_model.py
disrupted: yeah, I think I would prefer a reduction to those fields that are most likely to be needed in KPOps

kpops/component_handlers/helm_wrapper/helm.py
disrupted: perhaps it some point when we would want to support CRDs it would make sense to make a separate module or namespace package such as `kpops.manifests`. it's ok like this for now

tests/pipeline/test_manifest.py
```diff
@@ -123,15 +123,19 @@ def test_manifest_command(self, snapshot: Snapshot):
...
-    def test_python_api(self, snapshot: Snapshot):
+    def test_python_api(self):
         generator = kpops.manifest_deploy(
             RESOURCE_PATH / "manifest-pipeline" / PIPELINE_YAML,
             environment="development",
         )
         assert isinstance(generator, Iterator)
         resources = list(generator)
         assert len(resources) == 2
-        snapshot.assert_match(yaml.dump_all(resources), "resources")
```
disrupted, line 134: the new isinstance check is great, but I think we would still need some sort of snapshot testing as well. how about we just save the output of `print_yaml(manifest.model_dump)`?

raminqaf: Added it :)

kpops/component_handlers/helm_wrapper/helm.py
raminqaf: moved everything under `kpops.manifests`

tests/pipeline/test_manifest.py
disrupted: I agree with you. we should probably split that up into two separate tests. In the API test we run assertions on the returned `KubernetesManifest` models and in the CLI test we do the snapshot test of the output. Let's do it in a followup then

disrupted (approved):

PR #552, opened 2024-12-10
Manifest Kubernetes resources for `destroy` command

### LINKED ISSUE #553: Manifest Kubernetes resources for destroy command
The manifestation of Kubernetes resources with the `destroy` command is the same for both `argo` and `manifest` operation modes. In both operation modes __ONLY__ the Kafka Topics should be manifested.

raminqaf:
closes #553


kpops/components/streams_bootstrap/base.py
```diff
@@ -96,3 +96,11 @@ def manifest_deploy(self) -> tuple[KubernetesManifest, ...]:
...
 
         return resource
+
+    @override
+    def manifest_destroy(self) -> tuple[KubernetesManifest, ...]:
+        return (
+            tuple(StrimziKafkaTopic.from_topic(topic) for topic in self.to.kafka_topics)
+            if self.to
+            else ()
+        )
```
disrupted, line 106: I would prefer early return pattern to make this a little easier to read

```suggestion
    if not self.to:
        return ()
    return tuple(StrimziKafkaTopic.from_topic(topic) for topic in self.to.kafka_topics)
```

also what about `super()` call?

disrupted (approved):

PR #561, opened 2024-12-12
Drop support for Python 3.10

disrupted:


.github/actions/update-docs/action.yml
```diff
@@ -27,8 +27,8 @@ runs:
     - name: Install Python and set up Poetry
       uses: bakdata/ci-templates/actions/python-setup-poetry@v1.5.2
       with:
+        python-version: "3.12"
```
raminqaf, line 30: Maybe it would be better to run the python actions in the lowest supported python (3.11)

raminqaf (changes requested):

.github/actions/update-docs/action.yml
disrupted: what's the advantage?

raminqaf (approved):

PR #564, opened 2024-12-12
Create generic `SerializeAsOptional` type for Pydantic

disrupted:


philipp94831 (commented):
Can you also add this to StreamsBootstrapValues?

philipp94831:
> Can you also add this to StreamsBootstrapValues?

And also for `Affinity`

kpops/components/streams_bootstrap/model.py
```diff
@@ -221,8 +222,8 @@ class StreamsBootstrapValues(HelmAppValues):
         description=describe_attr("affinity", __doc__),
     )
 
-    tolerations: list[Toleration] | None = Field(
-        default=None,
+    tolerations: list[Toleration] = Field(
```
philipp94831, line 225: You need SerializeAsOptional here, right?

philipp94831 (changes requested):

kpops/components/streams_bootstrap/model.py
disrupted: yep, missed that one

tests/utils/test_pydantic.py
```diff
@@ -64,3 +66,13 @@ def test_to_dot(input: str, expected: str):
...
+
+
+def test_serialize_as_optional():
+    class Model(BaseModel):
+        foo: SerializeAsOptional[list[str]] = []
+
+    assert Model().model_dump() == {"foo": None}
+    assert Model().model_dump(exclude_defaults=True) == {}
+    assert Model().model_dump(exclude_unset=True) == {}
+    assert Model().model_dump(exclude_none=True) == {}
```
raminqaf, line 78: If I understand the tests correctly, the field `foo` is only dumped as `"foo": None`. Will it ever be dumped as an empty list in this test-case?

disrupted: yep, that's exactly the point of the `SerializeAsOptional` type. If the value is falsy, i.e. collection is empty, it will be serialized to `None`

kpops/components/streams_bootstrap_v2/base.py
```diff
@@ -17,9 +17,11 @@
 from kpops.utils.pydantic import (
     CamelCaseConfigModel,
     DescConfigModel,
+    SerializeAsOptionalModel,
     exclude_by_value,
     exclude_defaults,
 )
+from tests.utils.test_pydantic import SerializeAsOptional
```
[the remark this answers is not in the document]
disrupted: my bad (do not blindly accept import suggestions) 😑

philipp94831 (approved):

raminqaf (approved):

PR #571, opened 2024-12-19
Hide `operation_mode` from KPOps config

raminqaf:


kpops/utils/gen_schema.py
```diff
@@ -1,3 +1,5 @@
+# type: ignore[reportGeneralTypeIssues]
```
raminqaf, line 1: pyright is breaking on CI: https://github.com/bakdata/kpops/actions/runs/12412577169/job/34652661630?pr=571

It passes locally on my machine.

poetry.lock
raminqaf, line 1: I have reverted this file to your last working commit: https://github.com/bakdata/kpops/pull/558/commits/28f3e294e3a3aa304634f9c110dd77a62de8b97f#diff-f53a023eedfa3fbf2925ec7dc76eecdc954ea94b7e47065393dbad519613dc89

kpops/utils/gen_schema.py
disrupted: is that after you reverted poetry.lock?

raminqaf: Yes. :/

disrupted: but why? you didn't make any changes to this file

raminqaf: I have no clue why. I couldn't find a good answer. The errors are also very generic, and I don't really understand why they are being triggered.

disrupted: I think you need to revert poetry.lock to `main` and run `poetry lock --no-update`

poetry.lock
disrupted: please make this change in #558

kpops/utils/gen_schema.py
raminqaf: The poetry.lock file keeps being overriten. Even if I delete the `.venv` folder and revert the `poetry.lock` file and run `poetry lock --no-update`.

disrupted: some changes to `poetry.lock` are to be expected from the update to Python 3.11, but none should be regarding Pyright or Pydantic
do you see the same errors when you run `pyright kpops/utils/gen_schema.py`? because I don't. Only with `pre-commit run pyright -a`

raminqaf: If I run the command `pre-commit run --all-files` I get the errors

disrupted (approved):

PR #578, opened 2025-01-07
Migrate from Poetry to uv

disrupted:


.github/workflows/ci.yaml
```diff
@@ -12,96 +12,173 @@ defaults:
...
+          snapshot_time=$(python -c "from datetime import datetime; time = datetime.now(); print(time.strftime('%Y%m%d%H%M%S'))")
+          echo "snapshot-time=${snapshot_time}" >> "$GITHUB_OUTPUT"
+
+      - name: Bump version patch
+        id: bump-version
+        uses: bakdata/ci-templates/actions/bump-version@1.50.4
+        with:
+          release-type: patch
+
+      - run: git restore .
```
yannick-roeder, line 169: Wait, so you are bumping and then restoring? Are you pushing in between or what is the purpose of doing this?

.github/workflows/release.yaml
```diff
@@ -15,12 +15,10 @@ on:
...
   create-github-release-push-tag:
-    uses: bakdata/ci-templates/.github/workflows/python-poetry-release.yaml@1.40.4
+    uses: bakdata/ci-templates/.github/workflows/bump-version-release.yaml@1.51.0
     name: Release
     with:
       release-type: ${{ inputs.release-type }}
-      python-version: "3.11"
-      poetry-version: "1.7.1"
       changelog: true
       changelog-config: "./.github/changelog-config.json"
```
yannick-roeder, line 23: This can be removed now since the changelog is generated with [Git Cliff](https://github.com/orhun/git-cliff) now. I guess we forgot to remove [the input in the workflow](https://github.com/bakdata/ci-templates/blob/1.51.0/.github/workflows/bump-version-release.yaml#L21-L25), but it is never being used.

yannick-roeder (dismissed):

.github/workflows/ci.yaml
disrupted: this is needed otherwise bumpversion (in the configuration we use) will complain that the Git working directory is unclean. We only need the new version result after the first `bumpversion patch` call here which is stored in a step output

disrupted: I've added a comment

yannick-roeder: Shouldn't `--allow-dirty` do the trick?

disrupted: it might, but I don't see this input parameter for the bump-version action. I could use a hacky workaround and pass it as `release-type: patch --allow-dirty` or we could add it but personally I don't mind the `git restore` here.

docs/docs/user/references/cli-commands.md
```diff
@@ -125,41 +141,25 @@ $ kpops generate [OPTIONS] PIPELINE_PATHS...
...
 
 ```console
-$ kpops reset [OPTIONS] PIPELINE_PATHS...
+$ kpops clean [OPTIONS] PIPELINE_PATHS...
 ```
 
 **Arguments**:
 
-* `PIPELINE_PATHS...`: Paths to dir containing 'pipeline.yaml' or files named 'pipeline.yaml'.  [env var: KPOPS_PIPELINE_PATHS;required]
+* `PIPELINE_PATHS...`: Paths to dir containing &#x27;pipeline.yaml&#x27; or files named &#x27;pipeline.yaml&#x27;.  [env var: KPOPS_PIPELINE_PATHS; required]
```
disrupted, line 162: FIXME: single quotes are rendered incorrectly here

disrupted: nvm, the Markdown renders correctly

pyproject.toml
```diff
@@ -1,12 +1,11 @@
...
 version = "9.1.0"
 description = "KPOps is a tool to deploy Kafka pipelines to Kubernetes"
-authors = ["bakdata <opensource@bakdata.com>"]
-license = "MIT"
+authors = [
+  { name = "bakdata", email = "opensource@bakdata.com" },
+]
 readme = "README.md"
-repository = "https://github.com/bakdata/kpops"
-documentation = "https://bakdata.github.io/kpops/latest"
```
JakobEdding, line 9: These are missing now, also `license`

pyproject.toml
```diff
@@ -16,55 +15,67 @@ classifiers = [
   "Topic :: Software Development",
   "Intended Audience :: Developers",
   "Typing :: Typed",
+  "License :: OSI Approved :: MIT License",
+]
+requires-python = ">=3.11"
```
JakobEdding, line 20: Are you intentionally allowing 3.13 now?

pyproject.toml
```diff
@@ -16,55 +15,67 @@ classifiers = [
...
+dependencies = [
+  "anyio>=4.3.0",
```
JakobEdding, line 22: Does this include anyio 5.x.x now as well? With the old Poetry `^4.3.0` version specification, the version would be constrained to `<5.0.0`. Similarly for all other dependencies.

pyproject.toml
```diff
@@ -16,55 +15,67 @@ classifiers = [
...
-mkdocs-exclude-search = "^0.6.5"
-mike = "^1.1.2"
-mkdocstrings = { extras = ["python"], version = "^0.25.1" }
+[tool.setuptools.packages.find]
+include = ["kpops"]
 
-[tool.poetry_bumpversion.file."kpops/const/__init__.py"]
+[tool.pytest.ini_options]
+asyncio_mode = "auto"
+asyncio_default_fixture_loop_scope = "function"
```
JakobEdding, line 78: These settings are new afaics. We can remove markers `@pytest.mark.asyncio` from tests in this case I think.

.github/workflows/ci.yaml
```diff
@@ -12,96 +12,188 @@ defaults:
...
+        uses: astral-sh/setup-uv@v5
+        with:
+          python-version: "3.11"
+          version: ${{ env.UV_VERSION }}
+
+      - name: Generate snapshot time
+        id: snapshot-time
+        run: |
+          snapshot_time=$(python -c "from datetime import datetime; time = datetime.now(); print(time.strftime('%Y%m%d%H%M%S'))")
+          echo "snapshot-time=${snapshot_time}" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
disrupted: fine by me but we should change it upstream too then: https://github.com/bakdata/ci-templates/blob/b3a514a630940728ad27d5c959af8f9c2e2e98de/actions/python-poetry-bump-version/action.yaml#L52

.github/workflows/ci.yaml
```diff
@@ -12,96 +12,188 @@ defaults:
...
         run: |
-          if [[ "$RUNNER_OS" == "Linux" && "${{ matrix.python-version }}" == "3.11" ]]
+          if [[ "${{ matrix.format-for-github }}" == "true" ]]
           then
             echo "::add-matcher::.github/pyright-matcher.json"
-            poetry run pre-commit run pyright --all-files
-          else
-            poetry run pre-commit run pyright --all-files
           fi;
+          uv run --frozen pyright
```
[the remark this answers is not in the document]
JakobEdding: Ah yes, sorry I overlooked that.

.github/workflows/publish.yaml
```diff
@@ -6,10 +6,21 @@ on:
...
+    name: Publish to PyPI
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          python-version: "3.11"
+          version: "0.5.14"
```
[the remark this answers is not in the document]
disrupted: I thought of that too and briefly added it to the repository environment. A huge drawback of it is that it's not part of VCS then, so if someone bumps the version and things break we don't know the previous version and cannot revert. That's why I changed it back. I don't really know if there are better practices to set an env variable across multiple workflows

pyproject.toml
disrupted: yes, these two are Poetry-specific and don't exist in the offical pyproject.toml spec. I went with the recommendation to set MIT license as classifier instead since it's standardized.

pyproject.toml
disrupted: yep, one of the maintainer of uv recommends not setting an upper bound. I don't have a link right now to the comment but it's on their GitHub

.github/workflows/ci.yaml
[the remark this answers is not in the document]
JakobEdding: Fine for you @yannick-roeder ?

.github/workflows/publish.yaml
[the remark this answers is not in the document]
JakobEdding: Good point! [There doesn't seem to be a better way.](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#defining-configuration-variables-for-multiple-workflows)

pyproject.toml
disrupted: found it: https://github.com/astral-sh/uv/issues/6783#issuecomment-2316648253

docs/docs/schema/config.json
```diff
@@ -203,11 +203,7 @@
             "type": "boolean"
         },
         "helm_config": {
-            "allOf": [
-                {
-                    "$ref": "#/$defs/HelmConfig"
-                }
-            ],
+            "$ref": "#/$defs/HelmConfig",
```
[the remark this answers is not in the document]
disrupted: exactly, creating the new lockfile installed newer versions of the dependencies, such as Pydantic in this case. It's fine to drop these descriptions because they are defined in the referenced object.

pyproject.toml
disrupted: yes, all upper constraints are dropped. The uv maintainers have compiled a list of reasons why this is best practice for the Python package ecosystem https://github.com/zanieb/poetry-relax#references

.github/pyright-matcher.json
```diff
@@ -4,12 +4,13 @@
             "owner": "pyright",
             "pattern": [
                 {
-                    "regexp": "^\\s\\s(\\S*\\.py):(\\d+):(\\d+)\\s-\\s(error|warning|information):([\\s\\S]+?)(?=(?:^\\s\\s\\S*\\.py:\\d+:\\d+\\s-\\s(?:error|warning|information):|\\z|\\d+ errors,))",
+                    "regexp": "^\\s\\s(\\S*\\.py):(\\d+):(\\d+)\\s-\\s(error|warning|information):\\s([\\s\\S]+?)\\s\\(([\\s\\S]+?)(?=(?:\\)|\\d+ errors,))",
```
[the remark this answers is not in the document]
disrupted: #579

pyproject.toml
JakobEdding: More asking generally and out of curiosity now: Is uv greedy then and tries to set up the project with the latest Python stable version when I clone an existing project for the first time? Or can uv be configured globally to try to use the latest Python stable version up to and including e.g. 3.12?

pyproject.toml
JakobEdding: Ok, I just imagine it to be a hassle to update dependencies in practice 😅 Will try to read a bit in the links from poetry-relax

pyproject.toml
disrupted: AFAIK you can do `uv python pin 3.12`. I haven't tested it thoroughly so far

.github/workflows/ci.yaml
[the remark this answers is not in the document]
JakobEdding: I opened a PR here now: https://github.com/bakdata/ci-templates/pull/237

JakobEdding (approved):

PR #580, opened 2025-01-10
feat: support symbolic linked pipelines to avoid repetition

daconstenla:
## why 

While using kpops and trying to reduce the amount of copy/pasting required to configure multi-tenant configuration I found that _symbolic links_ would cause kpops execution to break.

## what 

Replaces existing `Path.blob` with `blob.blob` and resolves the real file location so no other changes are required for it to be compatible with existing logic.

> [!NOTE]
> In python 3.13 symbolic links are supported [docs.python.org/3.13#pathlib.Path.glob](https://docs.python.org/3.13/library/pathlib.html#pathlib.Path.glob), probably it would be as easy as follows:
> `yield from sorted(pipeline_path.glob(f"**/{PIPELINE_YAML}", recurse_symlinks=True))`

### DIFF
--- a/kpops/cli/utils.py
+++ b/kpops/cli/utils.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from collections.abc import Iterable, Iterator
 from pathlib import Path
+from glob import glob
 
 from kpops.const.file_type import PIPELINE_YAML
 
@@ -20,7 +21,11 @@ def collect_pipeline_paths(pipeline_paths: Iterable[Path]) -> Iterator[Path]:
         if pipeline_path.is_file():
             yield pipeline_path
         elif pipeline_path.is_dir():
-            yield from sorted(pipeline_path.glob(f"**/{PIPELINE_YAML}"))
+            # TODO: In python 3.13 symbolic links become supported by Path.glob
+            # docs.python.org/3.13#pathlib.Path.glob, probably it make sense to use it after ugprading,
+            # likely the code will look like:
+            # yield from sorted(pipeline_path.glob(f"**/{PIPELINE_YAML}", recurse_symlinks=True))
+            yield from sorted(Path(p).resolve() for p in glob(f"{pipeline_path}/**/{PIPELINE_YAML}", recursive=True))
         else:
             msg = f"The entered pipeline path '{pipeline_path}' should be a directory or file."
             raise ValueError(msg)
--- a/tests/pipeline/resources/pipeline-folders-with-symlinks/pipeline-1/pipeline.yaml
+++ b/tests/pipeline/resources/pipeline-folders-with-symlinks/pipeline-1/pipeline.yaml
@@ -1,0 +1,1 @@
+../../pipeline-folders/pipeline-1/pipeline.yaml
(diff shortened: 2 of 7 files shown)


disrupted:
> Note
> 
> In python 3.13 symbolic links are supported [docs.python.org/3.13#pathlib.Path.glob](https://docs.python.org/3.13/library/pathlib.html#pathlib.Path.glob), probably it would be as easy as follows: `yield from sorted(pipeline_path.glob(f"**/{PIPELINE_YAML}", recurse_symlinks=True))`

that's a good head-up. how about a short todo comment in the code to refactor this once we drop support for older Python versions?

disrupted (changes requested):
please consider adding a test for it. If you need any help let me know

daconstenla:
> > Note
> > In python 3.13 symbolic links are supported [docs.python.org/3.13#pathlib.Path.glob](https://docs.python.org/3.13/library/pathlib.html#pathlib.Path.glob), probably it would be as easy as follows: `yield from sorted(pipeline_path.glob(f"**/{PIPELINE_YAML}", recurse_symlinks=True))`
> 
> that's a good head-up. how about a short todo comment in the code to refactor this once we drop support for older Python versions?

Sure I can add a note, I actually had one originally but I moved it to the description of the pull-request.

disrupted:
> > please consider adding a test for it. If you need any help let me know
> 
> How would you suggest we test it? creating a pipeline that's just a symlink of another and expect same result? 🤔

yes. feel free to commit the symlink to https://github.com/bakdata/kpops/tree/main/tests/pipeline/resources

daconstenla:
> the new test itself would make sense in https://github.com/bakdata/kpops/blob/main/tests/pipeline/test_generate.py
> 
> edit: I believe ideally we would need to cover the following scenarios:
> 
> * `kpops generate pipelines/regular-folder/symlinked-pipeline.yaml`
> * `kpops generate pipelines/symlinked-folder`
> * `kpops generate pipelines` where `pipelines` contains symlinked folders
> 
> wdyt?

Mmmm, ok, more than I thought, let me try to cover those scenarios.

daconstenla:
> > the new test itself would make sense in https://github.com/bakdata/kpops/blob/main/tests/pipeline/test_generate.py
> > edit: I believe ideally we would need to cover the following scenarios:
> > 
> > * `kpops generate pipelines/regular-folder/symlinked-pipeline.yaml`
> > * `kpops generate pipelines/symlinked-folder`
> > * `kpops generate pipelines` where `pipelines` contains symlinked folders
> > 
> > wdyt?
> 
> Mmmm, ok, more than I thought, let me try to cover those scenarios.

As suggested I've included 3 types of symlinks to the tests:

- symlinked pipeline
- symlinked folder
- folder with symlinked folder and symlinked pipeline as well as a regular pipeline

the test just validates the rendered results is as the original sources of the symlinks.

disrupted (approved):
LGTM, thanks for your contribution

daconstenla:
@disrupted I assume you'll merge it when ready, thank you for the inputs.

PR #581, opened 2025-01-10
feat: support custom namespace configuration for `kafkatopic`s

daconstenla:
`strimzi-topic-operator` does not have support to monitor all namespaces in a kubernetes cluster https://github.com/strimzi/strimzi-kafka-operator/issues/1206.

Depending on the use case, it might be useful to be able to configure the namespace where `kafkatopic`s will be deployed which not necessarily will be the same namespace where streams and other components are deployed.


daconstenla:
@disrupted could you help me understand why, even though `namespace` is set to `None` it gets manifested as `metadata.namespace: null`? From what I understand all `None` values are removed from the final manifestations.

daconstenla:
> @disrupted could you help me understand why, even though `namespace` is set to `None` it gets manifested as `metadata.namespace: null`? From what I understand all `None` values are removed from the final manifestations.
> 
> Note
> 
> this is the reason why tests are failing now)

@disrupted 🙏

daconstenla:
@yannick-roeder can you help here?

disrupted:
Hi @daconstenla, sorry for the late reply. I was sick for the last 2 weeks.

> From what I understand all `None` values are removed from the final manifestations.

This is only true for `kpops generate` since `None` is a sentinel value there. For the manifestations we currently don't remove any values. If we would want to remove `None` values this would need to be changed here:

```patch
diff --git a/kpops/manifests/kubernetes.py b/kpops/manifests/kubernetes.py
index 1f47a5cd..8f68ac46 100644
--- a/kpops/manifests/kubernetes.py
+++ b/kpops/manifests/kubernetes.py
@@ -77,4 +77,4 @@ class KubernetesManifest(CamelCaseConfigModel):
 
     @override
     def model_dump(self, **_: Any) -> dict[str, Any]:
-        return super().model_dump(mode="json")
+        return super().model_dump(mode="json", exclude_none=True)
```

Is it really necessary to alter the manifests like this? I feel like I would prefer to keep them unchanged/frozen.

disrupted:
on second thought, I believe we can change this in Kubernetes [`ObjectMeta`](https://github.com/bakdata/kpops/blob/18ec312f23166deccf4c2a73112f847c3c76d711/kpops/manifests/kubernetes.py#L16) model without changing the behavior globally. I believe there it makes sense to exclude `None` values as they also act as sentinel.

I'll push a commit with the proposal.

daconstenla:
@disrupted I'm confused, [`ObjectMeta`](https://github.com/bakdata/kpops/blob/18ec312f23166deccf4c2a73112f847c3c76d711/kpops/manifests/kubernetes.py#L16) does have `namespace` field as `namespace: str | None = None`, and `StrimziKafkaTopic` refers to an [`ObjectMeta`](https://github.com/bakdata/kpops/blob/main/kpops/manifests/strimzi/kafka_topic.py#L55) which is the same as any other [`kubernetes`](https://github.com/bakdata/kpops/blob/main/kpops/manifests/kubernetes.py#L53) resource where `None` values don't get a representation in the resultant yaml, am I missing something?

disrupted:
> @disrupted I'm confused, [`ObjectMeta`](https://github.com/bakdata/kpops/blob/18ec312f23166deccf4c2a73112f847c3c76d711/kpops/manifests/kubernetes.py#L16) does have `namespace` field as `namespace: str | None = None`, and `StrimziKafkaTopic` refers to an [`ObjectMeta`](https://github.com/bakdata/kpops/blob/main/kpops/manifests/strimzi/kafka_topic.py?rgh-link-date=2025-02-03T13%3A30%3A22Z#L55) which is the same as any other [`kubernetes`](https://github.com/bakdata/kpops/blob/main/kpops/manifests/kubernetes.py?rgh-link-date=2025-02-03T13%3A30%3A22Z#L53) resource does where `None` values don't get a representation in the resultant yaml, am I missing something?

I think the main point of confusion here is that Pydantic keeps track of _set_ and _unset_ fields. We use a custom serializer function to exclude unset fields during model dump. https://github.com/bakdata/kpops/blob/18ec312f23166deccf4c2a73112f847c3c76d711/kpops/manifests/kubernetes.py#L46
This meant that all the `None` values were hidden from the result. If you do however explicitly set `metadata.namespace = None` it would be included in the serialization result. I've now changed the serializer function to also exclude `None` values.
Does that answer your question?

daconstenla:
> > @disrupted I'm confused, [`ObjectMeta`](https://github.com/bakdata/kpops/blob/18ec312f23166deccf4c2a73112f847c3c76d711/kpops/manifests/kubernetes.py#L16) does have `namespace` field as `namespace: str | None = None`, and `StrimziKafkaTopic` refers to an [`ObjectMeta`](https://github.com/bakdata/kpops/blob/main/kpops/manifests/strimzi/kafka_topic.py?rgh-link-date=2025-02-03T13%3A30%3A22Z#L55) which is the same as any other [`kubernetes`](https://github.com/bakdata/kpops/blob/main/kpops/manifests/kubernetes.py?rgh-link-date=2025-02-03T13%3A30%3A22Z#L53) resource does where `None` values don't get a representation in the resultant yaml, am I missing something?
> 
> I think the main point of confusion here is that Pydantic keeps track of _set_ and _unset_ fields. We use a custom serializer function to exclude unset fields during model dump.
> 
> https://github.com/bakdata/kpops/blob/18ec312f23166deccf4c2a73112f847c3c76d711/kpops/manifests/kubernetes.py#L46
> 
> 
> This meant that all the unset `None` values were hidden from the result. If you do however explicitly set `metadata.namespace = None` it would be included in the serialization result. I've now changed the serializer function to also exclude `None` values.
> Does that answer your question?

Aha, that explains it, I'm not that familiar with pydantic, I guess then, alternatively we could skip setting the value in case is `None` in the configuration here `kpops/manifests/strimzi/kafka_topic.py` which should produce the expected result.

disrupted:
> Aha, that explains it, I'm not that familiar with pydantic, I guess then, alternatively we could skip setting the value in case is `None` in the configuration here `kpops/manifests/strimzi/kafka_topic.py` which should produce the expected result.

yes, that would work as well. Your call.

daconstenla:
> > Aha, that explains it, I'm not that familiar with pydantic, I guess then, alternatively we could skip setting the value in case is `None` in the configuration here `kpops/manifests/strimzi/kafka_topic.py` which should produce the expected result.
> 
> yes, that would work as well. Your call.

I think it's cleaner if avoiding setting the value which keeps the change more contained.
Pushed, if tests are positive I'll mark it as ready, thanks for your support and I hope you're fully recovered 😉

daconstenla:
Ok, it seems to work as expected and tests went through, I think it's ready for review now 🙏

kpops/config/__init__.py
```diff
@@ -33,6 +33,12 @@ class StrimziTopicConfig(BaseSettings):
         description="The label to identify the KafkaTopic resources managed by the Topic Operator. This does not have to be the name of the Kafka cluster. It can be the label assigned to the KafkaTopic resource. If you deploy more than one Topic Operator, the labels must be unique for each. That is, the operators cannot manage the same resources.",
     )
 
+    namespace: str | None = Field(
+        default=None,
+        alias="namespace",
```
disrupted, line 38: the alias is not necessary if it's the same as the attribute name

kpops/manifests/strimzi/kafka_topic.py
```diff
@@ -60,15 +60,25 @@ class StrimziKafkaTopic(KubernetesManifest):
...
+                }
+            )
+            if strimzi_topic.namespace is not None
+            else ObjectMeta.model_validate(
+                {
+                    "name": topic.name,
+                    "labels": {cluster_domain: cluster_name},
+                }
+            )
         )
```
disrupted, line 82:
```suggestion
        metadata = ObjectMeta(
            name=topic.name,
            labels={cluster_domain: cluster_name},
        )
        if strimzi_topic.namespace:
            metadata.namespace = strimzi_topic.namespace
```

tests/manifests/strimzi/test_kafka_topic.py
```diff
@@ -48,6 +48,7 @@ def test_topic_spec_validation():
 def test_strimzi_kafka_topic_from_topic(kafka_topic: KafkaTopic, mocker: MockerFixture):
     mock_config = MagicMock()
     mock_config.strimzi_topic.cluster_labels = ("bakdata.com/cluster", "my-cluster")
+    mock_config.strimzi_topic.namespace = None
```
disrupted, line 51: isn't that the default value?

kpops/config/__init__.py
```diff
@@ -46,7 +52,7 @@ def label_validator(cls, label: dict[str, str]) -> dict[str, str]:
             raise ValidationError(msg)
         if len(label) > 1:
             log.warning(
-                "'resource_label' only reads the first entry in the dictionary. Other defined labels will be ignored."
+                "'strimzi_topic.label' only reads the first entry in the dictionary. Other defined labels will be ignored."
```
disrupted, line 55: thanks for fixing it!

kpops/manifests/strimzi/kafka_topic.py
```diff
@@ -60,15 +60,25 @@ class StrimziKafkaTopic(KubernetesManifest):
     def from_topic(cls, topic: KafkaTopic) -> Self:
         strimzi_topic = get_config().strimzi_topic
         if not strimzi_topic:
-            msg = "When manifesting KafkaTopic you must define 'strimzi_topic.resource_label' in the config.yaml"
+            msg = "When manifesting KafkaTopic you must define 'strimzi_topic.label' in the config.yaml"
```
disrupted, line 63: thanks!

tests/manifests/strimzi/test_kafka_topic.py
daconstenla: It fails due to mocker assigning an unexpected type, let me get it to be more concrete:

```python
E       pydantic_core._pydantic_core.ValidationError: 1 validation error for ObjectMeta
E       namespace
E         Input should be a valid string [type=string_type, input_value=<MagicMock name='mock.str...espace' id='4596251488'>, input_type=MagicMock]
E           For further information visit https://errors.pydantic.dev/2.10/v/string_type
```

disrupted: alright, then it's fine

daconstenla:
Last commit makes the documentation action fail, probably missing `uv run --frozen` to have dependencies available on command 👀

daconstenla:
@disrupted are you looking into the errors of the actions? (they seem unrelated to the changes we did)

disrupted (approved):

disrupted:
> @disrupted are you looking into the errors of the actions? (they seem unrelated to the changes we did)

yes, I'll look into those. perhaps they're just temporary CI issues. I've otherwise approved the changes in your PR. Thanks for your contribution!

daconstenla:
> > @disrupted are you looking into the errors of the actions? (they seem unrelated to the changes we did)
> 
> yes, I'll look into those. perhaps they're just temporary CI issues. I've otherwise approved the changes in your PR. Thanks for your contribution!

Thank you for your support, I'll leave it to you then. 
Looking forward for this version to be ready ❤️

PR #582, opened 2025-01-13
feat: add substitution variable `${pipeline.parent.name}`

jkbe:


tests/compiler/test_pipeline_name.py
```diff
@@ -23,6 +23,8 @@ def test_should_set_pipeline_name_with_default_base_dir():
     assert ENV["pipeline.name_3"] == "for"
     assert ENV["pipeline.name_4"] == "testing"
 
+    assert ENV["pipeline.parent_name"] == "some-random-path-for"
```
disrupted, line 26: so this is essentially equivalent to `${pipeline.name_0}-${pipeline.name_1}-${pipeline.name_2}-${pipeline.name_3}`?

JakobEdding: Exactly!

disrupted: while I am generally fine with adding this for convenience, it raises the question whether we will need the rest of those in the future. essentially we could do `pipeline.parent.parent.<...>.name` for the each higher level

JakobEdding: Yup, I was thinking about this too. Does this information perhaps already exist in the networkx graph that we're creating somewhere?

JakobEdding: But the variables available in the scope of this method should be enough to create env vars like you proposed.

disrupted: please also check for overlap with the existing `pipeline.path` environment variable.

JakobEdding: Wdym by overlap? As far as I can tell, `${pipeline.path}` is the absolute path to the pipeline.

disrupted: to me the meaning gets a little muddy here. I guess `pipeline.parent` doesn't mean parent pipeline but directory name of the parent path. We should decide which namespace fits better or how we can make this easy to understand and extend. e.g. `pipeline.parent_name` like you suggested or something like `pipeline.path.parent.name`

jkbe: Generally, I think you have a better overview of where this best fits among existing concepts of KPOps. 

> please also check for overlap with the existing ${pipeline.path} environment variable.
I think calling it something along the lines of `path` is misleading because the value is only derived from the path. 

I'd be in favor of `pipeline.parent.name` because it is easier to extend to `pipeline.parent.parent.<...>.name` later on compared to `pipeline.parent_name`.

jkbe: I've changed it now accordingly @disrupted

disrupted (commented):
please add it to the docs https://bakdata.github.io/kpops/latest/user/core-concepts/variables/substitution/#pipeline-name-variables

PR #594, opened 2025-01-16
fix: address CI warnings introduced by #580

disrupted:
didn't pay intention that the CI was skipped on the previous PR

### DIFF
--- a/kpops/cli/utils.py
+++ b/kpops/cli/utils.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
 from collections.abc import Iterable, Iterator
-from pathlib import Path
 from glob import glob
+from pathlib import Path
 
 from kpops.const.file_type import PIPELINE_YAML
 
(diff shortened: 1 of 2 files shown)


kpops/cli/utils.py
```diff
@@ -25,7 +25,10 @@ def collect_pipeline_paths(pipeline_paths: Iterable[Path]) -> Iterator[Path]:
             # docs.python.org/3.13#pathlib.Path.glob, probably it make sense to use it after ugprading,
             # likely the code will look like:
             # yield from sorted(pipeline_path.glob(f"**/{PIPELINE_YAML}", recurse_symlinks=True))
-            yield from sorted(Path(p).resolve() for p in glob(f"{pipeline_path}/**/{PIPELINE_YAML}", recursive=True))
+            yield from sorted(
+                Path(p).resolve()
+                for p in glob(f"{pipeline_path}/**/{PIPELINE_YAML}", recursive=True)  # noqa: PTH207
+            )
```
JakobEdding, line 31: Perhaps actually extract a method here, e.g. for the `glob()` call

JakobEdding (approved):

kpops/cli/utils.py
disrupted: good point, but I am rather hoping to resolve this with the todo comment above

PR #595, opened 2025-01-16
style: format YAML & JSON

disrupted:


tobbber (approved):
looks good to me. Should we also update the developer [docs](https://bakdata.github.io/kpops/latest/developer/contributing/)?

PR #596, opened 2025-01-16
ci: run lint job only for target Python version on Linux

disrupted:
atm I don't really see the benefit of repeating lint runs against multiple targets, especially since the Windows runners are so slow (usually ~6x slower than Ubuntu)
we can always reenable those if needed

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -23,18 +23,11 @@ jobs:
       matrix:
         os:
           - ubuntu-24.04
-          - windows-2022
-        python-version: ["3.11", "3.12"]
-        is-draft-pr:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -23,18 +23,11 @@ jobs:
       matrix:
         os:
           - ubuntu-24.04
-          - windows-2022
-        python-version: ["3.11", "3.12"]
-        is-draft-pr:
-          - ${{ github.event.pull_request.draft }}
+        python-version: ["3.11"]
```
disrupted, line 26: note for reviewer: Python 3.11 because that's our min target set in pyproject.toml

tobbber (approved):

PR #601, opened 2025-02-27
refactor(helm-diff)!: configure on HelmApp component level

### LINKED ISSUE #509: Helm diff config should be configured on component level and not in global config
Currently, HelmDiff config is [global](https://github.com/bakdata/kpops/blob/main/kpops/config.py#L116). However, it makes more sense to configure this for HelmApp because different helm apps likely use different ignore configs

disrupted:
Fixes #509


kpops/components/base_components/helm_app.py
```diff
@@ -68,12 +69,21 @@ class HelmApp(KubernetesApp):
         expects a path to local Helm chart.
     :param version: Helm chart version, defaults to None
     :param values: Helm app values
+    :param diff_config: Helm diff config
     """
 
+    # TODO: combine Helm config, e.g.
+    # helm:
+    #   repo: {}
+    #   diff: {}
```
disrupted, line 78: shall we do that now since we're making breaking changes anyways?

philipp94831: Aren't they already combined because this is the Helm component? One could argue that values and version also belong to the helm config

disrupted: true, then I won't change it.

philipp94831 (approved):

PR #608, opened 2025-03-03
fix(cleaner): enrich on init

### LINKED ISSUE #605: ProducerAppCleaner and StreamsAppCleaner do not apply defaults of Cleaner
Consider a defaults definition
```yaml
cleaner:
  diff_config:
    ignore: []
```

For `KafkaConnectorResetter`, this is properly inherited by [calling the constructor](https://github.com/bakdata/kpops/blob/b6efa6876ccbb8ef371d586a936980fe36db4fde/kpops/components/base_components/kafka_connector.py#L149). For `ProducerAppCleaner` and `StreamsAppCleaner` this is not the case because they are instantiated by [Pydantic's `model_validate`](https://github.com/bakdata/kpops/blob/b6efa6876ccbb8ef371d586a936980fe36db4fde/kpops/components/streams_bootstrap/streams/streams_app.py#L107)

disrupted:
Fixes #605

turns out it's not related to Pydantic `model_validate`, but we were unintentionally skipping the defaults enrichment, because `enrich=False` was passed as init arg

### DIFF
--- a/tests/pipeline/resources/streams-bootstrap/defaults.yaml
+++ b/tests/pipeline/resources/streams-bootstrap/defaults.yaml
@@ -1,9 +1,11 @@
 cleaner:
   diff_config:
-    ignore:
-      - ["foo", "bar"]
+    ignore: []
 
 streams-bootstrap:
+  diff_config:
+    ignore:
+      - ["foo", "bar"]
   version: "3.6.1"
   values:
     imagePullPolicy: IfNotPresent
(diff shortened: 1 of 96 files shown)


tests/pipeline/resources/streams-bootstrap/defaults.yaml
```diff
@@ -1,3 +1,8 @@
+cleaner:
```
philipp94831, line 1: I don't see where you validate that in tests

disrupted: enrichment from defaults does work now as long as the parent `diff_config.ignore` is not set.

philipp94831 (approved):

philipp94831 (approved):

PR #611, opened 2025-03-04
refactor(pydantic): create `SkipGenerate` type

disrupted:
can be used to omit specific `PipelineComponent` fields from `kpops generate` output

### DIFF
--- a/kpops/components/base_components/pipeline_component.py
+++ b/kpops/components/base_components/pipeline_component.py
@@ -4,7 +4,8 @@ from abc import ABC
 from collections.abc import Iterator
 from typing import Any, ClassVar
 
-from pydantic import AliasChoices, ConfigDict, Field
+import pydantic
+from pydantic import AliasChoices, ConfigDict, Field, model_serializer
 
 from kpops.components.base_components.base_defaults_component import (
     BaseDefaultsComponent,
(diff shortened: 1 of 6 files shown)


kpops/components/base_components/pipeline_component.py
```diff
@@ -276,3 +278,15 @@ async def clean(self, dry_run: bool) -> None:
...
+    def serialize_model(
+        self,
+        default_serialize_handler: pydantic.SerializerFunctionWrapHandler,
+        info: pydantic.SerializationInfo,
+    ) -> dict[str, Any]:
+        result = default_serialize_handler(self)
+        # HACK: workaround for exclude_none, which is otherwise evaluated too early
+        if info.exclude_none:
+            return exclude_by_value(result, None)
+        return result
```
jkbe, line 292: Have you considered letting `PipelineComponent` also inherit from `SerializeAsOptionalModel`? Then you wouln't have to duplicate this method AFAICS.

jkbe (dismissed):

kpops/components/base_components/pipeline_component.py
disrupted: good point! applied

jkbe (approved):

PR #622, opened 2025-03-26
refactor(helm): cache repos and version across components

daconstenla:
In pipelines where there are many `helm` applications, each of them will execute `helm version --short` and if from those, multiple use the same helm repo, it'll run `helm repo add` and `helm repo update` for each of them, this change makes those values static and checks them before executing a `helm` command which is more resource heavy.

> [!WARNING]
> tests are failing but I don't want to invest time if this is not an interesting option, will wait for feedback before fixing.

## description of changes

Add static properties on `Helm` class to re-use already obtained helm outputs instead of executing them each time.
- `helm version --short`
- `helm repo add ...`
- `helm repo update ...`

## evaluation

- Before the version / repository cache was introduce, the test pipeline executed took: 
`poetry run kpops deploy ... >   9.06s user 2.60s system 64% cpu 18.059 total`
- After adding the `helm version` cache, number went down by a little: 
`poetry run kpops deploy ... >   7.70s user 2.19s system 61% cpu 16.175 total`
- After adding repository cache went down even more: 
`poetry run kpops deploy ... >   4.32s user 1.11s system 66% cpu 8.223 total`

### DIFF
--- a/kpops/component_handlers/helm_wrapper/helm.py
+++ b/kpops/component_handlers/helm_wrapper/helm.py
@@ -31,15 +31,21 @@ log = logging.getLogger("Helm")
 
 @final
 class Helm:
-    _version: Version = None
-    _repos: map[str, str] = {}
+    _already_run_helm_version: bool = False
+    _version: Version
+    _repos: dict[str, str] = {}
+    _context: str | None = None
+    _debug: bool = False
 
     def __init__(self, helm_config: HelmConfig) -> None:
-        Helm._version = Helm.get_version()
-        self._context = helm_config.context
-        self._debug = helm_config.debug
+        Helm._version = (
+            Helm.get_version() if not Helm._already_run_helm_version else Helm._version
+        )
+        Helm._context = helm_config.context
+        Helm._debug = helm_config.debug
+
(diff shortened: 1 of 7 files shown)


disrupted:
thanks @daconstenla, I agree that this is a valuable improvement

kpops/component_handlers/helm_wrapper/helm.py
```diff
@@ -31,11 +31,14 @@
 
 @final
 class Helm:
+    _version: Version = None
```
disrupted, line 34: can we do this directly and remove it from init?
```suggestion
    _version: Version = Helm.get_version()
```

daconstenla: I tried and failed, it seems like we can't init static variables from static methods from the same class, I get the following error:

```
NameError: name 'Helm' is not defined
```

daconstenla:
> I've refactored it a little to use standard caching libraries and make `Helm` a proper singleton class. also added tests for the caching.

awesome, thanks @disrupted great updates.

disrupted:
thanks for your contribution 🤝

PR #648, opened 2026-06-01
Feat: Make helm flags `timeout` and `force` configurable

### LINKED ISSUE #647: Make Helm flags `timeout` and `force` configurable.
Helm flags `--timeout`, `--force`, `--wait` and `--wait-for-jobs` are currently not configurable.

They should be configurable either for individual [HelmApps](https://github.com/bakdata/kpops/blob/dc1c51d280ae5ee5abc10e33835cdb4cf33194f2/kpops/components/base_components/helm_app.py) or in the (global) HelmConfig.

tobbber:
Closes #647


tests/component_handlers/helm_wrapper/test_helm_diff.py
```diff
@@ -169,27 +169,28 @@ def test_calculate_changes_new_release(self, helm_diff: HelmDiff):
...
-        )
-        assert caplog.messages == [
-            "\n"
-            "\x1b[32m+ apiVersion: v1\n"
-            "\x1b[0m\x1b[32m+ kind: Deployment\n"
-            "\x1b[0m\x1b[32m+ metadata:\n"
-            "\x1b[0m\x1b[32m+   a: '1'\n"
-            "\x1b[0m"
-        ]
+        with caplog.at_level(logging.INFO, logger=logger.name):
```
tobbber, line 172: unrelated fix. test required other tests to run first and failed if ran alone because the logger was never configured with log level info

tests/components/test_helm_app.py
```diff
@@ -172,6 +174,67 @@ async def test_should_raise_not_implemented_error_when_helm_chart_is_not_set(
...
+            "prioritize local over default timeout",
+            "fallback to default timeout",
+        ],
+        argnames="local_timeout, global_timeout, expected_timeout",
+        argvalues=[
+            ("30m", "10m", "30m"),
+            (None, "10m", "10m"),
+            ("30m", None, "30m"),
+            (None, None, "5m0s"),
+        ],
```
disrupted: I would prefer the ids together with the test case, e.g. `pytest.param("30m", "10m", "30m", id="prioritize local over global timeout")`
that way they can't get mixed up in the future

kpops/component_handlers/helm_wrapper/model.py
```diff
@@ -117,6 +115,7 @@ def to_command(self) -> list[str]:
         if self.version:
             command.extend(["--version", self.version])
         if self.force:
+            # TODO: --force was renamed to --force_replace in Helm v4, we should switch when dropping helm v3 support
```
disrupted:
```suggestion
            # TODO: --force was renamed to --force-replace in Helm v4, we should switch when dropping helm v3 support
```
good to know

disrupted: technically we could pass the Helm Version to the `to_command` method and decide between `--force` and `--force-replace`. that would be a little more clean and avoid the deprecation warning

tobbber: yeah i thought about that too, maybe as an optional parameter? only thing is that then the method signature is different from the parent class `RepoAuthFlags`

kpops/component_handlers/helm_wrapper/model.py
```diff
@@ -76,25 +83,23 @@ class HelmConfig(DescConfigModel):
     :param context: Name of kubeconfig context (`--kube-context`)
     :param debug: Run Helm in Debug mode
     :param api_version: Kubernetes API version used for `Capabilities.APIVersions`
+    :param timeout: Helm flag --timeout. Duration to wait for any individual Kubernetes operation
+    :param force_replace: Helm flag --force. Forces resource updates by replacement
```
disrupted: should we update the description here as well since we'll use Helm 4 going forward?

```suggestion
    :param force_replace: Helm flag --force-replace. Forces resource updates by replacement
```

disrupted (approved):
one small remark, other than that LGTM

kpops/component_handlers/helm_wrapper/model.py
tobbber: yeah good idea

disrupted (approved):

PR #651, opened 2026-07-28
chore(dev): report missing Python type annotations

disrupted:


pyproject.toml
disrupted, line 1: configured new rules for Ruff and BasedPyright

kpops/utils/environment.py
```diff
@@ -11,32 +14,37 @@ class Environment(UserDict[str, str]):
     def __init__(
         self, mapping: MutableMapping[str, str] | None = None, /, **kwargs: str
     ) -> None:
-        self._global = os.environ
+        self._global: _Environ[str] = os.environ
```
tobbber: maybe use `Mapping[str, str]` as type annotation instead of private _Environ?

tobbber (approved):

kpops/utils/environment.py
disrupted: good idea!

PR #659, opened 2026-07-31
feat(streams-bootstrap): add KEDA triggers and scaling modifiers

### LINKED ISSUE #660: Type errors in the autoscaling model
There are currently two bugs in KPOps:

1. `additionalTriggers` has an incorrect type hint. This results in a validation error in KPOps.
2. Large values for `autoscaling.lagThreshold` are converted to floats in the KEDA resource. For example, `999999` becomes `1e+07`.

The `lagThreshold` issue can be fixed in either of the following ways:

Change `StreamsAppAutoScaling.lagThreshold` type to `str`, but this will require updates to the pipeline files.
Alternatively, extend the model with a field serializer that converts the `lagThreshold` value to `str`.

raphala:
Fixes #660


tobbber (approved):

kpops/components/streams_bootstrap/common/model.py
```diff
@@ -161,7 +165,9 @@ class StreamsAppAutoScaling(
     )
     internal_topics: SerializeAsOptional[list[str]] = []
     topics: SerializeAsOptional[list[str]] = []
-    additional_triggers: SerializeAsOptional[list[str]] = []
+    triggers: SerializeAsOptional[list[dict[str, Any]]] = []
+    additional_triggers: SerializeAsOptional[list[dict[str, Any]]] = []
+    scaling_modifiers: dict[str, Any] | None = None
```
disrupted, line 170: did you consider using
```suggestion
    scaling_modifiers: SerializeAsOptional[dict[str, Any]] = {}
```
here as well?

raphala: `None` seemed to fit better, because `scalingModifiers` is more of a record than a collection (and has mandatory fields `target`/`formula` if set).

disrupted (approved):

kpops/components/streams_bootstrap/common/model.py
```diff
@@ -161,7 +165,9 @@ class StreamsAppAutoScaling(
...
 
     model_config: ClassVar[ConfigDict] = ConfigDict(extra="allow")
```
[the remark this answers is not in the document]
disrupted: this is validated by the Helm Chart itself during deploy dry-run which is sufficient for now. further static validation is not necessary at this point, but could be added later

PR #661, opened 2026-08-05
Fix type errors in the autoscaling model

### LINKED ISSUE #660: Type errors in the autoscaling model

mzmbq:
Fixes #660:

1. Fixes the incorrect type for `additionalTriggers`.
2. Adds a `pydantic.field_serializer` to convert `lagThreshold` to `str`.

### DIFF
--- a/kpops/components/streams_bootstrap/common/model.py
+++ b/kpops/components/streams_bootstrap/common/model.py
@@ -165,6 +165,12 @@ class StreamsAppAutoScaling(
 
     model_config: ClassVar[ConfigDict] = ConfigDict(extra="allow")
 
+    @pydantic.field_serializer("lag_threshold")
+    def serialize_lag_threshold(self, lag_threshold: int | None) -> str | None:
+        if lag_threshold is None:
+            return None
+        return str(lag_threshold)
+
 
 class PersistenceConfig(CamelCaseConfigModel, DescConfigModel):
     """streams-bootstrap persistence configurations.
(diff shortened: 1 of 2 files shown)


kpops/components/streams_bootstrap/common/model.py
```diff
@@ -161,10 +161,16 @@ class StreamsAppAutoScaling(
     )
     internal_topics: SerializeAsOptional[list[str]] = []
     topics: SerializeAsOptional[list[str]] = []
-    additional_triggers: SerializeAsOptional[list[str]] = []
+    additional_triggers: SerializeAsOptional[list[dict[str, Any]]] = []
```
jnsrnhld, line 164: Can you add a test that adds an additional trigger?

For example:

```yaml
additional_triggers:
- type: cron
  metadata:
    timezone: Asia/Kolkata
    start: 0 6 * * *
    end: 0 20 * * *
    desiredReplicas: "10"
```

mzmbq: Added 0c03ae3

disrupted:
in its current state duplicate of #659

